PR #49, opened 2014-08-28
Better Team Pages

NoxiusInfernus:
First attempt at adding Better Team Pages to BetterTTV

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -31,7 +31,7 @@
  *  Free to modify for personal use
  *  Need permission to distribute the code
  *  Can't sell addon or features of the addon
- *  
+ *
  */
 /** @license
  * Gritter for jQuery
@@ -326,7 +326,7 @@ module.exports = function() {
         if(emote.restrictions) {
             if(emote.restrictions.channels.length && emote.restrictions.channels.indexOf(bttv.getChannel()) === -1) return;
             if(emote.restrictions.games.length && tmi().channel && emote.restrictions.games.indexOf(tmi().channel.game) === -1) return;
-     
+
             if(emote.restrictions.emoticonSet && emoteSets.indexOf(emote.restrictions.emoticonSet) === -1) return;
         }
 
@@ -722,7 +722,7 @@ var lookupDisplayName = exports.lookupDisplayName = function(user) {
     if(vars.userData.isLoggedIn && user === vars.userData.login) {
         store.displayNames[user] = Twitch.user.displayName() || user;
     }
-    
+
     // Get subscription status (Night's subs)
     bttv.io.lookupUser(user);
 
@@ -1309,7 +1309,7 @@ var newRoom = exports.newRoom = function(name) {
     var emberRoom = null;
     var groupRooms = bttv.getChatController().get('connectedPrivateGroupRooms');
     var channelRoom = bttv.getChatController().get('currentChannelRoom');
-    
+
     if(channelRoom.get('id') === name) {
         emberRoom = channelRoom;
     } else {
@@ -1862,7 +1862,7 @@ var bttvMessageTokenize = exports.bttvMessageTokenize = function(sender, message
         var piece = tokenizedString[i];
 
         if(bttv.settings.get('chatImagePreview') === true) {
-            var imageTest = new RegExp('(https?:\/\/.)([a-z\-_0-9\/\:\.\%]*\.(jpg|jpeg|png|gif))', 'i');
+            var imageTest = new RegExp('(https?:\/\/.)([a-z\-_0-9\/\:\.\%\+]*\.(jpg|jpeg|png|gif))', 'i');
             if (imageTest.test(piece)) {
                 piece = bttv.chat.imagePreview(piece);
                 tokenizedString[i] = piece;
@@ -2392,7 +2392,7 @@ var main = function () {
                                 clearClutter();
                                 channelReformat();
                                 if (
-                                    App.__container__.lookup("controller:channel").get("theatreMode") === false && 
+                                    App.__container__.lookup("controller:channel").get("theatreMode") === false &&
                                     bttv.settings.get('autoTheatreMode') === true
                                 ) {
                                     enableTheatreMode();
@@ -2527,13 +2527,13 @@ var debug = require('../helpers/debug');
 
 module.exports = function () {
 	if(
-		!window.Ember || 
-		!window.App || 
+		!window.Ember ||
+		!window.App ||
 		App.__container__.lookup("controller:application").get("currentRouteName") !== "channel.index"
 	) {
 		return;
 	}
-	
+
     // Call 'toggleTheatre' action on the channel controller in Ember
     App.__container__.lookup('controller:channel').send('toggleTheatre');
 }
@@ -2820,7 +2820,7 @@ module.exports = function () {
 
     $("#right_col .bottom #controls #control_buttons .primary_button").css({
         float: 'right',
-        marginRight: '-1px' 
+        marginRight: '-1px'
     });
     $("#right_nav").css({
         'margin-left': 'auto',
@@ -2866,7 +2866,7 @@ module.exports = function () {
 },{"../../helpers/debug":43,"../../keycodes":46,"../../vars":57,"./handle-resize":16,"./twitchcast":18}],18:[function(require,module,exports){
 module.exports = function() {
     if($('.archive_info').length) return;
-    
+
     var template = '<iframe id="twitchcast" src="https://nightdev.com/twitchcast/?ontwitch={{hostname}}&channel={{channel}}" width="100%" height="100%" style="position: absolute;top: 0px;left: 0px;border: none;"></iframe>';
 
     var openTwitchCast = function() {
@@ -2887,7 +2887,7 @@ module.exports = function() {
         }
         window.addEventListener("message", close, false);
     }
-    
+
     var placeButton = function() {
         if($('#twitchcast_button').length) return;
 
@@ -2903,7 +2903,7 @@ module.exports = function() {
         if(window.chrome.cast && window.chrome.cast.isAvailable) {
             return callback(false);
         }
-        
+
         setTimeout(function() {
             castAvailable(callback);
         }, 1000);
@@ -3052,7 +3052,7 @@ var checkBroadcastInfo = module.exports = function() {
     bttv.TwitchAPI.get("channels/"+channel).done(function(d) {
         if(d.game) {
             var $channel = $('#broadcast-meta .channel');
-            
+
             if($channel.find('.playing').length) {
                 $channel.find('a:eq(1)').text(d.game).attr("href", Twitch.uri.game(d.game)).removeAttr('data-ember-action');
             }
@@ -3161,7 +3161,7 @@ module.exports = function () {
                 if (notificationsLoaded === true && notifications < j) {
                     $.get('/messages/inbox', function (data) {
                         var $message = $(data).find("#message-list .unread:first");
-                            
+
                         if ($message) {
                             var $senderData = $message.children("div.from_to_user"),
                                 $messageData = $message.children("div.message_data"),
@@ -3321,7 +3321,7 @@ var debug = require('../helpers/debug');
 
 function load(file, key){
     if(!bttv.settings.get(key)) return;
-    
+
     var css = document.createElement("link");
     css.setAttribute("href", "//cdn.betterttv.net/style/stylesheets/betterttv-"+file+".css?"+bttv.info.versionString());
     css.setAttribute("type", "text/css");
@@ -3334,7 +3334,7 @@ function unload(key){
 }
 
 module.exports.load = load;
-module.exports.unload = unload; 
+module.exports.unload = unload;
 },{"../helpers/debug":43}],26:[function(require,module,exports){
 var debug = require('../helpers/debug'),
     handleBackground = require('./handle-background');
@@ -3493,7 +3493,7 @@ module.exports = function (pollId) {
     if(!bttv.settings.get('embeddedPolling')) return;
 
     var $poll = $('#bttv-poll-contain');
-    
+
     // Dont replace the poll with the same one
     if($poll.length && pollId == lastPollId) return;
 
@@ -3508,12 +3508,12 @@ module.exports = function (pollId) {
 
     // Reset $poll to newly created poll
     $poll = $('#bttv-poll-contain');
-    
+
     // If timeout exists already, clear it
     if(frameTimeout !== null) {
         clearTimeout(frameTimeout);
     }
-    
+
     // After 30 seconds, remove poll if user doesn't open it
     frameTimeout = setTimeout(function() {
         if($poll && !$poll.children('.frame').is(':visible')) $poll.remove();
@@ -3532,7 +3532,7 @@ module.exports = function (pollId) {
     });
 
     $poll.slideDown(200);
-    
+
     lastPollId = pollId;
 }
 
@@ -3658,7 +3658,7 @@ module.exports = function () {
 },{"../helpers/debug":43}],34:[function(require,module,exports){
 module.exports = function handleBackground(tiled) {
     var tiled = tiled || false;
-    
+
     var canvasID = 'custom-bg';
 
     if($("#"+canvasID).length === 0) {
@@ -4049,7 +4049,7 @@ module.exports = function () {
         data.emotes.forEach(function(emote, count) {
             emote.urlTemplate = data.urlTemplate.replace('{{id}}', emote.id);
             emote.url = emote.urlTemplate.replace('{{image}}', '1x');
-            
+
             bttv.chat.store.bttvEmotes[emote.code] = emote;
         });
 
@@ -4097,7 +4097,7 @@ module.exports = function () {
         }).on('click', '.chat-line .emoticon', function() {
             var $emote = $(this);
             if($emote.data('channel') && $emote.data('channel') === 'BetterTTV Emotes') return;
-            
+
             if(bttv.TwitchEmoteIDToChannel && $emote.data('id') && bttv.TwitchEmoteIDToChannel[$emote.data('id')]) {
                 window.open('http://www.twitch.tv/'+bttv.TwitchEmoteIDToChannel[$emote.data('id')],'_blank');
             } else if($emote.data('channel')) {
@@ -4937,7 +4937,7 @@ function SocketClient() {
     // Night's legacy subs
     this.socket.on('new_subscriber', function(data) {
         if(data.channel !== bttv.getChannel()) return;
-        
+
         bttv.chat.helpers.notifyMessage("subscriber", bttv.chat.helpers.lookupDisplayName(data.user) + " just subscribed!");
         bttv.chat.store.__subscriptions[data.user] = ['night'];
         bttv.chat.helpers.reparseMessages(data.user);
@@ -4978,7 +4978,7 @@ SocketClient.prototype.lookupUser = function(name) {
 SocketClient.prototype.joinChannel = function() {
     if(!this._connected || !this.socket.connected) return;
     if(!bttv.getChannel().length) return;
-    
+
     this.socket.emit('join_channel', bttv.getChannel());
 
     // Night's legacy subs
@@ -8268,7 +8268,7 @@ JSONPPolling.prototype.doPoll = function () {
   this.script = script;
 
   var isUAgecko = 'undefined' != typeof navigator && /gecko/i.test(navigator.userAgent);
-  
+
   if (isUAgecko) {
     setTimeout(function () {
       var iframe = document.createElement('iframe');
--- a/src/chat/templates.js
+++ b/src/chat/templates.js
@@ -133,7 +133,7 @@ var bttvMessageTokenize = exports.bttvMessageTokenize = function(sender, message
         var piece = tokenizedString[i];
 
         if(bttv.settings.get('chatImagePreview') === true) {
-            var imageTest = new RegExp('(https?:\/\/.)([a-z\-_0-9\/\:\.\%]*\.(jpg|jpeg|png|gif))', 'i');
+            var imageTest = new RegExp('(https?:\/\/.)([a-z\-_0-9\/\:\.\%\+]*\.(jpg|jpeg|png|gif))', 'i');
             if (imageTest.test(piece)) {
                 piece = bttv.chat.imagePreview(piece);
                 tokenizedString[i] = piece;


src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+    var jquiCSS = $("<link>", {"href":"//ajax.googleapis.com/ajax/libs/jqueryui/1.11.1/themes/dark-hive/jquery-ui.min.css", "type":"text/css", "rel":"stylesheet"}),
+        jquiJS  = $("<script>", {"src":"//ajax.googleapis.com/ajax/libs/jqueryui/1.11.1/jquery-ui.min.js", "type":"text/javascript"}),
+        teamCSS = $("<link>", {"href":"//cdn.betterttv.net/style/stylesheets/betterttv-team-page.css?"+bttv.info.versionString(), "id":"betterTwitchTeams", "rel":"stylesheet", "type":"text/css"});
+    
+    if(bttv.settings.get("darkenedMode") === true) {
+        teamCSS = $("<link>", {"href":"//cdn.betterttv.net/style/stylesheets/betterttv-team-page-dark.css?"+bttv.info.versionString(), "id":"betterTwitchTeams", "rel":"stylesheet", "type":"text/css"});
+    }
+    $('body').append(jquiCSS, jquiJS, teamCSS);
+
+    $("h2").remove();
```
night: h2? This removes every h2 on the page. Please specifically name which h2s to remove, if any.

NoxiusInfernus: h2 is for the "Members" heading text below the team Icon in the left hand column. It's literally the only place on the page it is used. But I do see your point, in case they use it else where in the future.

I'm a little busy now, but I will change it tonight/tomorrow to only remove that specific heading from it's parent container.

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+    $("#about").prepend($("#banner_header"));
+
+    var followTeamButton = $("<a>", {"id":"followTeamBtn", "data-ember-action":"135", "class":"js-follow follow button primary action"});
```
night: ``data-ember-action`` should not included.

I'm not sure why, bet I've tended to use camelCase for IDs, and split-with-dashes for classes in BetterTTV.

Please also prefix any id inserted into Twitch's DOM with ``bttv`` to prevent any sort of namespace issues.
For classes, prefix with ``bttv-``.

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+        $("#team_member_list").css({"height":teamListHeight+"px"});
+    });
+
+    var maxId = setTimeout(function() {}, 0);
+    for(var i=0; i < maxId; i+=1) { 
+            clearTimeout(i);
+    }
+
+    var followList = "none",
+        uName = cookie.get("login");
```
night: BetterTTV holds your login info in ``vars.userData``. We have isLoggedIn and login keys: https://github.com/night/BetterTTV/blob/master/src/vars.js

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+        if($("#team_chat").length) {
+            var chatHeight = $("#live_player").height() - 2;
+            $("#team_chat").css("height", chatHeight+"px");
+        }
+
+        var teamListHeight = ( ($("#live_player").height() - $("#team_logo").height()) - 76 );
+        $("#team_member_list").css({"height":teamListHeight+"px"});
+    });
+
+    var maxId = setTimeout(function() {}, 0);
```
night: This line and the next 3 do not do anything. Why do they exist?

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+    
+    var throttledFollow = function() {
+        if(followList.length > 0) {
+            var targetChan = followList[0];
+            $("#followTeamBtn").css({"background-color":"#B9A3E3"});
+            $("#followTeamBtn").text("Following "+targetChan+" ...");
+
+            Twitch.api.put("/kraken/users/"+uName+"/follows/channels/"+targetChan)
```
night: We do not use Twitch.api.put, as Twitch requested us to attach our Client ID to requests. Please use bttv.TwitchAPI: https://github.com/night/BetterTTV/blob/master/src/twitch-api.js

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                    debug.log(uName+" is not subbed to "+chan);
+                }
+            })
+            .fail(function(d) {
+                debug.log("check if "+uName+" is subbed to "+chan+" failed");
+            });
+        }
+    }
+
+    var addStoof = function(val) {
```
night: Stoof? Lets try to use appropriate function names.

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+    .fail(function(d) {
+        debug.log(chan+" subs check failed");
+        
+        if(d.status == 404) {
+            debug.log(chan+" not in sub program");
+        }
+        addStoof(0);
+    });
+
+    var uName = cookie.get("login");
```
night: Same issue about vars

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    var followList = "none",
+        uName = cookie.get("login");
+        
+    vars.teamCurrentChannel = ($(".js-playing").attr("id")).replace("channel_", "");
+    vars.teamFirstLoad = 1;
+    setTimeout(loadTeam, 1000);
+    
+    var followTeam = function() {
```
night: How you're following teams is not done well. Instead of recursively calling with a setTimeout, I'd suggest dumping channels to follow into a queue, and rolling that queue on an interval until the queue is empty. Thus, you avoid three separate functions. If a follow fails, you might just dump the user back into the queue (and keep track of errors).

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+                followList.push(a.channel.name);
+            });
+            throttledFollow();
+        }
+    }
+    
+    var throttledFollow = function() {
+        if(followList.length > 0) {
+            var targetChan = followList[0];
+            $("#followTeamBtn").css({"background-color":"#B9A3E3"});
```
night: You call ``$("#followTeamBtn")`` repeatedly. Save it to a variable like ``$followTeamBtn``, where the ``$`` means it's a jQuery object.

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+        player = $("<object>", {"type":"application/x-shockwave-flash", "id":chan+"_video_embed", "class":"ttvFlashPlayer", "wmode":"transparent"});
+    player.attr("data", "http://www.twitch.tv/widgets/live_embed_player.swf?channel="+chan);
+    player.append(p1, p2, p3);
+    $("#standard_holder").append(player);
+
+    $("#team_chat").empty();
+    var chatEmbed = $("<iframe>", {"id":"chatEmbed", "src":"http://twitch.tv/chat/embed?channel="+chan, "frameborder":"0"});
+    $("#team_chat").append(chatEmbed);
+
+    Twitch.api.get("/api/channels/"+chan+"/product")
```
night: Same issue about Twitch.api

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+    var p1 = $("<param>", {"name":"allowFullScreen", "value":"true"}),
+        p2 = $("<param>", {"name":"allowScriptAccess", "value":"always"}),
+        p3 = $("<param>", {"name":"flashvars", "value":"channel="+chan+"&auto_play=true&start_volume=100"}),
+        player = $("<object>", {"type":"application/x-shockwave-flash", "id":chan+"_video_embed", "class":"ttvFlashPlayer", "wmode":"transparent"});
+    player.attr("data", "http://www.twitch.tv/widgets/live_embed_player.swf?channel="+chan);
+    player.append(p1, p2, p3);
+    $("#standard_holder").append(player);
+
+    $("#team_chat").empty();
+    var chatEmbed = $("<iframe>", {"id":"chatEmbed", "src":"http://twitch.tv/chat/embed?channel="+chan, "frameborder":"0"});
```
night: Old chat embed!

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+
+    $("div.member.js-playing").removeClass("js-playing");
+    $("#channel_"+chan).addClass("js-playing");
+
+    $("#standard_holder").empty();
+    var p1 = $("<param>", {"name":"allowFullScreen", "value":"true"}),
+        p2 = $("<param>", {"name":"allowScriptAccess", "value":"always"}),
+        p3 = $("<param>", {"name":"flashvars", "value":"channel="+chan+"&auto_play=true&start_volume=100"}),
+        player = $("<object>", {"type":"application/x-shockwave-flash", "id":chan+"_video_embed", "class":"ttvFlashPlayer", "wmode":"transparent"});
+    player.attr("data", "http://www.twitch.tv/widgets/live_embed_player.swf?channel="+chan);
```
night: Old embed player!

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+        if(d.status == 404) {
+            debug.log(chan+" not in sub program");
+        }
+        addStoof(0);
+    });
+
+    var uName = cookie.get("login");
+
+    var checkIsSubbed = function() {
+        if(typeof uName !== "undefined") {
```
night: I think ``if(typeof uName === "undefined") return;`` would be cleaner.

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                debug.log("check if "+uName+" is subbed to "+chan+" failed");
+            });
+        }
+    }
+
+    var addStoof = function(val) {
+        var jsnTeam = vars.jsnTeam;
+        
+        for(var i=0; i<jsnTeam.length; i++) {
+            if(jsnTeam[i].channel.name == chan) {
```
night: I think ``if(jsnTeam[i].channel.name !== chan) continue;`` would be cleaner

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                var chanActions = $("<div>", {"id":"channel_actions"}),
+                    followBtn = $("<a>", {"id":"followbtn", "data-ember-action":"135", "class":"js-follow follow button primary action"});
+                followBtn.text(" Follow ");
+                
+                followBtn.click(function(e) {
+                    followcurrentchannel();
+                });
+                chanActions.append(followBtn);
+
+                var shareBtn   = $("<div>", {"id":"sharebtn", "class":"button action primary"});
```
night: extraneous spacing

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    var addStoof = function(val) {
+        var jsnTeam = vars.jsnTeam;
+        
+        for(var i=0; i<jsnTeam.length; i++) {
+            if(jsnTeam[i].channel.name == chan) {
+                $("#channel_url").val("http://www.twitch.tv/"+chan);
+                $("#live_embed").val('<object type="application/x-shockwave-flash" height="378" width="620" id="live_embed_player_flash" data="http://www.twitch.tv/widgets/live_embed_player.swf?channel='+chan+'" bgcolor="#000000"><param name="allowFullScreen" value="true" /><param name="allowScriptAccess" value="always" /><param name="allowNetworking" value="all" /><param name="movie" value="http://www.twitch.tv/widgets/live_embed_player.swf" /><param name="flashvars" value="hostname=www.twitch.tv&channel='+chan+'&auto_play=true&start_volume=25" /></object>');
```
night: Old embed code!

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                debug.log("is following checked failed");
+                
+                if(d.status == 404) {
+                    debug.log(uName+" is not following "+chan);
+                }
+            });
+        }
+    }
+
+    var followcurrentchannel = function() {
```
night: Bad naming.

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+            Twitch.api.put("/kraken/users/"+uName+"/follows/channels/"+chan)
+            .done(function(d){
+                debug.log(uName+" is now following "+chan);
+                $("#followbtn").hide();
+            })
+            .fail(function(d){
+                debug.log(uName+" follow "+chan+" failed");
+            });
+        } else {
+            alert("You need to log in first!");
```
night: Pop up the login form instead!

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                if(d.status == 404) {
+                    debug.log(uName+" is not following "+chan);
+                }
+            });
+        }
+    }
+
+    var followcurrentchannel = function() {
+        if(typeof uName !== "undefined") {
+            Twitch.api.put("/kraken/users/"+uName+"/follows/channels/"+chan)
```
night: Twitch.api comment.

src/features/team-load-team.js
```diff
@@ -0,0 +1,91 @@
+var debug = require('../debug'),
+    vars = require('../vars'),
+    loadChannel = require('./team-load-channel');
+
+var loadTeam = module.exports = function() {
+    debug.log("Loading team data");
+    var theTeam = (window.location.pathname).replace("/team/", "");
+    theTeam = theTeam.replace("/event/", "");
+
+    Twitch.api.get("/api/team/"+theTeam+"/all_channels.json")
```
night: Twitch.api comment

src/features/team-load-team.js
```diff
@@ -0,0 +1,91 @@
...
+    .done(function(d) {
+        vars.jsnTeam = d.channels;
+        createButtons();
+        setTimeout(loadTeam, 60000);
+        
+        if(vars.teamFirstLoad == 1) {
```
night: ===

betterttv.js
```diff
@@ -4056,7 +4036,389 @@ module.exports = function () {
...
+            $("#followTeamBtn").css({"background-color":"#B9A3E3"});
+            $("#followTeamBtn").text("Following "+targetChan+" ...");
+
+            Twitch.api.put("/kraken/users/"+uName+"/follows/channels/"+targetChan)
+            .done(function(d) {
+                debug.log("follow success for:"+targetChan);
+                $("#followTeamBtn").css({"background-color":"green"});
+                $("#followTeamBtn").text("Followed "+targetChan);
+                
+                if(targetChan == vars.jsnTeam[vars.jsnTeam.length - 1].channel.name) {
```
night: ===

betterttv.js
```diff
@@ -4056,7 +4036,389 @@ module.exports = function () {
...
+    Twitch.api.get("/api/channels/"+chan+"/product")
+    .done(function(d) {
+        debug.log(chan+" has subs:"+d.price);
+        addStoof(d.price);
+        checkIsSubbed();
+    })
+    .fail(function(d) {
+        debug.log(chan+" subs check failed");
+        
+        if(d.status == 404) {
```
night: ``===``

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+    followTeamButton.click(function(e) {
+        followTeam();
+    });
+    $("#team_logo").after(followTeamButton);
+
+    var newDiv = $("<div>", {"id":"team_chat"});
+    $(".wrapper.c12.clearfix").append(newDiv);
+
+    $(document).click(function(e) {
+        if(e.target.id =="sharebtn") {
```
night: ``===``

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
...
+            $("#followTeamBtn").css({"background-color":"#B9A3E3"});
+            $("#followTeamBtn").text("Following "+targetChan+" ...");
+
+            Twitch.api.put("/kraken/users/"+uName+"/follows/channels/"+targetChan)
+            .done(function(d) {
+                debug.log("follow success for:"+targetChan);
+                $("#followTeamBtn").css({"background-color":"green"});
+                $("#followTeamBtn").text("Followed "+targetChan);
+                
+                if(targetChan == vars.jsnTeam[vars.jsnTeam.length - 1].channel.name) {
```
night: ``===``

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+    Twitch.api.get("/api/channels/"+chan+"/product")
+    .done(function(d) {
+        debug.log(chan+" has subs:"+d.price);
+        addStoof(d.price);
+        checkIsSubbed();
+    })
+    .fail(function(d) {
+        debug.log(chan+" subs check failed");
+        
+        if(d.status == 404) {
```
night: ``===``

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+        if(typeof uName !== "undefined") {
+            Twitch.api.get("/kraken/users/"+uName+"/follows/channels/"+chan)
+            .done(function(d) {
+                debug.log(uName+" is following "+chan);
+                $("#followbtn").hide();
+            })
+            .fail(function(d) {
+                debug.log("is following checked failed");
+                
+                if(d.status == 404) {
```
night: ``===``

src/features/team-load-team.js
```diff
@@ -0,0 +1,91 @@
...
+                chanGame = a.channel.meta_game,
+                chanStatus = a.channel.status,
+                chanViewers = a.channel.current_viewers,
+                newDiv = $("<div>", {"id":"channel_"+chanName, "class":"member", "title":dispName+" is offline"});
+            
+            newDiv.click(function(e) {
+                loadChannel(chanName); 
+            });
+            
+            if(chanStatus == "live") {
```
night: ``===``

src/features/team-load-team.js
```diff
@@ -0,0 +1,91 @@
...
+                    },
+                    position:{
+                        my:"left top",
+                        at:"right top"
+                    },
+                    content: function() { return $(this).attr("title"); }
+                });
+            }
+            
+            if(vars.teamCurrentChannel == chanName) {
```
night: ``===``

src/features/team-load-team.js
```diff
@@ -0,0 +1,91 @@
...
+                countImg = $("<img>", {"src":"http://www-cdn.jtvnw.net/images/xarth/g/g16_live_viewers.png", "class":"viewers_icon"}),
+                chanInfoHolder = $("<div>", {"class":"chanBtnChanInfoHolder"}),
+                chanImg = $("<img>", {"src":chanImgUrl, "class":"chanBtnChanImg"}),
+                chanNameSpan = $("<span>", {"class":"member_name"});
+            
+            countSpan.text(chanViewers);
+            chanNameSpan.text(dispName);
+            chanInfoHolder.append(chanImg, chanNameSpan);
+
+            if(chanStatus == "live") {
```
night: ``===``

src/features/team-format-page.js
```diff
@@ -0,0 +1,132 @@
+var debug = require("../debug"),
+    vars = require("../vars"),
+    loadTeam = require("./team-load-team");
+
+module.exports = function () {
+
+    if(bttv.settings.get("formatTeamPage") !== true || $("body").attr("data-page") != "teams#show") return;
```
night: ``!==``

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+        addStoof(0);
+    });
+
+    var uName = cookie.get("login");
+
+    var checkIsSubbed = function() {
+        if(typeof uName !== "undefined") {
+            Twitch.api.get("/api/users/"+uName+"/tickets?channel="+chan)
+            .done(function(d) {
+                if((d.tickets).length != 0) {
```
night: ``!==``

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                
+                followBtn.click(function(e) {
+                    followcurrentchannel();
+                });
+                chanActions.append(followBtn);
+
+                var shareBtn   = $("<div>", {"id":"sharebtn", "class":"button action primary"});
+                shareBtn.text("Share");
+
+                if(val != 0) {
```
night: ``!==``

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                }
+
+                $("#stats_and_description").append(chanLinkSpan, " playing ", jsnTeam[i].channel.meta_game, "<br>", viewerCount, viewsCount, followersCount, chanActions, descHolder);
+                $(window).resize();
+                break;
+            }
+        }
+        
+        if(typeof uName !== "undefined") {
+            Twitch.api.get("/kraken/users/"+uName+"/follows/channels/"+chan)
```
night: Twitch.api

src/features/team-format-page.js
NoxiusInfernus: These lines kill the Twitch timers to update the member list, and viewer count below the player when a channel is loaded, etc. If allowed to fire, they make the team list revert to the "pages" style and generally cause issues with my layout.

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                viewerCount.text(jsnTeam[i].channel.current_viewers);
+
+                var viewsCount = $("<span>", {"id":"views_count", "class":"stat"});
+                viewsCount.text(jsnTeam[i].channel.total_views+" ");
+
+                var followersCount = $("<span>", {"id":"followers_count", "class":"stat"});
+                followersCount.text(jsnTeam[i].channel.followers_count);
+
+                var descHolder = $("<div>", {"id":"description"});
+                descHolder.html("<b>Channel Description:</b><br>"+jsnTeam[i].channel.description+"<br><br><b>Broadcast Title:</b><br>"+jsnTeam[i].channel.title);
```
night: HTML should be placed in jade templates and imported: https://github.com/night/BetterTTV/blob/master/src/templates/setting-switch.jade

src/features/team-load-channel.js
```diff
@@ -0,0 +1,156 @@
...
+                    subTxt.text("Subscribe");
+                    subPrice.text(val);
+                    
+                    subBtn.append(subTxt, subPrice);
+                    chanActions.append(subBtn, shareBtn);
+                } else {
+                    chanActions.append(shareBtn);
+                }
+
+                $("#stats_and_description").append(chanLinkSpan, " playing ", jsnTeam[i].channel.meta_game, "<br>", viewerCount, viewsCount, followersCount, chanActions, descHolder);
```
night: HTML should be placed in jade templates and imported: https://github.com/night/BetterTTV/blob/master/src/templates/setting-switch.jade

src/features/team-load-team.js
```diff
@@ -0,0 +1,91 @@
...
+            
+            newDiv.click(function(e) {
+                loadChannel(chanName); 
+            });
+            
+            if(chanStatus == "live") {
+                newDiv.addClass("live");
+
+                var ttTime = new Date().getTime(),
+                    ttImgUrl = "http://static-cdn.jtvnw.net/previews-ttv/live_user_"+chanName+"-320x200.jpg?"+ttTime;
```
night: This doesn't support ssl going forward, use ``//``

Additionally, should we really be hardcoding links? That's a bad idea.

style/stylesheets/betterttv-team-page-dark.css
```diff
@@ -0,0 +1,256 @@
+body{
+    color: #9A9A9A;
+    background-image: none;
+    background-color: #141414;
+    //background-color: red;
```
night: Why?

src/features/team-load-channel.js
NoxiusInfernus: Holy crap, I'm so sorry for causing you to spend this much time holding my hand. But again, I would like to thank you for your patience in explaining all of this to me.

I will work on this rather long list (which I'm sure will grown longer hehe. sorry!) over the next few days, and get a new commit up ASAP.

style/stylesheets/betterttv-team-page-dark.css
```diff
@@ -0,0 +1,256 @@
+body{
```
night: Fix the formatting in this file please.

``name{`` to ``name {``

night:
Final Notes:

1. Please review all comments on your code, and apply fixes.
2. I suggest making sure you follow naming schemes we employ already in BetterTTV the best you can.
3. You removed all of your comments with your latest pull request. While we don't comment our source a lot, we really should. I recommend keeping your comments.
4. We have been trying to modularize code. I suggest moving your files in features to features/team/ and using functions/methods as much as possible.
5. Please migrate as much styling to CSS as possible, as it seems a lot is still handled in JS. I also would like to see Jade templates with variables for as much HTML as possible. Since our gulp isn't setup for subdirectories in the templates folder, I might recommend prefixing your templates with ``teams_`` if there are many.
6. Make the use of timeouts and timers at a minimum, as they can lead to unintentional results. You should use JavaScript callbacks instead, as often as possible (this comes back to using functions/methods).

night:
Updates @NoxiusInfernus ?

NoxiusInfernus:
Yeah sorry man, had some personal issues that kept me from getting back to this as quickly as I wanted to. But I have made a lot of progress over the last few days, so I'll clean up what I have and push a commit to my fork tonight/tomorrow morning after I get up (I need to sleep lol).

This commit will be heavily commented both for clarity of what I'm doing, and in terms of asking for your preference in specific places. I figure that way you can comment at your convenience while I'm still working on finishing up the changes you have noted so far. I think the only really "big" thing  on my todo are the jade templates... Which I've never used or even heard of before lol, so it will likely take a bit to figure out how to use the templates (making them seems straight forward though).

src/features/team-load-team.js
NoxiusInfernus: No problem on using //

As far as hardcoding the preview image link: The team API doesn't return preview images in the channel objects. So instead of making an additional API call for each preview image I thought it would just be better to use the URL because it's defined in the streams API as a template URL anyways.

But I can make a call to /streams for the preview URL if you think it would be better.

night: If it's not included in the call, then I guess just use a hardcoded link for now.

NoxiusInfernus:
Crap, I forgot I had experimented with loading a jqueryui lib from features/team. That was not meant to be part of the last commit. Will be fixed (removed) along with some other things in the next commit

night:
So how's it going?

NoxiusInfernus:
Been waiting for input.

betterttv.js
```diff
@@ -4098,7 +4078,449 @@ module.exports = function () {
...
+        
+        if(bttv.settings.get("darkenedMode") === true) {
+            teamCSS = $("<link>", {"href":"//cdn.betterttv.net/style/stylesheets/betterttv-team-page-dark.css?"+bttv.info.versionString(), "id":"betterTwitchTeams", "rel":"stylesheet", "type":"text/css"});
+        }
+        $("body").append(teamCSS);
+        
+        getFeatured();
+        formatTeamPage();
+        //if it wasn't for the bttv.TwitchAPI._ref is null error in Firefox, I would just load the team without the timeout
+        setTimeout(loadTeam, 1000);
```
night: Introduces a race condition on slower computers. Find the fault upstream instead.

betterttv.js
```diff
@@ -4098,7 +4078,449 @@ module.exports = function () {
...
+    }
+    
+    var getFeatured = function() {
+        bttv.TwitchAPI.get("/kraken/streams/featured?limit=1")
+        .done(function(d) {
+            //debug.log("got preview url template on load");
+            var chanName = d.featured[0].stream.channel.name;
+            var updatedPreviewUrl = d.featured[0].stream.preview;
+            vars.updatedPreviewUrl = updatedPreviewUrl.replace(chanName, "CHANNEL");
```
night: ...?

betterttv.js
```diff
@@ -4098,7 +4078,449 @@ module.exports = function () {
...
+                
+                theButtonArray["isPlaying"] = true;
+            }
+            
+            if(chanStatus === "live") {
+                var ttTime = new Date().getTime(),
+                    ttImgUrl = "//static-cdn.jtvnw.net/previews-ttv/live_user_"+chanName+"-320x200.jpg?"+ttTime;
+                    
+                if(typeof vars.updatedPreviewUrl !== "undefined") {
+                    ttImgUrl = vars.updatedPreviewUrl.replace("CHANNEL", chanName) + "?"+ttTime;
```
night: ..?

You're carrying something from a different file over here through vars? Lets not do that..

betterttv.js
NoxiusInfernus: This was added because you didn't like the idea of hardcoding the URL to the stream preview images (which are not included in the Team endpoint). So I make a call to streams/featured to get the preview URL.

night:
I glanced at it a bit, but don't have time to really sift through it all now. The big things I've noticed are that you're not really coding this in a object oriented way, which needs to be changed. Ideally this is a module that needs to expose hooks for BetterTTV to call, and everything else needs to be internalized nicely.

The use of the old vars global is scary, and really shouldn't be used at all. It's only used in some places in BetterTTV, and is legacy code.

We use browserify in order to allow us to use nodejs-esque requires. Check out other modules under src/features/ to see how they work. Ideally on page load the settings loader would call your module's external hooks to toggle the feature on and off, and everything else should be encapsulated to your module's eyes only.

Where I've commented that you're carrying something from a different file through vars is something you need to deal with differently. This is all surrounding the fact that it seems your module lacks organization. You really need to make a master object that does all the leg lifting for your different files, and handles transport of info between files without the use of our legacy code.

betterttv.js
night: If that's the functionality, I suggest changing to instead pull from ``kraken/streams?channel=channel1,channel2,channel3``. This might make more sense than replacing a channel name in the url?

There's also an odd edge case to doing what your doing right now. What if the user's username is in the URL twice?

NoxiusInfernus: Because there's no guarantee that any supplied channel name(s) will be online (the streams endpoint only returns useful data for online channels). That's why I went with featured, because there's always at least 1 featured stream. Also, the channels endpoint does not include a preview image URL.

As far as the channel name appearing twice, it doesn't happen. It's a templated URL... Please refer to https://github.com/justintv/Twitch-API/blob/master/v3_resources/streams.md#get-streamsfeatured. Granted, I could have used the actual template, but I was trying to save some steps by only replacing the channel name instead of replacing the channel name and dimensions.

NoxiusInfernus: Again, this was done to avoid making an additional API call for each channel on the team simply for a stream preview image. They all share the same URL template, so I was attempting to make sure that I was grabbing an updated template only as needed.

night: > As far as the channel name appearing twice, it doesn't happen. It's a templated URL... Please refer to https://github.com/justintv/Twitch-API/blob/master/v3_resources/streams.md#get-streamsfeatured. Granted, I could have used the actual template, but I was trying to save some steps by only replacing the channel name instead of replacing the channel name and dimensions.

Your assumption is not a good one. Lets say I own the username ```preview``` on Twitch. In your code, the template URL becomes ```http://static-cdn.jtvnw.net/CHANNEL-ttv/live_user_preview-{width}x{height}.jpg```. Additionally, I think mustache tags might work better than ```CHANNEL```, but since the URL is guaranteed to be lowercase at this point it doesn't matter.

Someone own's the username ```null``` on Twitch, so I wouldn't say that this would never happen.

NoxiusInfernus: Fair point about a channel named "preview". I will just burn another API call for the preview URLs.

But I really don't see how a channel named "null" plays into this? It's moot at this point since I'll make another API call for each preview image, but I am just trying to understand the point you are trying to make here? If a channel name was "null", it would only be in the URL once, and the replace would work as expected... Unless I'm missing something (which I probably am hehe)?

night: I think the appropriate solution here is to replace:

```live_user_username-{width}x{height}``` with ```live_user_{user}-{width}x{height}```

betterttv.js
NoxiusInfernus: Is there any reason why you inject the emote script on pages that do not have chat? IE: Current team pages, messages, etc? Because every time that happens, I notice the error (except that it's now: "e.TwitchAPI._ref is null", instead of "bttv.TwitchAPI_ref is null") in my console on Firefox, right after the "Injecting Twitch Chat Emotes Script" debug message. This also causes a huge Angular error (referencing bttvEmotes) to follow shortly after. I didn't include the error message because of long it is (but will upon request).

I would like to note, this happens even when I'm NOT using my local files for BTTV.

NoxiusInfernus: Ok, an update. This is/was proving to be a pain to hunt down in Firefox hehe. I commented out all of my code, and that null error still kept popping up... Disabled all emote options, and the error was still there...

It seems like the problem keeps coming back to using bttv.TwitchAPI.get() too quickly during page load in Firefox. Even your functions for grabbing the channel info or following list it seems. If I comment out the functions to load the following list (check-following), channel data (check-broadcast-info), and remove all all of my code, the null error is gone.

But as soon as I "uncomment" any function using bttv.TwitchAPI.get() at start up (check-following for example), the null error returns. Are you seeing this type of behavior under Firefox too?

NoxiusInfernus:
Ok, I'm gonna table the whole null error thing for now because I'm getting no where fast and it's driving me crazy hehe. I'll come back to it, but I think I have bigger problems overall rooting from my organization (or lack thereof hehe).

As far as lacking organization, I would have to agree. The only reason I'm using 3 files now is because I had figured that I should put each major "step" of my script in it's own file when I had first looked into the features directory. I could certainly just combine all 3 of my files into 1 file, but I don't think that's what you meant? Could you elaborate a little on how you would have set this up? Just generally, for example: 1 main file, 1 file to do this, 1 file to do that, etc? Even with looking at the other modules in features, or features/channel-reformat, I'm just really confused.

Which also brings me to another point: I'm a handle jiggler lol. If you would rather not try to work with me on this, I would totally understand hehe. I wouldn't take offense, because let's face it, I'm not on your level lol. But if you have the patience, I'll keep jiggling the handles until I (hopefully) get it right. I'm just giving you an easy out if you don't think I'm up to the task hehe :)

night:
> Which also brings me to another point: I'm a handle jiggler lol. If you would rather not try to work with me on this, I would totally understand hehe. I wouldn't take offense, because let's face it, I'm not on your level lol. But if you have the patience, I'll keep jiggling the handles until I (hopefully) get it right. I'm just giving you an easy out if you don't think I'm up to the task hehe :)

I assumed your goal here was to improve your own skills. If that's still the case, you should consider googling object oriented programming. There's a plethora of information online if you search for it. Self learning through practice is how you can really improve your skills. For a quick example of what I'm looking for, you can refer to my TwitchStatus code: https://github.com/night/TwitchStatus (please note that this is probably not the best example as I do not rank myself to be that great at programming either). In my code, you can see twitchstatus.js pulls together all of the other resources and shares common data amonst them. This is the type of module I'm looking for from you.

NoxiusInfernus:
Oh yes, learning and improving are always great :D I just wasn't sure if the time and effort (on your part) was becoming too much hehe. I'm gonna be tied up until this weekend for the most part, but I'll use these next few days to try to read and digest as much as I can about OOP.

betterttv.js
NoxiusInfernus: Ok, back to the TwitchAPI._ref is null error hehe... Again, this is specific to Firefox and team pages on Twitch.

While using the published version of BTTV in Firefox, if I visit any team page, I get the "e.TwitchAPI._ref is null" type error. So it's not something that I'm doing per say that's causing the error, all I know is that any function that makes a call to bttv.TwitchAPI.get within the first second or so of a team page loading in Firefox will cause it.

When I test using a local version of BTTV, the error is still present under the same conditions in Firefox, but it becomes "bttv.TwitchAPI._ref is null". The only ways I know to get around it are to either delay the use of bttv.TwitchAPI.get for a second or so, ooor...

If I comment out line 30 from the "_untakeover" function in src/twitch-api (this._ref = null) then everything *appears* to work fine. I am not suggesting that this is a fix (I know it's not), I'm only making the observation because I honestly don't why it's happening. I get the gist of what's going on in twitch-api, but not enough to be able to find the cause or resolve the issue. I was hoping maybe you might have some input @night ?

PR #207, opened 2015-02-02
changed comma after name completion to editable setting

spartam:
Fixed a small issue to get familiar with this project.
I expect to contribute more.

bttv setting tab complete postfix setting
new option in the chat setting to change postfix


teak:
The word case is wrong.. http://s.teaksoftware.com/i/s_nU60NYgq.png

spartam:
My apologies, I fixed the word case

DTittel:
This is a minor binary setting it doesn't belong in the direct chat
settings it should be in the BTTV settings with all the other binary
settings.

spartam:
I agree it could be made a binary setting and I'll make it a binary setting if that's preferred the most. I extented it allow users to change the postfix to whatever they want instead of the requested binary setting of having the ", " added or not

DTittel:
I see OK sorry 4:30 am just saw this on my phone

night:
We really need to refactor how settings are managed. There's far too many settings in the list already. Ideally settings need to be grouped under categories.

vendethiel:
> We really need to refactor how settings are managed. There's far too many settings in the list already. Ideally settings need to be grouped under categories.


Please don't forget about

spartam:
> We really need to refactor how settings are managed. There's far too many settings in the list already. Ideally settings need to be grouped under categories.

Do you want to refactor both the settings in the chat and the bttv dashboard or just one of them?
Do you have an idea of what you want it to look like?

we can change the settings to have tabs per categorie and move the chat settings to a "chat" tab on the bttv dashboard settings page

teak:
I could wip up an nice settings interface, including sections, string editors, and channel specific settings. The important part is someone would have to write a proper settings API for it..

spartam:
Can you give more details about what this API would be required to do and what would be kept in mind when writing it?

vendethiel:
I have a JSON storage API in another PR, if that's what you need, I'll just rebase it.

PR #220, opened 2015-02-11
Non-legacy Tab completion cycling

voloshink:
Tab to cycle through the suggestions

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -948,7 +948,7 @@ bttv.chat = {
 
                 var $suggestions = $chatInterface.find('.suggestions');
                 if($suggestions.length) {
-                    $suggestions.find('.highlighted').find('.suggestion').eq(0).click();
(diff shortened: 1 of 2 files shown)


vendethiel:
this is not the correct file to change – you're supposed to change files in the `src/` directory.

for the PR: is legacy tab completion cycling working? If so, then there's definitely some duplication here

@night Should we add a "GENERATED" warning header in the compiled .js file?

voloshink:
Moved changes to the correct file.

voloshink:
Yes, legacy tab completion also allows you to cycle through suggestions but the legacy and non legacy completion do it in different ways. This change lets the user cycle through the suggestions pop up after they enter @. It's a quality of life change. I'm sure both tab completions can be merged at some point.

night:
Not needed any longer.

PR #230, opened 2015-02-15
added remove subscriber button option

cenderin:
I don't always like to stare at a subscribe button when im watching, so I added this.  You can add it if you want or let me know if you want something changed/added to make it more useful.


vendethiel:
why not just go theater mode?

cenderin:
I guess thats kinda true actually, never really used it before

PR #236, opened 2015-02-18
Preserve emoticons in copy/paste

### LINKED ISSUE #234: Copy emoticons in text selection to clipboard
When an emoticon is among the text in a selection, the text to produce the emoticon should be substituted for the emoticon picture in the clipboard when the selection is copied.

That way we could do copy pastas that have emoticons in them.

acrisci:
Add an event listener to the "copy" event that fires when text is copied
to the clipboard to convert any emoticons within the text selection to
the corresponding text that will create the emoticon. This allows
copy/pasted text to preserve emoticons within the Twitch chat.

fixes #234

### DIFF
--- a/src/features/emoticon-text-in-clipboard.js
+++ b/src/features/emoticon-text-in-clipboard.js
@@ -1,0 +1,46 @@
+// Add an event listener to the "copy" event that fires when text is copied to
+// the clipboard to convert any emoticons within the text selection to the
+// corresponding text that will create the emoticon. This allows copy/pasted
+// text to preserve emoticons within the Twitch chat (issue #234).
+module.exports = function() {
+    if (!('oncopy' in document)) {
+        // Copy event is not supported.
+        return;
+    }
+
+    var onCopy = function(e) {
+        if (!e.clipboardData || !e.clipboardData.setData) {
+            // Setting clipboard data is not possible. This is not currently
+            // possible to detect without actually firing a real copy event.
+            document.removeEventListener('copy', onCopy);
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


acrisci:
I'm testing this with Chromium 40 and it seems to work ok for me.

There could be an issue with extra spaces sometimes being added to the clipboard because of the conversion of the html fragment to text. Let me know if you want me to look into fixing that.

I can also do any cleanup you want, such as moving the event handler to a different part of the code.

night:
This shouldn't be placed in main.js

Furthermore, your PR breaks on browsers without the new clipboardData API. Not everyone runs the latest version of a browser.

acrisci:
Hey, thanks for the quick reply.

I tested adding the "copy" event on a browser that does not support the clipboard api, and I think the worst thing that could happen is that the even simply won't fire. I don't see anywhere the feature will break unsupported browsers. There will not be a way to support browsers that do not support the clipboard api yet.

As for the location of the code, let me know if you have a preference. Otherwise, I will try to place it somewhere else that it makes sense and see if that is better for you.

night:
You should place this in its own feature file (see https://github.com/night/BetterTTV/tree/master/src/features) and call the feature from main. We will eventually rid all of the mess in main.js to their own feature files.

Inside the feature file, you should return early should the browser a client is using be incompatible with the clipboardData API.

acrisci:
Hey @night, I made the changes you requested.

It's a little difficult to detect whether or not the browser supports the clipboard api (see [here](https://github.com/Modernizr/Modernizr/pull/659)), but I did the best I could.

Issue #246, opened 2015-02-22
T-Shirt Merch Panel


Sleekness:
The T-Shirt Merch Panel on dark theme is fixed only on user channel, but not on the profile page.
User channel (fixed) - http://i.imgur.com/d45bdR6.png
User profile page (not fixed) - http://i.imgur.com/2gCOvPs.png
@matthewjk123 or whoever was fixing this panel

matthewjk123:
@Sleekness Thanks! I'll make the change right now! :D

vendethiel:
247

teak:
Can some please fix the miss aligned bar at the bottom :P

PR #249, opened 2015-02-23
fixed progress-bar width so it aligns correctly

cenderin:
Teak commented about this on [#246](https://github.com/night/BetterTTV/issues/246#issuecomment-75452070)

### DIFF
--- a/style/stylesheets/betterttv-dark.css
+++ b/style/stylesheets/betterttv-dark.css
@@ -250,6 +250,9 @@ body {
     height: 12px;
     margin-bottom: 6px;
 }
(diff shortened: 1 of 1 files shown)


DTittel:
Lol seems like this patch is more about removing trailing whitespace than the progress bar width. Code looks fine but haven't tested it out.

cenderin:
yea... It's automatic normally, maybe I should turn it off when i'm doing stuff like this.

teak:
Tested. Looks good..

vendethiel:
You can just append ?w=1 after the URL to get a whitespace-free changeset.

Issue #252, opened 2015-02-24
Autocomplete Issues (CTRL+TAB)


ryanbraganza:
https://community.nightdev.com/t/autocomplete-issues-ctrl-tab/3163

While using firefox I have the habit of using CTRL+TAB to cycle through tabs, but recently this problem started to haunt me: If I CTRL+TAB out of a Twitch Chat it still acts as if I simply pressed TAB, even if I have not entered a single letter, cycling through namesat the same time I leave the tab. Could this be corrected so it ONLY cycles if you use TAB or SHIFT+TAB?

night:
A pull request was accepted that improves the functionality of the tab completion, and allows you to disable the visual suggestions that popup. The feature will be pushed live probably later today.

PR #272, opened 2015-03-04
refactor bttv.chat-related code

vendethiel:
Refactor a big chunk of main.js' "bttv.chat" code.

This should enable future refactorings and allow to untangle the code a lot.

PR #274, opened 2015-03-09
Don't autocomplete when ctrl pressed

ryanbraganza:
Autocomplete shouldn't fire when a user is ctrl-tabbing through their
tabs.

This is the change I wanted for #252

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -989,7 +989,7 @@ bttv.chat = {
             }
 
             // Actual tabs must be captured on keydown
-            if(e.which === keyCodes.Tab) {
+            if(e.which === keyCodes.Tab && !e.ctrlKey) {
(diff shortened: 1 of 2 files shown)


vendethiel:
Could probably check for meta and such. Although I'd prefer to wait for my #272 PR to get merged before this one, because this sort of things will become easier.

ryanbraganza:
I hesitate to check for any other modifiers, since the only behavior I want to change is when ctrl is pressed. Unless you have other reasons to do so?

Since it's only a one line change, I'd prefer to merge this before #272 since, due to its size, it could take a long time to review, and the current behavior is a little annoying.

vendethiel:
It seems the browser behavior is just dismissed for meta-tab anyways. Merge anytime

PR #310, opened 2015-03-29
Added uptime command: Fixes #245

### LINKED ISSUE #245: Add an /uptime command
When `/uptime` is entered into chat, it should show how long the current stream has been online.

amamut:
Created `/uptime` command which displays the amount of time the stream has been up.

Uptime


amamut:
Updated. Added check to make sure `created_at` isn't null.

PR #311, opened 2015-03-29
Added mouseover image preview for chat: Fixes #131

### LINKED ISSUE #131: show imgur pics when hover over the link
When someone adds an Imgur to chat you should be able to hover over it so the pic shows up in a little popup window for the time that you're hovering over it. It would save a lot of people a lot of time! :) :+1:

amamut:
Added ability to enable and disable chat image preview within the chat. Preview opens next to mouse cursor and closes when moused off. Works for any link which ends in .jpg, .jpeg, .gif, and .png.

Preview:
preview

Config screen:
config

### DIFF
--- a/src/features/image-preview.js
+++ b/src/features/image-preview.js
@@ -8,23 +8,23 @@ var enablePreview = exports.enablePreview = function() {
     /* END CONFIG */
     $(document).on({
         mouseenter: function (e) {
-            $("body").append("<p id='preview'><img height=\"200px\" src='"+ this.href +"' alt='Image preview' /></p>");
-            $("#preview")
+            $("body").append("<p id='chat_preview'><img width=\"250px\" src='"+ this.href +"' alt='Image preview' /></p>");
+            $("#chat_preview")
                 .css("top",(e.pageY - yOffset) + "px")
                 .css("left", (e.pageX - xOffset) + "px")
                 .css("position", "absolute")
                 .css("z-index", '100')
                 .fadeIn("fast");
         }, mouseleave: function (e) {
-            $("#preview").remove();
+            $("#chat_preview").remove();
         }, mousemove: function (e) {
-            $("#preview")
+            $("#chat_preview")
             .css("top",(e.pageY - yOffset) + "px")
             .css("left",(e.pageX + xOffset) + "px");
         }
-    }, 'a.preview');
+    }, 'a.chat-preview');
 };
 
 var disablePreview = exports.disablePreview = function() {
-    $(document).off('mouseenter mouseleave mousemove', 'a.preview');
+    $(document).off('mouseenter mouseleave mousemove', 'a.chat-preview');
 };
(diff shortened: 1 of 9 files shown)


src/features/image-preview.js
```diff
@@ -0,0 +1,30 @@
+var debug = require('../helpers/debug');
+
+var enablePreview = exports.enablePreview = function() {
+    /* CONFIG */
+    xOffset = -250;
```
vendethiel: this should probably be lifted off this function, and into the file scope

amamut: I added "var" but I am not sure what you mean by the first part. Can you explain?

Thanks.

teak:
I this should probably be in the main bttv settings panel..

Also I dont really like use of such a general id ('#preview').

src/features/image-preview.js
vendethiel: move it outside of `enablePreview`

amamut:
@teak No problem.

amamut:
@teak Should be good now. Let me know if you see any other issues.

night:
I have reviewed the pull request, and the logic is fine.

Unfortunately, however, the most popular image sharing site, Imgur, 302 redirects all direct image links from Twitch since there's a referrer header. A solution would need to be devised as a way to fix this, like a serverside redirector to remove referrer from redirections.

Additionally, I'd like to see support for all imgur content, not just images (videos, galleries) (again, since Imgur is the de-facto image sharing site at the present).

amamut:
@night I just tested it and both http://i.imgur.com/KkLzywe.png and http://imgur.com/KkLzywe.png work just fine. Do you have an example where it doesn't work?

As for the other content types maybe that can be a separated out into another ticket?

night:
Your browser is reading from cache when you paste an image you've already visited.

amamut:
You're right. Ughhh, that sucks!

night:
Have you tried experimenting with iframes? I think maybe an iframe + embedding an image with width/height 100% would work, as long as imgur isn't redirecting the referring host (in which case they only redirect from sites like Twitch, Youtube, and other social networks).

This wouldn't work for other image hosts that block referrers, but there's nothing we could do short of reverse proxying the images (which is too costly to invest into).

I made a BetterTTV API endpoint that embeds an iframe. It requires a referral from Twitch as well as a valid image URL. Please let me know how it works:

    http://api.betterttv.net/2/image_embed/ + encodeURIComponent(url)

amamut:
Thanks I'll take a look into it. I've been really busy lately and haven't had time to experiment much.

amamut:
@night I tried it out and it was working for a while. But now all images return 404. Also it was a bit slow but loaded within 5 seconds.

night:
Can you explain what exactly is returning a 404 error? I cannot reproduce.

amamut:
Sorry that was my bad. I had an extra " in the src and it wasn't sending the proper request. Looks good now. Let me know.

goldbattle:
Want to say this is an amazing feature, thanks!

PR #316, opened 2015-03-31
Added timestamps to notification messages: Fixes #291

### LINKED ISSUE #291: time stamps for notifications
Add time stamps for chat notifications when time stamps are enabled in twitch settings like:
- welcome message
- subscription
- host started
- host stopped
- timeout
- ?

amamut:
Notification messages are now tied to the time stamp option in the settings panel. Also added an added override within to code to force time stamps to be disabled because sometimes you don't want them. (e.g. BTTV version load message).

timestamp message

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -578,7 +578,7 @@ var privmsg = exports.privmsg = function (channel, data) {
             vars.userData.isLoggedIn ? helpers.isModerator(vars.userData.login) : false,
             {
                 message: data.message,
-                time: data.date.toLocaleTimeString().replace(/^(\d{0,2}):(\d{0,2}):(.*)$/i, '$1:$2'),
+                time: data.date == null ? '' : data.date.toLocaleTimeString().replace(/^(\d{0,2}):(\d{0,2}):(.*)$/i, '$1:$2'),
                 nickname: data.from || 'jtv',
                 sender: data.from,
                 badges: data.badges || (data.from==='twitchnotify'?[{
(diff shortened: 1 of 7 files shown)


night:
I'm not liking this implementation, as the extra argument to toggle timestamps off seems superfluous. ```date``` is already an object key, so you should be setting that to null if date isn't wanted.

amamut:
@night With the current implementation all server messages hide the timestamp. I thought that the "loaded" messages didn't really need a timestamp and looked weird with it. So I added a way to suppress the timestamp. This had minimal impact on existing code. I can make a separate server message call which will suppress timestamps if you'd like. Or just remove the override all together.

night:
I am all for hiding timestamps for messages like the loaded message, but I do not think there should be an additional argument added to the function.

The code should be modified to instead show dates only when a date is seen in the message object. You can then append a message that is printed with a hide timestamps class if the date isn't in the message object.

This seems like a more correct way of doing things.

amamut:
@night I can make the change you requested but the reason why I did it this way is because I believe that the "BTTV Loaded" message is the only message which is formatted uniquely to where the time stamp looks bad. So it was easier to add a bool to that call instead of looking through all of the code and adding a Boolean to ALL serverMessage calls.

amamut:
I've made the updates but I've noticed that Twitch may have updated their chat because all BTTV chat commands come back with BTTV response AND another message saying "Unrecognized command". I will have to see if I can disable these messages.

night:
Those messages happen because BetterTTV passes through our custom commands to the Twitch chat server. Twitch used to not return errors when a command did not exist.

amamut:
@night Ok thanks for the heads up. I disabled internal bttv commands from being sent to the twitch chat servers.

PR #318, opened 2015-04-02
Added automatic theatre mode in settings panel: Fixes #117

### LINKED ISSUE #117: Automatic theater mode?
Would be great to add an option to automatically go in to theater mode when a stream is opened.

amamut:
Added automatic theatre mode in the settings panel.

auto theatre mode


night:
Your code would not function for channels joined from the directory (or basically any channel navigated through on Ember). You would need to call your toggle function when the page transitions to a channel page (code for this is in main.js).

amamut:
Updated. Thanks for finding that.

PR #396, opened 2015-05-01
Hides zalgo-text from chat

hansskogvold:
Disables zalgo-characters ( http://www.zalgotext.net/ ) going outside the .chat-line div.


hansskogvold:
Before: 
Before
After:
After

night:
Great idea and all, except this breaks emotes which happen to be over the bounds of the chat lines. This would need to be filtered before being printed to the screen.

PR #437, opened 2015-05-24
url encoding fix for image preview

BarryCarlyon:
Fix the image preview regext to support +
As in the other version of %20 the URL encode of space

--

I found that a URL Encoded image the encoded a space in the file name as + instead of %20 which is technically valid would not trigger the image preview feature. This fixes that.


BarryCarlyon:
I wasn't paying attention and my text editor has ripped all the spaces off the end of lines it seems…
My apologies

PR #484, opened 2015-06-11
Add whisper reply support and whisper reply template in chat history

EhsanKia:
The first commit appends "/w [user] " to the chat history, just like the default twitch chat does.
There's a tiny issue with the implementation, but I'm not sure what the best way to fix is.
The prefered behavior would be that your cursor automatically gets set at the end of the line.
All the solutions I have seem a bit too hacky, I'd like to hear from you guys.

The second commit adds support for "/r" to reply to the last whisper.
There's also a warning if your reply recipient changes within 3s of you replying.
The warning messages seem a bit weak and may need tweaking.

Any code review suggestions are welcomed.
As a side note, it seems like my editor stripped some whitespace, I hope that isn't an issue.

### DIFF
--- a/src/chat/handlers.js
+++ b/src/chat/handlers.js
@@ -100,6 +100,7 @@ var commands = exports.commands = function (input) {
         helpers.serverMessage("/suboff -- Shortcut for /subscribersoff");
         helpers.serverMessage("/t [username] [time in seconds] -- Shortcut for /timeout");
         helpers.serverMessage("/u [username] -- Shortcut for /unban");
+        helpers.serverMessage("/r [message] -- Shortcut for /w [last whisper sender] [message]");
         helpers.serverMessage("/uptime -- Retrieves the amount of time the channel has been live");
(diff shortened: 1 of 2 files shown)


EhsanKia:
I fixed the cursor placement, as well as improved the error messages. Also fixed a little bug.

night:
#486 was a better solution.

Why? Using a setTimeout to decide if a reply is to be sent to the wrong recipient is a bad assumption. This other solution replaces the text as you type, so you know exactly where your message is going.

As for implementing chat field history as Twitch does (using the up key in the textarea), we are not interested in supporting their unique solution at this time since we already store all previous messages. Pressing up is useful as it lets you know what you said, and since whispers do not get stored anywhere it might be useful in a fast chat to know what you've said and to whom. Additionally, the /r command should fulfill everyone's desire to easily reply.

EhsanKia:
I agree, the other implementation was the better UX experience.
It was still a good experience in getting to know the codebase anyways.
Thanks for the feedback.

PR #487, opened 2015-06-12
You tube links (chrome only)

OurFlagIsMined:
Shows the title of a YouTube video or playlist link when the mouse hovers over it.  There are two potential ways of doing it (the other is in a separate branch, the one that I didn't pull request, and includes more details in its commit notes).  This way gives permissions for YouTube URLs to the content script, which makes calls to the YouTube oembed API endpoint (which doesn't require authentication, and is compatible with their recent API changes) on behalf of the injected portion, but requires updating the betterttv.js and manifest.json files of the Chrome extension to work (and so wouldn't work with other browsers as it currently stands).  In short, the other way would be to use the currently-existing BTTV servers to lookup and cache YouTube titles (with some simple PHP and SQL or something similar) and have the client go through that, instead of directly calling the YouTube API.
temp-temp-temp-rger-crop

### DIFF
--- a/chrome/betterttv.js
+++ b/chrome/betterttv.js
@@ -1,0 +1,36 @@
+
+window.addEventListener('message', function(event) {
+	if (event.data.type === 'youtube-title-query') {
+			var query = 'https://www.youtube.com/oembed?url='+event.data.url+'&format=json';
+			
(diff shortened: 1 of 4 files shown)


night:
We do not make code changes which require a specific browser to be used, sorry.

Issue #88, opened 2014-12-02
Link Unshortening


night:
It would be useful to see the Top Level Domain that a shortened url is pointing to.

example view:

bit.ly/BLAH <google.com>

unshorten.me could be used for this, API has a 100k per IP/API_KEY per day limit.

Optionally a cached list on a night server to prevent 1000+ users in chat all unshortening a link at the same time?

amamut:
@night Why use a third party API when writing a quick GET method in node is trivial (node-curl). You can have it cache on your servers automatically instead of sending a second call. Just thinking out loud here.

vendethiel:
node-curl on bttv?

night:
I didn't make this request, it was just imported from Redmine. Due to CORS in browsers, we would have to use something serverside though.

amamut:
I meant using node-curl on a node js server. Or using curl on any back-end server.

metarmask:
> Due to CORS in browsers, we would have to use something serverside though.

Chrome has a way to make cross-orgin requests: [https://developer.chrome.com/extensions/xhr]("Cross-Origin XMLHttpRequest - Google Chrome")

PR #488, opened 2015-06-12
You tube links (middleman server required)

OurFlagIsMined:
Here's the other way, just so that both are in the same place.  This one has also been working perfectly in all my testing, but has the aforementioned cross-site Access-Control-Allow-Origin issue unless browser security is bypassed.  The requests could be routed through (and cached in the process) the BTTV servers instead of directly fetching the info from the YouTube oembed API (the oembed API is non-authenticated), or Night could create a YouTube API v3 key and see if it stays within the free quota (a few million requests per day from my understanding).

### DIFF
--- a/src/chat/templates.js
+++ b/src/chat/templates.js
@@ -24,12 +24,39 @@ var modicons = exports.modicons = function() {
 var escape = exports.escape = function(message) {
     return message.replace(/</g,'&lt;').replace(/>/g, '&gt;');
 };
+var linkifyCache = {};
 var linkify = exports.linkify = function(message) {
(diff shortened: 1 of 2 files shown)


night:
This is not really a decent solution to the problem at hand. Rather than parsing all youtube links on the fly, you would want to lazy load API information on hover. This ensures that a client isn't bombarding an API for every message containing a YouTube link, and makes it a lot more cost effective to run. If we implemented this feature, we would host on our API an endpoint for resolving appropriate metadata of links, even more than just YouTube (like #88).

Issue #464, opened 2015-06-05
Copy user's name to chat box by pressing rmb


alex-f-k:
Right mouse button, yeah.
Fast and easy way to answer somebody in chat.

Maxhype:
+1.

rtainc:
I agree there should be some shortcut, but wouldn't this be counterintuitive as RMB?

night:
This has been implemented as double-clicking in #500

PR #500, opened 2015-06-15
Add #464

Naththo:
#464 

I'm not quite sure about the double click though since it brings up the popup box upon clicking the users name which is why I used the line to begin with.

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -229,12 +229,21 @@ exports.attrs = function attrs(obj, terse){
  * @api private
  */
 
-exports.escape = function escape(html){
-  var result = String(html)
-    .replace(/&/g, '&amp;')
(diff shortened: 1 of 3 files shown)


OurFlagIsMined:
This is a feature I've wanted for a long time, and double-click feels like it would work great if you can get it to work well on the chat line. If you go with username click activation though, you might consider using control+click or mousewheel-click instead of double-click, since two clicks are harder than one on a small moving target (username hitbox versus entire chat line hitbox).

Naththo:
I had it bonded to the chat line before but @night said that it wasn't specific enoug .  I'll try a few things when I get home later to try make it easier.

night:
Looks good

Issue #290, opened 2015-03-18
Show Stream Title in Browser Title


RenaKunisaki:
Simple (maybe) request; set the page title to the stream title (maybe also keeping the "StreamerName - Twitch" part). Helpful for Theatre Mode (where stream title is hidden), history searching (usually you can search your browser's history by page title), etc.

PR #514, opened 2015-06-23
Update window title to reflect stream title

EhsanKia:
Very simple commit for #290, just a single line.
Feel free to change the formatting.

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -229,12 +229,21 @@ exports.attrs = function attrs(obj, terse){
  * @api private
  */
 
-exports.escape = function escape(html){
-  var result = String(html)
-    .replace(/&/g, '&amp;')
-    .replace(/</g, '&lt;')
-    .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;');
+var jade_encode_html_rules = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;'
+};
+var jade_match_html = /[&<>"]/g;
+
+function jade_encode_char(c) {
+  return jade_encode_html_rules[c] || c;
+}
+
+exports.escape = jade_escape;
+function jade_escape(html){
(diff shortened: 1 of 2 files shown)


night:
Please read the README.md before submitting a pull request.

Additionally, this is not a good solution. This should be a toggleable option, and you should be limiting the length of the title.

EhsanKia:
By option, do you mean full on button in the settings page? That seems overkill for such a small "feature". Or just a setting that can be toggled through console, similar to consoleLog?

night:
Because this impacts a small space's functionality and we don't want complaints, it needs a setting toggle.

OurFlagIsMined:
I agree that it needs to be a setting, since the clean interface and titles are part of what makes Twitch such a user friendly site; and in my opinion one that's disabled by default since it adds a lot of clutter to the title, and would make bookmarked streams save with inconsistent names.

I do think however that it would be welcomed by a lot of people if it did so for VODs, since each VOD is a unique bookmark, and would allow searching by stream title.

EhsanKia:
You guys make the call, and I'll code it. Honestly this was mostly me getting my feet wet by taking on some smaller feature request, so I don't mind if it gets scratched for now.

Maybe wait for the new settings page? I'm not sure what kind of features you guys are planning for that, but [something like RES](http://i.imgur.com/7mYnlpv.png) seems like the better platform. Then we can hide this somewhere deep in the UI setting hierarchy.

Right now, settings are just a bunch of big buttons, and already as is, I always have a hard time finding the one I'm looking for, adding more small ones like this would just make it worse.

But yeah I'm happy to code whatever.

night:
We are working on a new settings system (@teak is still designing the interface), but there's not really an ETA on when it might be done. If you want to add in the setting now, go for it. Otherwise you might be waiting a while.

night:
@EhsanKia I would just go ahead and just add a setting for this now, and we'll refactor settings down the road.

Issue #471, opened 2015-06-06
Automatically Refresh Views and Followers


evodoc:
It would be nice if the page would automatically refresh those parts just like the viewers the title and played game sections.

night:
This was added by #515

PR #515, opened 2015-06-23
Keep channel viewers and followers up to date

EhsanKia:
This one is for #471

It currently doesn't actually work, as the `/channels/:channel` API request seems to be V2, but the extra information needed (views and followers) are only in V3. Not quite sure where the header for the API version is set (is it in the client-id?), and I don't know how safe it is to switch to V3, but once you do, this code should work as expected, and come at no extra cost since it's done alongside the title/game update request.

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -2968,10 +2968,10 @@ var checkBroadcastInfo = module.exports = function() {
 
     debug.log("Check Channel Title/Game");
 
-    bttv.TwitchAPI.get("channels/"+channel+"?api_version=3").done(function(d) {
+    bttv.TwitchAPI.get("channels/"+channel, {}, {version: 3}).done(function(d) {
         if(d.game) {
             var $channel = $('#broadcast-meta .channel');
-            
+
             if($channel.find('.playing').length) {
                 $channel.find('a:eq(1)').text(d.game).attr("href", Twitch.uri.game(d.game)).removeAttr('data-ember-action');
             }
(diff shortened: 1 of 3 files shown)


night:
Please read the README.md before submitting a pull request.

Additionally, we do not accept non-working code.

EhsanKia:
I went over the README multiple times, it was very useful in setting up the environment, but contains nothing about pull requests. Not quite sure what you're referring to. I've explicitly not included betterttv.js in my commit, is that what you're referring to?

As for the not working, I was hoping you could shed some light on the API version issue, although I guess IRC would be a better place for this. Looking closer at the code, it looks like it's just calling the underlying Twitch method. I managed to find a way to force it to V3, but I'd still like your opinion on if this is a good solution or not. It does work now though.

night:
The method you are using to force v3 is not the right way. You can do the following:

    bttv.TwitchAPI.get('channels/:channel', {}, { version: 3 }).done(function(data) { })

EhsanKia:
Thank you.

Any comment on what I was doing wrong though, which brought up the README comment? I'd like to do things better in the future if possible. I've looked at other pull requests and I can't quite tell what I did wrong.

night:
We do require betterttv.js compiled in pull requests. This may change in the future, since we are restructuring slowly, but for now it is required.

EhsanKia:
Hmm, that method you described works on Twitch.api.get, but bttv.TwitchAPI.get only retains the first parameter and does not pass the rest, so it doesn't work unfortunately. Is that worth fixing?

night:
> Hmm, that method you described works on Twitch.api.get, but bttv.TwitchAPI.get only retains the first parameter and does not pass the rest, so it doesn't work unfortunately. Is that worth fixing?

Feel free to extend our Twitch API wrapper to include support for that.

EhsanKia:
Updated. It now matches almost exactly Twitch's own code, although I'm not entirely sure on the name for that 3rd parameter (since their code is minified).

Issue #257, opened 2015-02-27
Follower Count in Moderator Cards


RuEmileo:
I have a question or I guess a suggestion. You know when you click on someones name in chat and you see the name/follow/message/block etc is there anyway to add in a followers/following count on to that through BTTV? The reason I asked is because sometimes when I mod for streamer I check for bots during giveaways and wanted to know if it was easier to click their name and see how many people they are following to help determine if they were a bot. 
![f292a5eb027ed1ff490e6bfa2eb653f1](https://cloud.githubusercontent.com/assets/11223923/6407850/bc243e26-be16-11e4-84a9-7f3011121cc7.png)

night:
Well, it would probably look a lot better than that screenshot, heh.

The purple "Message" text is a bug, and shouldn't be happening. It's barely readable, and should be a light grey. I'll open a separate issue about that.

RuEmileo:
Lol well thank you for the consideration

night:
Closed by #516

PR #516, opened 2015-06-23
Enhanced mod card

EhsanKia:
This is partly a solution to #257 

It doesn't do following count, but it does add followers and channel views count (which come for free).
Not sure if "following" is worth an extra request, but I can add it if you think it is.
I also considered adding account creation date, but I didn't want to clutter the interface too much.

Here is what it looks like:
 

There's also a background image bug which I fixed in a separate commit:


night:
While the green color you chose for the font looks great for your profile background, it would not for all colors. You should change the color to be more appropriate for all users.

EhsanKia:
I didn't really choose the green hue to match my BG, it just looked bad being the same color as `real-name`. Maybe a yellow-ish hue will look more neutral?

OurFlagIsMined:
The dark yellow text does look pretty nice.

The 'following' count could go either way; would be nice to have there, but for at-a-glance stats, 'followers' and 'views' are much more useful (very useful!) given limited space. You might be able to fit it by right justifying the third stat, but then it may start to be information overkill.

Account creation date, I would omit all together. Many of the newer message boards don't even display that information because it adds very little and doesn't mean all that much.

EhsanKia:
I actually would disagree. As a mod, I use account creation a lot to find fake/troll accounts that were made on the spot. I would even argue that it's more useful than channel views, which generally gives the same amount of information as follower count.

OurFlagIsMined:
In that case, maybe a flag (metaphorical or literal) in the corner, if the account is less than 1 week old?
temp-temp-temp-mod_card_flag-01

night:
I think we need to put the exact date an account was opened, which would look better than a flag. A *lot* of people want account creation date to shown, so that cannot be omitted.

EhsanKia:
Sounds good to me. I think you guys are under a transition to SASS, and since this plays a lot in the CSS, I guess I should wait for that to be over.

DTittel:
You could just write a _modCard.sass if you want :D

night:
@EhsanKia You can again just commit these changes now, and I'll merge them. We're still making plans for style and settings refactoring, and rather than keeping these pull requests open for a long period of time they might as well be merged in.

OurFlagIsMined:
On second thought, following count might be a really useful tool too for weeding out troll accounts. Might be worth that extra API request.

EhsanKia:
Sorry for the delay on these pull requests. My ISP screwed me over and I was left without internet for a week. I'll fix these up real soon.

EhsanKia:
New design, to fit the extra information:

 

 

Currently, hovering on views/follows doesn't show a tooltip though (like under the stream).
Haven't looked into how complicated it'd be, not sure if it's worth it.

Issue #492, opened 2015-06-13
Play Sound on Incoming Whisper


loxplox:
Make it so when someone whispers you, it plays a small sound. I need that! :+1: 
Becuase sometimes people msg me on twitch and i don't see it.

night:
This has been implemented in #517

PR #517, opened 2015-06-25
Whisper Feedback

OurFlagIsMined:
Added an option to play the tink sound on incoming whisper.

#492

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -606,7 +606,11 @@ var onPrivmsg = exports.onPrivmsg = function (channel, data) {
         if (data.style === 'whisper') {
             store.chatters[data.from] = {lastWhisper:Date.now()};
             if (bttv.settings.get('disableWhispers') === true) return;
-            if (data.from !== vars.userData.login) audibleFeedback();
+            if (data.from !== vars.userData.login) {
+                audibleFeedback();
+                if(bttv.settings.get("desktopNotifications") === true && bttv.chat.store.activeView === false)
+                    bttv.notify("You received a whisper from "+((data.tags && data.tags['display-name']) || data.from));
+            }
         }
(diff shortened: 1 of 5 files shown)


night:
Do you think we need an extra option for this? Would grouping it with the other highlight sound option be more appropriate? Thoughts?

cletusc:
IMO highlight sound should cover both.

OurFlagIsMined:
I changed it to one setting for both, and whisper feedback now also displays the visual notice box if desktop notifications are enabled and if the window is inactive.

I originally gave it two settings because of the difference in the way highlight feedback and whisper feedback work. Highlight feedback only plays if desktop notifications are turned on and the window is inactive; whereas the desired behavior for whisper feedback is simply to play a tink for every whisper received, even when the window is active.

If you're watching with chat hidden, or leave the window open while you walk away for a minute, or simply aren't paying attention to chat, you would miss the whisper if audio feedback were tied to desktop notifications and window activity the way highlight feedback is. And whispers are more like PMs than highlighted messages, but if you miss them once they're gone.

Personally I think even the highlight feedback is kind of obscure to use this way, and I don't know of any other chat client that only plays its audio feedback if the window is inactive, but I didn't think it was important to bring up or suggest changing for now as long as whisper audio feedback works the way it should.

temp-temp-temp-git_betterttv_whisper_notice-01

night:
If you change the storage key, you break everyone's settings since the feature is off by default. You should remain with the ```highlightFeedback``` storage key.

OurFlagIsMined:
Fixed.

Tunnel vision; I noticed it getting reset during testing, but forgot it would reset the setting for everyone else too.

night:
Looks good

PR #586, opened 2015-07-22
Added mtgbot back to bot list.

danbopes:
MTGBot was removed a while back during the cleanup, but will be the primary bot in most of the magic channels.


DTittel:
The alignment is wrong, so just going to reject based on that. Still need to get Night's approval to add a new bot to that list.

night:
We're not adding additional bots. Channels can add your bot to their channel bots list at https://manage.betterttv.net

Issue #610, opened 2015-07-30
Pin Highlighted Messages


evodoc:
In popular streams where the chat is very active it's very hard to read the highlighted messages, so how about pinning them? Made a sample image to clarify what I'm talking about.

![pin](https://cloud.githubusercontent.com/assets/11561641/8992192/88e4fcd4-36fd-11e5-94b4-f38c2d2288fb.png)

voloshink:
Added in #705

PR #705, opened 2015-09-12
Pinning the Latest Highlighted Messages above Chat

melalawi:
Saw an open issue for this (#610) and decided to implement it.

Images:

Set to 'disabled' by default. Tried my best to mirror how similar modules were implemented, namely 'embedded-polling'.

Only holds a maximum of 10 pinned messages at a time; newer pins will push out the oldest. Each pin can be closed via the 'X' button.

Gulped and everything. Let me know if there's a problem.


melalawi:
Whoops, forgot to actually commit the gulped fork.

Also I've noticed a strange ESLint rule that had already been set. Apparently `"space-after-keywords"` is set to `2`, which causes the gulp to fail by default since the code is organized in such a way that `"space-after-keywords"` should be set to `0`. What's with that?

src/features/channel-reformat/pin-container.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function() {
+    if (!bttv.settings.get('pinnedHighlights')) {
+        return;
+    }
+
+    // Push pin container to DOM on start if the option is enabled
+    $('.ember-chat .chat-room').append($('<div id="bttv-pin-container">'));
```
vendethiel: I could very well see this function caching its result, and returning it, instead of the code relying on this ID.

src/features/pinned-highlights.js
```diff
@@ -0,0 +1,47 @@
...
+            totalHeight += $(this).outerHeight();
+        });
+
+        // Update the height of the container
+        $highlightContainer.height(totalHeight + 'px');
+    };
+
+    // If the next highlight will bump the container over the limit, remove the oldest highlight
+    if ($highlightContainer.children().length + 1 > maximumPinCount) {
+        $highlightContainer.children().first().remove();
```
vendethiel: shouldn't this call `onPinCountChange` as well? Technically, the pin count *doesn't* change, but its height could, right? (longer message)

melalawi: Thanks for the feedback!

I've gone back and cleaned up the code a bit. I figured that the entirety of the feature should be contained within one js file as opposed to two for the sake of cleanliness and whatnot, so I removed `pin-container.js` and added a simple check within `pinned-highlights.js` that creates and appends the container if it doesn't find it.

I also removed `onPinCountChange` as it is redundant (I never really liked how I implemented it anyways).

melalawi:
Cleaned up the code and changed the contents of the `.close` element so that it actually looks like a close button as opposed to an X.

.eslintrc
```diff
@@ -140,7 +140,7 @@
       "before": false,
       "after": true
     }],
-    "space-after-keywords": 2,       // http://eslint.org/docs/rules/space-after-keywords
```
night: Please do not modify our ESLint settings. They are properly set already.

If it isn't working properly for you, you most likely have some older copy of ESLint locally conflicting.

src/features/pinned-highlights.js
```diff
@@ -0,0 +1,35 @@
...
+
+    // Push pin container to DOM if it doesn't exist
+    if (!$highlightContainer.length) {
+        $highlightContainer = $('<div id="bttv-pin-container">').appendTo($('.ember-chat .chat-room'));
+    }
+
+    // Format the time string
+    var timeSent = message.date.getHours() + ':' + ('0' + message.date.getMinutes()).slice(-2);
+
+    var $nextHighlight = $(highlightTemplate({ time: timeSent, displayName: message.tags['display-name'] + ':', message: message.message }));
```
night: Why are you applying a colon here? It would make more sense if it were just a part of the template itself, no?

test/server.js
```diff
@@ -36,6 +36,6 @@ var server = function(req, res) {
 https.createServer({
   key: fs.readFileSync(path.join(__dirname, 'test-cdn.betterttv.net.key')),
   cert: fs.readFileSync(path.join(__dirname, 'test-cdn.betterttv.net.cert'))
-}, server).listen(443);
+}, server).listen(443, '127.0.0.1');
```
night: Please remove the changes to this file from your PR. If you want to force localhost, please open a separate PR and justify why we should make this change. Some people may want to listen to all interfaces and this change is not related to your PR.

melalawi:
Thank you for responding!

I forgot to revert the ESLint/Server settings before the commit. My bad!

I moved the colon to the template file. Also, to keep consistent with how Twitch displays chat message times, I've set pinned messages to display the message time in the 12 hour format.

night:
Thanks

PR #767, opened 2015-10-14
Fixes Twitch emote menu w/Dark mode + Theater mode

conceptionals:
Originally, when using Dark mode while in Theater mode Twitch's emote selector would still have most of Twitch's dark mode CSS instead of BTTV's (first screenshot), and these CSS changes (second screenshot) should fix that.

githu_pullrequest_1_img1 githu_pullrequest_1_img2

### DIFF
--- a/style/stylesheets/betterttv-dark.css
+++ b/style/stylesheets/betterttv-dark.css
@@ -2032,31 +2032,28 @@ span.subscribe-text {
 
 /* Twitch emote menu */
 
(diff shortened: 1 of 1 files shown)


night:
Anything that requires an "!important" CSS tag is not permitted because it causes other extension developers to hate us. Please revise your pull request to not use "!important" tags.

conceptionals:
There ya go @night, hopefully this is better!

Issue #785, opened 2015-10-24
Slowmode timer is not precise


ozon671games1:
BetterTTV's slowmode timer ticks slower than twitch's slowmode timer
![slowmode](https://cloud.githubusercontent.com/assets/11681316/10711449/394aa000-7a84-11e5-8aa6-aed505418bfb.jpg)

night:
This is unfortunately not something we can fix. We use timeouts in JS, but they are not accurate for time keeping. If your computer is lagging behind or there is a quickly moving chat then the timer may not be accurate. It should still be close enough to give a good judgement of when you can speak again.

PR #786, opened 2015-10-24
Make slowmode timer more precise

ozon671games1:
for issue https://github.com/night/BetterTTV/issues/785

I didn't test this code because i'm not sure how to do it. My version of chrome doesn't support "--ignore-certificate-errors" option


night:
Unfortunately your patch does not function properly. Please follow the testing procedure in README.md

PR #787, opened 2015-10-24
Make slowmode timer more precise [2]

ozon671games1:
for issue #785

### DIFF
--- a/src/features/channel-state.js
+++ b/src/features/channel-state.js
@@ -31,17 +31,14 @@ var resetCountDown = function() {
 var initiateCountDown = function(length) {
     if (bttv.chat.store.chatCountDown) clearInterval(bttv.chat.store.chatCountDown);
 
-    var startTimestamp = Date.now();
-    var endTimestamp = startTimestamp + (length * 1000);
+    var endTimestamp = Date.now() + (length * 1000);
 
     bttv.chat.store.chatCountDown = setInterval(function() {
-        var currentTimestamp = Date.now();
-        var remainingTime = endTimestamp - currentTimestamp;
+        var remainingTime = endTimestamp - Date.now();
         var $chatButton = $(chatButton);
 
         if (remainingTime <= 0) {
             resetCountDown();
-            return;
         } else {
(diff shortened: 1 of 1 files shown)


ozon671games1:
Also, problem with self-signed certificate can be solved by adding ./test/test-cdn.betterttv.net.cert in browser's certificate store.

src/features/channel-state.js
```diff
@@ -31,20 +31,22 @@ var resetCountDown = function() {
...
     bttv.chat.store.chatCountDown = setInterval(function() {
+        var currentTimestamp = Date.now();
+        var remainingTime = endTimestamp - currentTimestamp;
         var $chatButton = $(chatButton);
 
-        if (timer === 0) {
+        if (remainingTime <= 0) {
             resetCountDown();
             return;
+        } else {
```
night: If you're returning in the above `if`, then the `else` block is superfluous. That's why it doesn't exist in the code you're editing.

src/features/channel-state.js
```diff
@@ -31,20 +31,22 @@ var resetCountDown = function() {
 var initiateCountDown = function(length) {
     if (bttv.chat.store.chatCountDown) clearInterval(bttv.chat.store.chatCountDown);
 
-    var timer = length;
+    var startTimestamp = Date.now();
+    var endTimestamp = startTimestamp + (length * 1000);
```
night: You're setting a variable and then using it one time (and never again) in the next line. Just use `Date.now()` in the next line.

src/features/channel-state.js
```diff
@@ -31,20 +31,22 @@ var resetCountDown = function() {
...
 
     bttv.chat.store.chatCountDown = setInterval(function() {
+        var currentTimestamp = Date.now();
+        var remainingTime = endTimestamp - currentTimestamp;
```
night: You're setting a variable and then using it one time (and never again) in the next line. Just use `Date.now()` in the next line.

night:
Please address the commented lines by updating your pull request. Once updated we can test and merge it in.

night:
Thanks

Issue #794, opened 2015-10-28
CSS support for tag filter arrows


AlcaDesign:
![Example](https://i.gyazo.com/45e7aa17b614f3b1c0a4733ee767ce94.png)

On the [creative game directory](http://www.twitch.tv/directory/game/Creative) and probably others, there is a tag filter.

Example from creative: (at second "page")
```html
<div class="creativetag-list-contain filter-list-contain js-filter-list-contain clearfix">
  <a class=" filter-nav goleft js-goleft"></a>
  <a class=" filter-nav goright js-goright"></a>
  <div class="creativetag-list filter-list js-filter-list clearfix">
    <ul style="left: -1284px;">
      <li><a id="ember3121" class="ember-view active" href="/directory/game/Creative" data-ytta-id="-"><span class="creativetag">All Channels</span></a></li>
    </ul>
  </div>
</div>
```

night:
It's [Hacktoberfest](https://hacktoberfest.digitalocean.com)! Feel free to open a pull request to fix this, since you seem capable to do so. Otherwise we'll get it fixed eventually.

Issue #788, opened 2015-10-25
Notify when a message's length is too long to send


AlcaDesign:
When sending some messages during a subscriber spam, for example, I want to make sure that my message goes through with maximum ![HY](https://static-cdn.jtvnw.net/emoticons/v1/11301/1.0)![PE](https://static-cdn.jtvnw.net/emoticons/v1/11302/1.0).

Messages over [~500 characters](https://tools.ietf.org/html/rfc2812#section-2.3) should notify that it probably won't be sent through.

voloshink:
Sounds like it could be a good adition, what would be the best way to let the user know though?

AlcaDesign:
Either just set the max length on the chat box or a message just like when you host someone, time-out, etc.

EhsanKia:
@voloshink Just a notice probably, aka `serverMessage` in helpers.

voloshink:
Might just end up limiting the textarea to 500 characters, maybe a notice in chat too

AlcaDesign:
A whisper's setup (/w name) should be taken into account, though

voloshink:
is the character limit different for whispers?

AlcaDesign:
No, but typing "/w <name>" would be changed into a whisper event rather than sending "/w <name> <message>" to the server.

voloshink:
I see

voloshink:
I tested it out, for both /me and /w the limit is still 500 total, otherwise the message doesn't go through

EhsanKia:
Is that 500 including the "/w username" and "/me" part or only the messages after it?

voloshink:
500 with the command itself

night:
Twitch limits user input, regardless of command, to 500 characters. It's still [Hacktoberfest](https://hacktoberfest.digitalocean.com)! Feel free to open a pull request that adds a `maxlimit="50"` onto the textarea for chat input.

PR #796, opened 2015-10-29
Add creative directory tag filter styles and limit chat to 500 chars

AlcaDesign:
Fixes night/BetterTTV#794
Fixes night/BetterTTV#788

### DIFF
--- a/style/stylesheets/betterttv-dark.css
+++ b/style/stylesheets/betterttv-dark.css
@@ -2170,18 +2170,18 @@ span.subscribe-text {
     opacity: .5;
 }
 
-.filter-nav.goleft {
+.creativetag-list-contain .filter-nav.goleft {
     border-right: 1px solid #666;
     border-left-color: transparent;
 }
-.filter-nav.goright {
+.creativetag-list-contain .filter-nav.goright {
     border-left: 1px solid #666;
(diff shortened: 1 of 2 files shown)


style/stylesheets/betterttv-dark.css
```diff
@@ -2156,3 +2156,32 @@ span.subscribe-text {
...
+
+.creativetag-list-contain .filter-nav:before {
+    margin-top: -9px;
+    border: 5px solid #fff;
+    border-top-color: transparent;
+    border-bottom-color: transparent;
+    opacity: .5;
+}
+
+.filter-nav.goleft {
```
night: `.filter-nav` alone is not specific enough, since this css file is loaded everywhere. This may cause theme conflicts elsewhere.

AlcaDesign: This is what Twitch is using their stylesheet.
Twitch's stylesheet

night: That means nothing to me, since Twitch re-uses classes in multiple different CSS files and we have no way to verify it won't break now or in the future (hence the "may cause theme conflicts elsewhere").

night:
Thanks!

PR #866, opened 2015-12-19
Emote only auto completion

ghost:
I use @ style auto completion for users so when trying to auto complete emotes without using @ I would have to tab many times to get past users with similar names to emotes.

A couple examples:

In TriHex's stream. If I type ```tri``` the first completion would be ```Trihex, ```. With this pull request ```tri``` will complete to ```TriHard```. If I type ```kap``` at this moment in chat it brought up ```KappaMasterSwag3, ``` with the feature on it now brings up ```KappaCool``` with the next tab being ```Kappa```. Finally, ```dan``` without became ```dank__doge, ``` while with it becomes ```DansGame```.

@ style auto completion still works so I can now have the best of both worlds.

The setting headline is **Emote Only Auto-Complete** with the description **Tab auto-completion only uses emotes**. These could be changed, I'm not sure what the best would be.

Thanks for the great addon!

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -1075,7 +1075,7 @@ exports.tabCompletion = function(e) {
 
             // Mix in emotes if not directly asking for a user
             if (lastWord.charAt(0) !== '@' && !detectServerCommand(input)) {
-                users = users.concat(emotes);
+                users = bttv.setting.get('emoteOnlyAutoComplete') ? emotes : users.concat(emotes);
             }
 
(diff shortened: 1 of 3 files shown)


jameskopacz:
Thanks!

jordanbtucker:
I would love to see this feature implemented! Thanks.

voloshink:
Once we implement our new settings system, i'd like to add an option where you choose which list takes priority: emotes or user names.

night:
closing in favor of #896

Issue #267, opened 2015-03-01
Add Regular Expressions to Highlight/Blacklist


rue92:
I'd like to request adding regular expression support for highlighting text/users in chat. For some this allows highlighting username variants without typing out all permutations. Another potential (though unlikely) use would be for highlighting game series, e.g. `[Ff]{2}(\d*|[IVX]*)`. I can see this requiring a relatively large rewrite though due to use of () and {} in the configuration already. Just a thought :D. 

P.S. I didn't actually test that regex, so don't blame me if it's wrong! :P

vendethiel:
Yeah, considering the current state of the highlighting, this isn't possible. A rewrite may allow that, but we need to warn users properly not to use if they're not really tech-savvy.

skerg:
I would like to second this request. It would be especially useful to have when blacklisting messages.

metarmask:
:+1: This would be really useful.

You could add it to the settings with an explaination: "Regex Keywords -- Use regular expressions in keywords. Only enable if you know what you are doing."

Codonbyte:
You could do what Chatty does and precede regular expressions with reg::{regex goes here}.

PR #891, opened 2016-01-06
Regex Highlighting

rue92:
This includes my modifications in order to support regular expression highlighting for messages as per Issue #267 which I created in March. It's nothing fancy, but I figure most people have no desire to use JS regular expressions so I included the option to turn it on/off. Otherwise, I made it so the user has to delimit their regular expression with semicolons, but can otherwise treat it like a regular highlight keyword -- it's not the greatest, but it worked. 

Comments, criticisms, etc. welcome. Thanks.

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -4760,8 +4760,31 @@ exports.highlighting = function(data) {
     var highlightUsers = [];
 
     var extraKeywords = bttv.settings.get('highlightKeywords');
-    var phraseRegex = /\{.+?\}/g;
 
+    // Pull the regular expressions out first so curly braces
+    // in the expression won't double count as phrases
+    var highlightRegex = [];
+    var regexPhrase = /;.+?;/g;
+    var regexStrings;
+    var i;
+    try {
+        regexStrings = extraKeywords.match(regexPhrase);
+    } catch (e) {
+        debug.log(e);
+        return false;
+    }
+    if (regexStrings) {
+        for (i = 0; i < regexStrings.length; i++) {
+            var regexString = regexStrings[i];
+            debug.log(regexString);
+            extraKeywords = extraKeywords.replace(regexString, '')
+                .replace(/s\s\s+/g, ' ').trim();
+            highlightRegex.push(regexString.replace(/(^;|;$)/g, '')
+                                .trim());
+        }
+    }
+
+    var phraseRegex = /\{.+?\}/g;
     var testCases;
     try {
         testCases = extraKeywords.match(phraseRegex);
@@ -4770,7 +4793,6 @@ exports.highlighting = function(data) {
         return false;
     }
 
-    var i;
     if (testCases) {
         for (i = 0; i < testCases.length; i++) {
             var testCase = testCases[i];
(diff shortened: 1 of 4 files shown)


night:
Hi, this is an interesting pull request. Is there a reason you're using semicolons and not forward slashes? Forward slashes are typically used to denote regex and may better fit for this application.

Additionally, there's a large worry users will lock up their page with a bad regex. You may want to wrap the user input in an atomic group so the regex will return quickly: http://www.regular-expressions.info/atomic.html

rue92:
Semicolons as opposed to forward slashes was really a choice I made before I put the settings in to enable/disable regular expressions. Otherwise, I agree forward slashes are the well established convention, and I'll make the change appropriately.

As for atomic groups, that's a new concept to me (the struggles of a novice programmer :P), but I can see the need so I'll take a look and figure something out. Thanks for the help!

rue92:
I've made the delimiter change, however, according to http://www.regular-expressions.info/javascript.html and my own verification, JS doesn't support atomic groups. I'll leave it up to your discretion whether you take the request with that information.

ar664:
Would it be possible to use this for Blacklisting as well?

rue92:
Sorry for very late response, @ar664. Definitely can be done, and it would make for a great excuse to refactor the code to make it cleaner in both hilighting and blacklisting. 

Only thing is, and I understand the concern by the maintainer, that there isn't an obvious way to keep a user from defining a regex that locks up their browser. On that note, I wonder if there's a way to wrap the regex match with a timeout of some sort? I'll have to investigate.

night:
Found this lib on npm that might allow regexes to be used: https://www.npmjs.com/package/safe-regex

If we can check that all (or at least most) regexes are safe before saving them (or I guess executing them) it might be safe enough.

ar664:
I looked around and I found this site to possibly generate new regexp to test. 
http://txt2re.com/index-javascript.php3?s=11:May:2016%20%22This%20is%20an%20Example!%22&7

Maybe you can do some sort of brute force method to see the avg run-time / exp-failures of the given expressions.

night:
The concern is not a predefined set of expressions, but rather any expression a user enters. Users who don't understand need to be safeguarded, which is why the package I linked may work for this use-case. Piping regexes into a web service to get a run time is not a good solution here, nor is it something we could accomplish locally in JavaScript due to how JavaScript executes regexes (they cannot be stopped early).

ar664:
Are you running linux? It is possible to kill a process beyond a threshold time using scripts, but I understand the reasoning behind why you wouldn't want to get pre-generated regex's. The only question I have is how you would got about testing the safe-regex for viability? 
My method is crude, but its a little better than randomly generating strings because I believe safe-regex tests for whether the expression is one of regex.

rue92:
I think there's a little confusion here.  It sounds like ar664 is suggesting we test safe-regex (which is a great find by the way) ourselves with a few crafted degenerate cases to verify it. I'm alright with doing this and I can try to tackle it this week/weekend as well as make the changes I promised way back when. How does that sound?

ar664:
@rue92 Yes, That is exactly what I'm trying to say thank you. @night Also, I'm sorry if I sounded ignorant or rude before English is my second language, so I tend to not give tone appropriately in writing.

night:
> I think there's a little confusion here. It sounds like ar664 is suggesting we test safe-regex (which is a great find by the way) ourselves with a few crafted degenerate cases to verify it. I'm alright with doing this and I can try to tackle it this week/weekend as well as make the changes I promised way back when. How does that sound?

That works fine. Just do note you need to also rebase your fork from master because it's severely out of date.

night:
closing this pr due to inactivity. open a new pr if you address the issues.

PR #896, opened 2016-01-09
Added option to tab complete emotes before usernames

OurFlagIsMined:
Highly requested feature. For those of us that use tab completion for emotes, we'd much rather have it jump to an emote first than a username (eg by default, 'pogc' will complete to the username PogCaamp before PogChamp if that user is in the chat).


OurFlagIsMined:
I didn't see the other open PR #866 when I made this, interesting how we both came to style it almost identically. That being said, I'd prefer emote priority matching over disabling name matching completely, as many of use use the bare syntax (no  "@") for both.

Issue #905, opened 2016-01-15
Auto-theatre mode


fgsfdss:
Edit: Auto-theatre mode is still there, it just doesn't work for me. Any help?

voloshink:
works for me, can't reproduce

fgsfdss:
Probably should have specified. I'm using Firefox 43.0.4, and I did try disabling other potentially conflicting addons.

x003:
Can confirm, it's not working for me either (I'm on Chrome)

EsDeKa:
I found a way to reproduce it, if go to a channel through a game directory (ex http://www.twitch.tv/directory/game/Dota%202 ) it doesn't activate. When I refresh the stream afterwards, theatre mode gets activated.

EhsanKia:
I'm not entirely sure it's the same bug, I could be wrong.

My understanding is that when you first load a page, `initialFuncs()` in `main.js` gets called, which calls `autoTheatreMode`.

Whereas when you switch pages, the Ember render callback is called, and if the template is `channel/index'`, we again call `AutoThreateMode`.

The latter is definitely broken in Chrome, and after some debugging, it's called too soon (even if it's wrapped in a ready callback).

`App.__container__.lookup('controller:channel').get('theatreMode')` is undefined, and the if statement fails.

The former is probably a similar issue where the controller is not fully loaded, but I can't debug because I can't get certificate exception working in FF.

EhsanKia:
Found the issue, `.get('theatreMode')` needs to be changed to `.get('isTheatreMode')`

This fixes the directory -> channel but, but I'm not sure if it fixes the Firefox issue.

night:
Fixed in e82fb21c02fb5d942573c258f35d852a7c5c0a44

PR #907, opened 2016-01-15
Change of variable name to check if theatre mode is enabled

EhsanKia:
This should fix, at least partly, the issue in #905 
This fixes theatre mode detection when switching from a non-channel Ember route into a channel route.


night:
Fixed already.

PR #913, opened 2016-01-20
Add hnlbot to bots list

GavoTrav:
hnlbot is a pretty big bot these days in channels such as sodapoppin, summit1g, timthetatman and lots more.

should have a global bot tag


night:
We are not maintaining a global bot list, sorry. People can add their channel bots to their channel bot list at https://manage.betterttv.net

PR #976, opened 2016-02-23
Fixed linkify regex

OurFlagIsMined:
Minor change.

Wasn't capturing some links properly; i.e. "http://www.google.com/" only captured "http://www.google.com".

I saw what changed the last time (ensuring a link could contain "!" or "," but not end in it) and left that part the same. Just fixed the the capturing and removed the 2 non-capturing groups that had no function in that context.

### DIFF
--- a/src/chat/templates.js
+++ b/src/chat/templates.js
@@ -32,7 +32,7 @@ var modicons = exports.modicons = function() {
 };
 
 var linkify = exports.linkify = function(message) {
-    var regex = /(?:https?:\/\/)?(?:[-a-zA-Z0-9@:%_\+~#=]+\.)+[a-z]{2,6}\b(?:(?:[-a-zA-Z0-9@:%_\+.~#?&\/=!,]+)(?:[-a-zA-Z0-9@:%_\+.~#?&\/=]))?/gi;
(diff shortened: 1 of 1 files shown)


OurFlagIsMined:
The solution to the spurious semicolon bug was also in the linkify regex, and includes allowing semicolons in links.

The bug occurs when a "<" or ">" gets escaped, then linkify cuts the "&lt;" or "&gt;" into two pieces, causing spurious semicolons to sometimes appear. It's especially noticeable when pasting code or other things into chat that other people are meant to copy, because what comes out in the message and on their clipboard is different from what you originally pasted (it should always come out the same as you entered it, and this is the only instance I've come across where it doesn't).

night:
I'm not sure why you're using a wildcard in place of our capture group. Allowing all characters is *not* ok.

OurFlagIsMined:
Haven't checked my notifications in a while.

There weren't any wildcards in it, the only functional change made was swapping out "+" for "*", allowing it to contain 0-or-more characters from that character set instead of 1-or-more; which was causing it to capture if there were 0, or 2+ valid characters, but not if there was exactly 1 (such as when "/" was the only character after the domain name).

It's irrelevant to the linkify RegEx as it currently stands (https://github.com/night/BetterTTV/blob/master/src/js/chat/templates.js#L39), but still a valid model if there's ever again the need to allow certain characters in the URL but not allow them as the last character.

Issue #1015, opened 2016-03-14
Emote Identification


ToxiClay:
BTTV does not properly identify whether a hovered-over emote is a BTTV emote or a partner emote.

PR #1031, opened 2016-03-22
differentiate BTTV channel emotes

EhsanKia:
Solution for #1015 
BTTV emotes have (BTTV) after the channel name

image

### DIFF
--- a/betterttv.js
+++ b/betterttv.js
@@ -344,7 +344,9 @@ module.exports = function() {
 
         emote.text = emote.code;
 
-        if (!emote.channel) {
+        if (emote.personal) {
+            emote.channel = 'Personal Emotes';
+        } else if (!emote.channel) {
             emote.channel = 'BetterTTV Emotes';
             emote.badge = 'https://cdn.betterttv.net/tags/developer.png';
         }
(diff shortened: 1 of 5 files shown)


src/features/override-emotes.js
```diff
@@ -32,7 +32,7 @@ module.exports = function() {
                         } else if ($emote.data('channel') && $emote.data('channel') === 'BetterTTV Emotes') {
                             return 'Emote: ' + raw + '<br />BetterTTV Emoticon';
                         } else if ($emote.data('channel')) {
-                            return 'Emote: ' + raw + '<br />Channel: ' + $emote.data('channel');
+                            return 'Emote: ' + raw + '<br />Channel: ' + $emote.data('channel') + ' (BTTV)';
```
vendethiel: every single channel-specific twitch emotes will have their `data-id` set to some emote in `TwitchEmoteIDToChannel`?

EhsanKia: Yeah, I'm wasn't 100% sure about this, I'm not 100% sure how up to date and accurate that list is kept.

The other two options I can think of are:
1. Set a data attribute on BTTV emotes
2. look at the href and check for BTTV cdn

EhsanKia:
So we don't have access to the emotemenu code from here.
Unless we get them to make specific changes, the only way I can see to get around it is by setting the channel name on pro emotes to "Personal Emote". It works, the only downside is, on your end, if you hover on your own emote, you won't see the actual channel.

image

image

betterttv.js
```diff
@@ -2769,6 +2769,7 @@ var main = function() {
                         waitForLoad(function(ready) {
                             if (ready) {
                                 handleBackground();
+                                debug.log('test 1');
```
vendethiel: oops!

betterttv.js
```diff
@@ -2851,6 +2852,7 @@ var main = function() {
         chatReplay = new ChatReplay();
         conversations();
         clearClutter();
+        debug.log('test 2');
```
vendethiel: oops!2

src/features/override-emotes.js
night: You should not worry about about how this data is managed. Every emote that we're aware of is mapped to a channel. If not, it just won't return a channel at all (which is fine).

src/ws.js
```diff
@@ -48,6 +48,9 @@ events.lookup_user = function(subscription) {
         bttv.chat.store.proEmotes[subscription.name] = {};
 
         subscription.emotes.forEach(function(emote) {
+            if (subscription.name === vars.userData.name) {
+                emote.channel = 'Personal Emote';
```
night: We still want personal emotes to still show the channel they originate from.

EhsanKia:
In this new version adds a property to emotes coming through the websocket.
It also sets an emote-type data attribute on the image, and uses that on the tooltip.
Lastly, the emotemenu getter uses the property set above to modify the channel right before passing it.

PR #1067, opened 2016-04-07
Version update (6.8)

0-Eclipse-0:


night:
We don't really follow any versioning guidelines at this time, but we may swap to semver at v7 when it eventually drops.

0-Eclipse-0:
Ah

PR #1068, opened 2016-04-08
Dark mode banner color fix [squashed]

samfun123:
On the side of the banner their is a small purple bar but it doesn't have a consistent width. The explanation button on the banner is also not colored correctly for dark mode.

Without fix:
image

With fix:
image

### DIFF
--- a/style/stylesheets/betterttv-dark.css
+++ b/style/stylesheets/betterttv-dark.css
@@ -1970,7 +1970,7 @@ color: white;
 .player-column .brick {
     background-color: #222;
     border-left: 5px solid #6441a5 !important;
-	box-shadow: 0px 0 0 #6441a4 inset !important;
+    box-shadow: 0px 0 0 #6441a4 inset !important;
 }
 
 .what {
--- a/style/stylesheets/betterttv-dark.css
+++ b/style/stylesheets/betterttv-dark.css
@@ -1969,6 +1969,12 @@ color: white;
 
 .player-column .brick {
     background-color: #222;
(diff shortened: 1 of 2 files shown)


style/stylesheets/betterttv-dark.css
```diff
@@ -1969,6 +1969,12 @@ color: white;
 
 .player-column .brick {
     background-color: #222;
+    border-left: 5px solid #6441a5 !important;
+	box-shadow: 0px 0 0 #6441a4 inset !important;
```
night: formatting

style/stylesheets/betterttv-dark.css
```diff
@@ -1969,6 +1969,12 @@ color: white;
...
+}
+
+.what {
```
night: `.what` is very nonspecific. What if that matches something else on the site and messes it up?

style/stylesheets/betterttv-dark.css
```diff
@@ -1969,6 +1969,12 @@ color: white;
 
 .player-column .brick {
     background-color: #222;
+    border-left: 5px solid #6441a5 !important;
+    box-shadow: 0px 0 0 #6441a4 inset !important;
+}
+
+.what {
```
samfun123: .what is used in the default twitch CSS so I followed that.

style/stylesheets/betterttv-dark.css
samfun123: ``.what`` is what twitch uses in their CSS, so I just followed that.

PR #1073, opened 2016-04-10
Translation fix + tooltip for original message

EhsanKia:
Tested the translation and it works great.
No more uri encoded spaces or undefined bugs.

Also added the original non-translated message as a tool tip upon successful translation. Sometimes you accidentally double click, or just wanna compare the two messages, and there's no way for you to get the original back otherwise.

### DIFF
--- a/src/settings-list.js
+++ b/src/settings-list.js
@@ -171,8 +171,8 @@ module.exports = [
         toggle: function(value) {
             if (value === true) {
                 $('body').on('dblclick', '.chat-line', function() {
-                    chat.helpers.translate($(this).find('.message'), $(this).data('sender'), $(this).find('.message').data('raw'));
-                    $(this).find('.message').text('Translating..');
+                    chat.helpers.translate($(this).find('.message'), $(this).data('sender'), $(this).find('.message').text());
+                    $(this).find('.message').text('Translating...');
                     $('div.tipsy').remove();
                 });
             } else {
--- a/betterttv.js
+++ b/betterttv.js
@@ -1717,7 +1717,19 @@ exports.translate = function($element, sender, text) {
     });
 
     $.getJSON('https://api.betterttv.net/2/translate?' + qs).success(function(data) {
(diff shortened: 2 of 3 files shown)


src/settings-list.js
```diff
@@ -171,8 +171,8 @@ module.exports = [
         toggle: function(value) {
             if (value === true) {
                 $('body').on('dblclick', '.chat-line', function() {
-                    chat.helpers.translate($(this).find('.message'), $(this).data('sender'), $(this).find('.message').data('raw'));
-                    $(this).find('.message').text('Translating..');
+                    chat.helpers.translate($(this).find('.message'), $(this).data('sender'), $(this).find('.message').text());
```
night: I'd rather no longer have `data-raw` be rendered.. afaik this is all it was used for, but if not you need to go back and remove it and update all its dependencies.. The only reason we used .raw was for emote support, but with alt tags it's no longer necessary since .text() can fill those in.

EhsanKia: It's used to recover deleted messages. I don't think there's any other way to get the message back in that case unless we keep it inside the closure of the "click to restore" event handler.

EhsanKia: Not sure if I'm being stupid, but I can't find anything about jQuery .text() picking up image alt text.

night: Since we do have a click to restore feature, I'm not sure how we'd handle this then. It might be best to just keep the raw for now. Know of a better way?

EhsanKia: For translation specifically, honestly I think `.text()` which filters out emotes is better, as you don't really want to give emotes to the translation. So it's more of a feature that emotes are being filtered, as they could throw off google translate.

So I think this PR is good as it is. If not, to fix the bug, you'd need to add `decodeURIComponent()` around the message as right now I'm pretty sure the data is being encoded twice (resulting in the bug).

night:
Fixed by #1095

Issue #996, opened 2016-03-03
AudioTwitch integration


Glaived:
An option to add a redirecting button to the service @AudioTwitch ?
e.g. : http://audiotwitch.tv/YOUR_CHANNEL_NAME_HERE

![ccpj-plxiaad3lp jpg large](https://cloud.githubusercontent.com/assets/1835170/13509856/06621120-e18e-11e5-8004-4a40df63d47c.jpg)

night:
this has been broken for a while

PR #1091, opened 2016-04-27
Add new feature #996

MergHQ:
Direct link to audiotwitch.tv under the video player.

### DIFF
--- a/src/features/audiotwitch-btn-below-video.js
+++ b/src/features/audiotwitch-btn-below-video.js
@@ -1,0 +1,15 @@
+module.exports = function() {
+    if (bttv.settings.get('audiotwitchButton') !== true) return;
+
+    var $btn = $('#bttv-audiotwitch-button');
+    if (!$btn.length) {
+        $btn = $('<span><span></span></span>');
+        $btn.addClass('button').addClass('action');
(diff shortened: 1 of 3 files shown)


night:
AudioTwitch doesn't seem to work..

MergHQ:
That's weird. I'll tweet them and ask what's up.

samfun123:
A optional replacement would be [radiotwitch.in](http://radiotwitch.in/) which seems to be working.

MergHQ:
Okay. It's now using radiotwitch.in.

night:
Meh. I definitely am not a fan of this integration. I'd rather have something that embeds an iframe over Twitch's player and looks a lot nicer/modern. A popup window isn't really a native implementation that makes sense.

MergHQ:
I can't use iframes because radiotwitch uses http and not https. Therefore this feature cannot be implemented.

night:
Sorry for the late response. You might try reaching out to either of these sites to ask if they would add ssl support. Many might be willing. ¯\\_(ツ)_/¯

Zalatis:
Atm radiotwitch and audiotwitch doesn't work :'( any others website who do the same job ?

PR #1097, opened 2016-04-30
Fixed Dark Mode channel title hover

voltaek:
The `div.title` channel title element has styling applied on its `:hover` state now rather than a child element (which isn't there anymore) getting `.over` applied to it (presumably was done by JS).

Text is still styled properly with some more general Dark Mode styling, so this fixes Twitch's new styling by that puts a `background-color: white;` on `.title:hover`.

### DIFF
--- a/style/stylesheets/betterttv-dark.css
+++ b/style/stylesheets/betterttv-dark.css
@@ -300,12 +300,8 @@ body {
     border-top: 1px solid rgba(0, 0, 0, 0.65);
 }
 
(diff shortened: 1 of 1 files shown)


night:
Fixed already by 38a721b33642ec131b98d0dbc824ba4dba4018ce

voltaek:
@night That commit didn't clean up the no longer used text color styling for `.over` and also uses black instead of the almost-black that the background actually is.

night:
cc @teak

PR #1098, opened 2016-04-30
Vertically center align chat header items

voltaek:
I vertically center aligned the text and icons in the header of the chat sidebar. I believe this is how it used to be before Twitch made some styling changes. Either way, it looks much better now.


night:
👍

PR #1099, opened 2016-04-30
Twitch Theater Mode overrides chat header

voltaek:
Fix for Twitch changing color of chat header text in theater mode.


night:
👍

Issue #1057, opened 2016-04-02
hide whisper icon in theater mode


evodoc:
It can be annoying, especially if you minimize the chat (at least hide it if there are no new messages and/or show it only when the video is hovered):
![untitled](https://cloud.githubusercontent.com/assets/11561641/14228214/3cf55d26-f90f-11e5-907d-e125984d3b54.jpg)

night:
You can now enable "Hide Conversations When Inactive" in BetterTTV settings.

PR #1110, opened 2016-05-04
Add #1057

MergHQ:
Setting to hide whisper conversation bar and icon in theatre mode.


night:
honestly if someone wants to disable whispers they should just use that existing option instead.

PR #1119, opened 2016-05-06
Option to disable front page video

MergHQ:
Someone requested this.

The video player is simply replaced with this:
80ee98327932f69a35c25d209292ef74

### DIFF
--- a/src/settings-list.js
+++ b/src/settings-list.js
@@ -272,6 +272,20 @@ module.exports = [
         storageKey: 'disableWhispers'
     },
     {
+        name: 'Disable Frontpage Video',
+        description: 'Disable video on the frontpage',
+        default: false,
+        storageKey: 'disableFPVideo',
+        load: function() {
+            if (window.location.href === 'https://www.twitch.tv/' && bttv.settings.get('disableFPVideo') === true) {
+                $(window).load(function() {
+                    $('#video-1').hide();
+                    $('.items').css('width', '940px');
+                });
+            }
+        }
+    },
+    {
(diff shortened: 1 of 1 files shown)


src/settings-list.js
```diff
@@ -272,6 +272,25 @@ module.exports = [
         storageKey: 'disableWhispers'
     },
     {
+        name: 'Disable Frontpage Video',
+        description: 'Disable video on the frontpage',
+        default: false,
+        storageKey: 'disableFPVideo',
+        load: function() {
+            if (window.location.href === 'https://www.twitch.tv/' && bttv.settings.get('disableFPVideo') === true) {
+                $(window).load(function() {
```
night: This is ran after the page loads: https://github.com/night/BetterTTV/blob/master/src/main.js#L311

night:
I would much prefer you auto-pause the existing player rather than remove it.

MergHQ:
So the video player would not be removed at all, just auto paused?

night:
That's the idea. It's a really bad idea to remove it entirely, as some users might want to still preview the featured stream.

MergHQ:
Okay, I made the changes. It's now paused instead of removed completely.

night:
You did not address my code comment from earlier: https://github.com/night/BetterTTV/pull/1119#discussion_r62616373

MergHQ:
The problem is that the DOM is loaded before the video iframe. Look like something the site dynamically creates the iframe when the window is ready.

At least i can't get it to work if i remove $(window).load

PR #1152, opened 2016-05-13
Hide only the recommended friends

### LINKED ISSUE #1147: Hiding recommended friends
I saw there is an option to hide the friends sidebar. I do like the friends feature.
The only thing that is really annoying in my opinion is the "recommended friends" sidebar.

Would be a nice feature to hide recommended friends but still see your friends list.

WobblyToddler:
Closes #1147


night:
I think a separate setting for the friends list feature for another one line css file is misguided. This will have to wait until we add our new settings menu and we can offer people more flexibility, sorry.

PR #1161, opened 2016-05-14
Bug fixes

voloshink:
will update this once the other pull request are merged


EhsanKia:
Hmm, wouldn't just removing the < > replace allow XSS attack?

PR #1168, opened 2016-05-16
Chat replay and emote tooltip fixes

voloshink:


src/js/chat/templates.js
```diff
@@ -42,7 +42,7 @@ var linkify = exports.linkify = function(message) {
 };
 
 var escapeEmoteCode = function(code) {
-    return code.replace(/('|"|>|<|&)/g, '');
```
night: This opens us up to potential xss in the event someone doesn't handle the data properly.. If you want to save the character it has to be encoded.

EhsanKia: From what I understand, a simple encode doesn't work either, because the plugin sets this as alt-text, and reads from it. You'd need to escape everything TWICE, but the escape function doesn't handle `&` so it breaks.

Test it out by force injecting an emote with code `<script>`

voloshink:
Damn it

EhsanKia:
Try adding `.replace(/&/g, '&amp;')` to escape and call escape twice, I'm curious if it'll work.

src/js/chat/templates.js
night: By encode I mean html entity encoding.. `<` becomes `&lt;`, `>` becomes `&gt;` and `&` becomes `&amp;`

Issue #1157, opened 2016-05-14
/t and /b does not work with the new Timeout reasons & durations


MattiSony:
Title :P If you add a reason after the command it's not included.

night:
hey, that's pretty fixed

PR #1170, opened 2016-05-16
Add ban reason to /t and /b commands

EhsanKia:
Fix for #1157

Also fixes all other issues with timeout reasons.

### DIFF
--- a/src/js/chat/handlers.js
+++ b/src/js/chat/handlers.js
@@ -31,7 +31,7 @@ exports.commands = function(input) {
     var oldSetting;
 
     if (command === '/b') {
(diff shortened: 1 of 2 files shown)


src/js/chat/handlers.js
```diff
@@ -341,6 +345,8 @@ exports.notice = function(data) {
         message: message
     });
 
+    if (messageId === 'timeout_success' || messageId === 'ban_success') return;
```
night: unnecessary now yeah?

EhsanKia: Is it 100% gone? Then yeah I'll remove it.

Issue #1062, opened 2016-04-05
add timestamps for chat replay


evodoc:
Timestamps as in playtime times.

night:
implemented by bb0438cb69f69d538bb0966b2745f10ac0bd6f4b

PR #1194, opened 2016-05-20
Add chat replay timestamp option

MergHQ:
Implements issue #1062

So I'm not really sure if it is better to create a new CSS file and load it to the DOM, or use a event listener and change the display style property for each chat line.

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -204,9 +204,9 @@ var main = function() {
                     chatReplay = new ChatReplay();
                     window.dispatchEvent(new Event('resize'));
                     break;
-                case 'following.index':
+                case 'directory.following.index':
                     // Switching between tabs in following page
-                    if (lastRoute.substr(0, 9) === 'following') break;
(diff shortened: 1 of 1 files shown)


EhsanKia:
Chat already has a built-in "show time-stamp" option. Can we just re-use that instead of making YET another option in the settings menu? I realize you don't get the cogwheel in the chat replay so you can't toggle it while watching a VOD, but anyone who cares about timestamps will have it enabled for chat, and will therefore get it also in replays. I don't see why'd you would want it in one and not the other.

Here's how to access the setting:
`Twitch.storage.getObject('chatSettings').showTimestamps`

MergHQ:
I didn't know you could get twitch user settings, thanks for showing me. I'll make the changes.

night:
I don't like loading one line CSS files. How about adding a show-timestamp class to chat lines as they render with our pre-existing chat replay logic?

MergHQ:
Sounds good. I'll do that.

PR #1197, opened 2016-05-22
Add whisper button to the BTTV moderation card

EhsanKia:
A bit of reverse engineering later, here's a clean 2 line solution for it.

I just reused the same icon as the whisper menu for the button.


night:
Implementation LGTM, but I'd rather you get teak to handle a proper svg since the current icon looks out of place..

cc @teak

PR #1198, opened 2016-05-22
Twitch emote fix and hide conversations bugfix

voloshink:


src/js/chat/templates.js
```diff
@@ -73,6 +73,11 @@ var jtvEmoticonize = exports.jtvEmoticonize = function(id) {
 };
 
 var emoticon = exports.emoticon = function(id, name) {
+    var fixedTwitchEmotes = ['15', '16', '17', '18', '19', '20', '21', '22', '26', '27', '33'];
```
night: I'd say store as `Number`, type convert `id` below to a `Number`

src/js/features/conversations.js
```diff
@@ -127,7 +127,9 @@ Conversations.prototype.newConversation = function(element) {
 
     if (bttv.settings.get('hideConversations')) this.slideUp();
     $(element).find('.header-button-container').children().last().on('click', function() {
-        if (bttv.settings.get('hideConversations') && $('.conversation-window').length === 1) _self.slideDown();
+        setTimeout(function() {
+            if (bttv.settings.get('hideConversations')) _self.slideDown();
```
night: What happened to `$('.conversation-window').length === 1` here? Would removing that without a `setTimeout` fix the original issue earlier?

voloshink: It was removed because it ended up being a duplicate check with the one in slideDown, but no, it won't work without the timeout unfortunately, it will still think there's a window open

src/js/chat/templates.js
```diff
@@ -73,6 +73,12 @@ var jtvEmoticonize = exports.jtvEmoticonize = function(id) {
 };
 
 var emoticon = exports.emoticon = function(id, name) {
+    var fixedTwitchEmotes = [15, 16, 17, 18, 19, 20, 21, 22, 26, 27, 33];
+    id = parseInt(id, 10);
+
+    if (fixedTwitchEmotes.indexOf(id) > -1) {
```
night: more efficient to just parseInt here.. ie `.indexOf(parseInt(id, 10))` otherwise js has to keep type converting to string

Issue #931, opened 2016-01-28
Clickable @username links in the chat


dannoe:
Option to make usermentions ("@username") clickable in chat. For easy access to profiles or last messages sent.

dannoe:
@xgerhard In case you don't know already: You can use special [keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in your commit message to auto-close issues on merge.

PR #1206, opened 2016-05-24
Implements clickable @usernames (#931)

xgerhard:
Request #931 

If a word in a message starts with an '@', get that word. Check if that word is in the chatters list, if yes, add a span-class. The span on click function works with the same function as the 'double-click auto-complete'. Double click = insert username in chat input box, single click is open mod card.
The setting can be turned on/off.

Edited the 'double-click auto-complete' with just putting an '@' before the username in the chat input box. So the mentioned user will also be clickable.

Let me know what you think.


night:
Looks pretty good, but instead of only updating some nickname completions to include the `@`, you should also update the tab completion system to do the same.

xgerhard:
Agreed, I added the '@' to the tab completion system, however I ran into an issue while testing. 
Whenever you type an '@' the suggestion box pops up, now for example you have a suggestions list with the following values:
- KappaUser1 
- KappaUser2, 
- Kappa 
- KappaRoss
+ Your tab completion tooltip setting is turned off.

When you type 'kapp' and press tab (without the starting @), KappaUser1 will be suggested and @KappaUser1 will be placed in the input box, so far so good.
However if the user tabs again, the tooltip will pop up since the user starts with an '@'.
When the user tabs untill he reaches the first emote, the box dissappears again (since emotes don't start with an @).

I couldn't really find a solution to tackle this issue, since the user can also tab through the suggestion list when typing just '@'.
Therefore I couldn't really determine by the pressed keys if someone directly tries to mention somebody or is just tabbing suggestions.

One solution which could work, which I tested, is to not always make '@' popup the suggestion box, and make it a combined setting: 'Shows a tooltip with suggested names when typing @ or using tab completion'.
Maybe someone can point me in the right direction for a better solution.

EhsanKia:
Hmm, why do we even show tooltip on @ when tooltip is disabled? In my opinion there should be no tooltip. I don't think we need to combine them though.

Does removing the tooltip fix this issue?

thinkpad4:
As long as Tab Completion Tooltip  —  Shows a tooltip with suggested names when using tab completion isn't touched I'm happy

xgerhard:
@EhsanKia Yeah the current setting is only for the tooltip when pressing tab, pressing @ ignores this and will always pop up. Thats why I suggested making the 'tab completion tooltip' setting just 'completion tooltip' - on/off. 
What do you mean by removing the tooltip, just for pressing @? This will most likely fix it, but not sure if you want to remove that feature.

night:
The tooltip toggle was to match default Twitch functionality while giving it as an option for the BTTV tab completion users.. if we do this then we need to default it to on and allow it to be turned off

xgerhard:
Heya, my apologies for the late update.
So I changed the Tooltip setting to: Show tooltip when using @ and Tab, instead of just the tab (default on), this fixed my previous issue.

Now the question is, how much do we need to force the users to use @, for example @thinkpad4 says he doesn't like @ usage, so I made the @ suggestion depending on the userMentions setting.

The way I've setup the suggestion systems now is:

- emote 
- server command
= Both never suggest @

- userMention setting ON:
= Always suggest @

- userMention setting OFF:
= Nothing changed, only suggest @ when the user types @

My branch is a bit outdated, let me know what you guys think about above changes and I'll update to master.

thinkpad4:
I would prefer the @ option not be used at all when the setting is off.  Not even if I type @name it would just show up as @name no suggestion.   So when your userMention setting OFF even using a @name won't suggest anything and the tab compete will work.  The way the Tab Completion Tooltip functions with the setting off right now is how I would like it to continue to function.

xgerhard:
@thinkpad4 I'm trying to think with you, but your comment goes two different ways. So either I'm understanding you wrong or my explanation wasn't clear. Because how you describe it, is not how it currently works.

In my version there will be 2 settings that matter:
- Completion Tooltip (This used to be: Tab Completion Tooltip)
- UserMentions

When Completion Tooltip is off, there will be no Tooltip popup, even if you type an @ (which is new).
However if you tab, the input box will suggest you names/emotes like you are used to (just no tooltip).

When Completion Tooltip is enabled, it will pop up by typing an @ or using tab.

Now the new setting, UserMentions:
When enabled @name will be clickable in chat.
The difference to the tab suggestions is, if UserMentions is enabled it will always suggest an leading @ before the username. If not enabled it will only suggest an leading @ when the user types an @ (like you're used to).

Hope this clears it up, if I understand you wrong just let me know.

thinkpad4:
YES!!! When Completion Tooltip is off, there will be no Tooltip popup, even if you type an @ (which is new).
However if you tab, the input box will suggest you names/emotes like you are used to (just no tooltip).  That is what I meant.   Thank you,  Sorry, I had trouble putting into words what I meant.  Thank you very much.

EhsanKia:
Looks go to me. Only a small change, do you think it'd be cleaner if the @ sign was part of the clickable span? Kinda how GitHub does it @ehsankia 

Looks strange without it.

thinkpad4:
Ehsan as long as what he described is how it will work, I'm happy.

xgerhard:
@EhsanKia You're right, with the @ in the span it looks much cleaner.

I pulled the latest night-master and merged with my branch ([874c1a0](https://github.com/night/BetterTTV/pull/1206/commits/874c1a0807e5cc88fd2fd0b03cdfcaf4ed6122f4)), I'm not sure why it still says conflicts? I just recently started using GIT, did I merch it wrong?

EhsanKia:
There seems to be a merge conflict on betterttv.css from changes I've made since your change. It's a pretty easy cleanup, not sure why git client you use but if you're having issues, you can give me access to your BTTV fork and I'll fix the merge for you.

xgerhard:
Hmm, there indeed was a conflict when I tried to merge, so I fixed that in betterttv.css and pushed it. But if you could take a look  that would be great, you should have an invite.

EhsanKia:
Seems like you merged into your master, instead of the branch you were working on. I merged it for you, should work now.

xgerhard:
Ah, I see, thanks for the help!

night:
imho clickable usernames should not be a configurable option.. they should always be linked and open a profile card.

xgerhard:
@night I dont know, I would expect an @username to be clickable/linked to a profile, but since some people are maybe used to not have it for years, they dont want it.
So what do you think about the suggestion systems then, should a username thats suggested always have a leading @?

night:
put it this way, it should not be configurable. we have too many settings already, and I don't want another one for a feature which doesn't need to be toggle-able.

as for suggested usernames, since we're using `@` as a sentinel we'd need to prefix it always.

xgerhard:
Ok, should be good now then.

PR #1218, opened 2016-05-29
change chat background to match twitch change

asterius1:
Twitch recently changed background  of their chat from #f2f2f2 to #efeef1.
This pull request is changing the same color in BetterTTV chat

It doesn't change to much on twitch stream page, but ruins it for pages embeding twitch chat, and trying to mach surrounding background to Twitch chat background.

There is also a discrepancy between twitch's dark mode, and BetterTTV darkmode. I didn't dare to edit it because the difference is more significant. Is BetterTTV darkmode on purpose that much different from Twitch darkmode?

### DIFF
--- a/src/css/betterttv.css
+++ b/src/css/betterttv.css
@@ -52,7 +52,7 @@ ul.tabs li.selected a {
 .ember-chat .chat-messages {
     top: 0px;
     bottom: 114px;
-    background-color: #f2f2f2; /* hide group chat feature depends on having a background */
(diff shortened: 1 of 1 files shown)


EhsanKia:
Can you provide screenshots of before/after for all these things?

asterius1:
Background used to be #f2f2f2 :
background-betterttv-before

After change it is #efeef1:
background-betterttv-after

Just like on pure twitch right now #efeef1:
background-twitch

Issue #1220, opened 2016-05-31
Twitch chat in theater mode is not black


tbodt:
In theater mode, it looks like this:

<img width="343" alt="screen shot 2016-05-31 at 1 25 06 pm" src="https://cloud.githubusercontent.com/assets/5678977/15689307/36562854-2733-11e6-813d-7d1ce3ac3480.png">

EhsanKia:
Seems like #1218 broke this.

tbodt:
Seems likely.

night:
Fixed by #1224

PR #1224, opened 2016-06-01
Update controller name for detecting theatre mode

EhsanKia:
Fix for #1220 

Turns out it wasn't actually that guys fault haHAA, just Twitch changing stuff.


antyone:
I just checked and its fixed in live channels, but the problem still persists in videos of previous broadcasts.

PR #1227, opened 2016-06-03
Replace inline css with class for deleted and spam message

### LINKED ISSUE #1162: Change <message deleted> back to purple on light mode
The current light grey is very hard to read on light mode.

Grey

I propose returning back to twitch's default purple.

Purple

EhsanKia:
Setup for @teak to fix #1162


night:
classy




get it? haHaa

PR #1230, opened 2016-06-03
Added support for Warcraft tags

joshuaferrara:
Not sure if this is up to spec, but thought I'd give it a try. Added support for the new alliance/horde tags as well as fixed a formatting issue I was having.


joshuaferrara:
image

EhsanKia:
Thank you for the pull request, but yeah our solution makes more sense within the framework we have setup! Cheers.

PR #1256, opened 2016-06-10
Potential fix for chat not updating messages

EhsanKia:
Replaces `setInterval` with a `setTimeout` and the function calls itself again 250ms later.


night:
I am curious if we should just stop using a timer altogether for this.. what if we just did something more sane, like jQuery's throttle?

EhsanKia:
You're right, throttle implementation is cleaner and more likely to fix this. I just tried it and it works very well.

Issue #1267, opened 2016-06-11
Personal Emotes in Whispers


night:
Personal emotes should work in whispers.

PR #1268, opened 2016-06-11
Abstract tokenize for conversations and chatreplay

EhsanKia:
Fix for #1267 

This will reuse the same tokenize function as chat.
It will add support for personal emotes, and as a bonus, url image hover.

That being said, personal emotes won't show unless they are already in the store,
which only happens if you have seen the user in chat type.

So we'd need some websocket protocol to request an update for a specific user?

### DIFF
--- a/src/js/chat/templates.js
+++ b/src/js/chat/templates.js
@@ -193,7 +193,7 @@ exports.bttvElementTokenize = function(senderEl, messageEl) {
     var tokens = $(messageEl).contents();
     var sender = $(senderEl).text().trim().toLowerCase();
     for (var i = 0; i < tokens.length; i++) {
-        if (tokens[i].nodeType === Node.TEXT_NODE) {
+        if (tokens[i].nodeType === window.Node.TEXT_NODE) {
             newTokens.push(bttvMessageTokenize(sender, tokens[i].data));
         } else {
             newTokens.push(tokens[i].outerHTML);
--- a/src/js/chat/templates.js
+++ b/src/js/chat/templates.js
@@ -141,7 +141,7 @@ var bttvEmoticonize = exports.bttvEmoticonize = function(message, emote, sender)
 };
 
 var bttvMessageTokenize = exports.bttvMessageTokenize = function(sender, message) {
-    var tokenizedString = message.split(' ');
+    var tokenizedString = message.trim().split(' ');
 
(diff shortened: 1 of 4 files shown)


EhsanKia:
Hmm, not best solution, this will only broadcast whenever you view a new message. And there's currently no reparsing of messages in whispers, but it's a good initial implementation.

src/js/chat/templates.js
```diff
@@ -188,6 +188,21 @@ var bttvMessageTokenize = exports.bttvMessageTokenize = function(sender, message
     return tokenizedString.join(' ');
 };
 
+exports.bttvElementTokenize = function(senderEl, messageEl) {
+    var newTokens = [];
+    var tokens = $(messageEl).contents();
+    var sender = $(senderEl).text().trim().toLowerCase();
+    for (var i = 0; i < tokens.length; i++) {
+        if (tokens[i].nodeType === Node.TEXT_NODE) {
```
night: where is "Node" defined? is it on window? can we reference `window.` in that case to make it clearer

src/js/features/conversations.js
```diff
@@ -34,6 +33,9 @@ function Conversations(timeout) {
     if (window.App && App.__container__.lookup('service:whispers-shim')) {
         var whisperShim = App.__container__.lookup('service:whispers-shim');
         whisperShim.on('whisper', _self.onWhisper);
+        whisperShim.on('thread', function(data) {
+            bttv.ws.joinConversation(data.id);
```
night: You can join on thread, but you should broadcast on whisper.

Additionally you should handle the case of parting convos too.

night:
Abstractions seem nice 👍 for that

src/js/features/conversations.js
EhsanKia: the "whipser" event is only called when you **receive** one, I couldn't find a way to hook on whisper sent. That being sent, this "thread" event is called every time you **read**, so technically, it's equivalent, and better so if someone spams you 20 whispers, you don't broadcast 20 times, only once when you click on the conversation.

As for parting, I'll try to find something for the window closing but it might be hack-ish.

night: > called every time you read

can you comment that in, because that's totally not the connotation one would assume here.

PR #1272, opened 2016-06-14
Fixes #1265

### LINKED ISSUE #1265: Timer on the messaging
Hello,

This little bug I saw while spamming shit on the channel eulcs1 : The message "You are banned from talking in eulcs1 for 287 more seconds." makes the timer go for 1 second instead of the 287 it should be. 

I guess that's because you search for the first number in the string. Maybe taking the last, or the one between "for" and "more" would solve that. 

By the way, thank you for your addon.

MergHQ:
Hello,

So this solution might not be so good, but i didn't really think of anything better with regular expressions. The timeout duration should always be the 9th word in the message, because twitch doesn't allow spaces in usernames, at least yet.

Let me know if you have any ideas.

### DIFF
--- a/src/js/features/channel-state.js
+++ b/src/js/features/channel-state.js
@@ -135,10 +135,11 @@ module.exports = function(event) {
             var msg = event.tags['msg-id'];
 
             if (msg === 'msg_slowmode' || msg === 'msg_timedout') {
-                var matches = /([0-9]+)/.exec(event.message);
+                var matches = event.message.split(' ');
                 if (!matches) return;
 
-                var seconds = parseInt(matches[1], 10);
+                var seconds = 0;
+                if (matches.length >= 9) seconds = parseInt(matches[9], 10);
                 initiateCountDown(seconds);
             } else if (msg === 'msg_banned') {
                 initiateCountDown(86400);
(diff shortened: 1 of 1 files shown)


night:
This is not a good solution. One could easily fix the existing regex by enforcing a prefixing space.

`/\s([0-9]+)/`

MergHQ:
Thanks for the hint. Should be gucchi now :+1:

src/js/features/channel-state.js
```diff
@@ -135,10 +135,10 @@ module.exports = function(event) {
             var msg = event.tags['msg-id'];
 
             if (msg === 'msg_slowmode' || msg === 'msg_timedout') {
-                var matches = /([0-9]+)/.exec(event.message);
+                var matches = /\s([0-9]+)/.exec(event.message);
                 if (!matches) return;
 
-                var seconds = parseInt(matches[1], 10);
+                var seconds = parseInt(matches[0], 10);
```
night: why did this line change?

MergHQ: @night No reason why it changed. I just modified it by mistake. Should I fix it?

PR #1288, opened 2016-06-20
Badges API support

EhsanKia:
Initial implementation of Badge API support


night:
is left to do still accurate?

EhsanKia:
Okay, should be complete now. Right now Twitch already injects CSS for subscriber badges, but god knows what they'll do once they switch, so just to be sure, we inject our own CSS for subscriber badges for partner channels, in case they remove it.

For click behavior, I append a data-click-action to the badge which is read on click. The only hacky part is for custom URLs, I get the badge info from the CSS class and use that to get the badge info.

It's also hard to test as there's currently no more custom badges. I currently inject warcraft badge on myself, and so far it all seems to work smoothly.

night:
I will review this over the weekend

Issue #1287, opened 2016-06-20
"@" symbol before name with tab completion even when I didn't start with "@"


AlcaDesign:
Tab completing a name forces the "@" character even though I didn't start with an "@" as it's my preference to not use the "@" symbol.

![Example](https://i.gyazo.com/6d273acd4a8347752efe0f9bb4e83058.gif)

EhsanKia:
This is part of BTTV's new user mention feature, which highlights usernames in chat and makes them clickable. It's a decision we took to just prepend @ sign instead of adding yet another option. Many platforms use this format, such as twitter or github, ie, @AlcaDesign. Default twitch also uses this behavior.

At the end of the day, it may be a bit difficult to adapt to change at first, but in the long run we believe this is a cleaner solution.

AlcaDesign:
It doesn't have to be an option, just recognize that if I don't put a "@" then don't complete with one.

EhsanKia:
This is now fixed by 9bf333e6926c50b2d759dfd7f9ec10d90fc267ac

PR #1289, opened 2016-06-20
Only use @ when tab complete starts with @

EhsanKia:
Fix for #1287


xgerhard:
If we want this @ check back, then I think [L151](https://github.com/night/BetterTTV/blob/master/src/js/chat/helpers.js#L151) in the on click function would need that check back aswell.

night:
@xgerhard then PR it

PR #1293, opened 2016-06-20
@ onclick check

xgerhard:


EhsanKia:
Can you explain what this does, I'm not completely sure what's going on there.

xgerhard:
It's the same change you did earlier with the leading @ issue with suggestions. When you click a suggestion from the list this on click function is called and will always suggest @. 



(The same could be applied to DoubleClick complete feature, but I dont know if this and your fix earlier is the final decision of handling the suggestions)

EhsanKia:
Ah, I see. For double click, I think appending @ is fine.

Issue #929, opened 2016-01-27
[Feature Request] Viewer count while in Theater Mode


adispen:
Been messing with this idea for a while since realizing you can't check viewer count in theater mode.  Wrote up some JS that allows you to see the current viewers while moused over the chat header.  When moused over, the title changes to the viewer count, and changes back when you mouse out.  Script is below

```javascript
var chan = $(".chat-header .room-title").text();
$(".chat-header")
.mouseover(function(){
  $(".chat-header .room-title").text(function(){
    return $(".live-count").text() + " viewers";
  });
})
.mouseout(function(){
  $(".chat-header .room-title").text(function(){
    return chan;
  });
});
```

night:
We would rather not clutter the chat UI with elements unrelated to chat. You're welcome to pull request a change adding the count into the video player UI, but I don't believe this is a good solution as-is.

PR #1334, opened 2016-07-01
Viewer Count in Player

adispen:
Added viewer count to player as request in #929 by @night.  It replaces the "LIVE" text in the top right.  Meant to make it possible to see how many viewers there are while in theater mode.

### DIFF
--- a/src/js/features/live-viewers.js
+++ b/src/js/features/live-viewers.js
@@ -1,18 +1,22 @@
 var debug = require('../helpers/debug');
 
 function checkBind() {
-    $.each($('.live-count').data('events'), function(i) {
-        if (i === 'DOMSubtreeModified') {
-            return true;
-        }
-    });
-    return false;
+    var isBound = false;
+    if ($('.live-count').length) {
+        $.each($('.live-count').data('events'), function(i) {
+            if (i === 'DOMSubtreeModified') {
+                isBound = true;
+                return;
+            }
+        });
+    }
+    return isBound;
 }
 
 
 module.exports = function() {
     if (bttv.settings.get('showViewersInPlayer') === true) {
-        if (checkBind === true) {
+        if (checkBind() === true) {
             $('.live-count').unbind('DOMSubtreeModified');
         }
         debug.log('Replacing "LIVE" with viewer count');
@@ -28,7 +32,7 @@ module.exports = function() {
 
     if (bttv.settings.get('showViewersInPlayer') === false) {
         debug.log('Restoring "LIVE" in the player');
-        if (checkBind === true) {
+        if (checkBind() === true) {
             $('.live-count').unbind('DOMSubtreeModified');
         }
         $('span.js-live-label').text(function() {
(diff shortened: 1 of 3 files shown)


EhsanKia:
Do you have a screenshot? Also maybe it'd be better if it was a second line under "LIVE" rather than completely replace it?

adispen:
Here's a screenshot. Let me know if you think I should make it two lines.

src/js/features/live-viewers.js
```diff
@@ -0,0 +1,38 @@
...
+            return true;
+        }
+    });
+    return false;
+}
+
+
+module.exports = function() {
+    if (bttv.settings.get('showViewersInPlayer') === true) {
+        if (checkBind === true) {
```
EhsanKia: Shouldn't you be calling this function?

EhsanKia:
Not sure what other people think, but personally I think it'd be a better solution to just have it  automatically work in theater mode, rather than having a whole option for this little feature.

NBruderman:
I Agree with @EhsanKia here... It's only needed on theather mode, and the appearnce is already enough to get the point that the broadcast is live.

EhsanKia:
You could just add a span under it with the viewer count updating, like so:


And use CSS to hide it when not in theater mode. A `.theatre` class is added to the channel container when you enter theater mode, so you can use that to only show the span when needed.

NBruderman:
@EhsanKia Looks amazing, I'm really up for it!

adispen:
Please excuse the typo in the commit. Here is an updated screenshot. 
liveviewerscapnewcrop

adispen:
Anyone interested in trying this out locally can copy the JS in [this gist](https://gist.github.com/adispen/7e5f6521863e7ae6dbbef1e2a187b09c) into their browser.

NBruderman:
@EhsanKia Any chance this will be implemented soon?

night:
Things only get implemented when they get code reviewed. Sorry for the delay.

src/js/features/live-viewers.js
```diff
@@ -0,0 +1,66 @@
...
+                var attributeValue = $(mutation.target).prop(mutation.attributeName);
+                if (attributeValue.includes('theatre')) {
+                    enableCount();
+                } else {
+                    disableCount();
+                }
+            }
+        });
+    });
+    observer.observe($div[0], {
```
night: A mutator on the entire page scope is *not* ok. You should hook into Twitch's Ember like we do in main.js to know when features should be executed.

src/js/features/live-viewers.js
```diff
@@ -0,0 +1,66 @@
...
+    $('#viewerSpan').text(function() {
+        return $('.live-count').text() + 'Viewers';
+    });
+
+    if (checkBind() === true) {
+        $('.live-count').unbind('DOMSubtreeModified');
+    }
+
+    debug.log('Creating viewer count span');
+    $('.live-count').bind('DOMSubtreeModified', function() {
```
night: `DOMSubtreeModified` is deprecated and should be replaced with mutation observers. You should also be able to drop the `unbind` since garbage collection should take care of the mutation observer when the watched element is removed.

PR #1361, opened 2016-07-09
More fixes

EhsanKia:
Bunch more fixes

- Make /join and /part work along side anonchat setting
- Fix BTTV spammers loading from API
- Added "/followed" command to get follow age
- Fix svg for leaving group chat not showing up
- Fix BVL icon not being centered
- Add chatters count to BVL

### DIFF
--- a/src/js/chat/helpers.js
+++ b/src/js/chat/helpers.js
@@ -445,7 +445,7 @@ exports.sendMessage = function(message) {
             return;
         }
 
-        if (bttv.settings.get('anonChat') === true) {
+        if (bttv.chat.store.isAnonMode === true && message !== '/join') {
             serverMessage('You can\'t send messages when Anon Chat is enabled. You can disable Anon Chat in the BetterTTV settings.');
             return;
         }
--- a/src/js/chat/helpers.js
+++ b/src/js/chat/helpers.js
@@ -977,6 +977,16 @@ exports.translate = function($element, sender, text) {
     });
 };
 
+exports.followDate = function(user, channel) {
+    bttv.TwitchAPI.get('users/' + user + '/follows/channels/' + channel).done(function(data) {
+        var m = moment(data.created_at);
+        reply = 'Followed ' + m.fromNow() + ' (' + m.format('LLL') + ')';
+        serverMessage(reply, true);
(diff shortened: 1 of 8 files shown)


src/js/features/better-viewer-list.js
```diff
@@ -21,12 +22,18 @@ function renderViewerList() {
...
             if (results[i].filter && results[i + 1].text === ' ') i++;
             else tmpResults.push(results[i]);
         }
         results = tmpResults;
+    } else {
+        count = chatterCount;
     }
     viewList.render(results);
+
+    var label = count + (count < 2 ? ' viewer' : ' viewers');
```
night: 1. Your counting system seems pointless. Why not just `results.length`?
2. "chatters" not "viewers"

src/js/chat/helpers.js
```diff
@@ -977,6 +977,16 @@ exports.translate = function($element, sender, text) {
     });
 };
 
+exports.followDate = function(user, channel) {
+    bttv.TwitchAPI.get('users/' + user + '/follows/channels/' + channel).done(function(data) {
+        var m = moment(data.created_at);
+        reply = 'Followed ' + m.fromNow() + ' (' + m.format('LLL') + ')';
```
night: missing `var` ?

also maybe make the response more verbose, include the channel name for clarity, maybe accept specifying a channel as a param, default to current channel

src/js/features/better-viewer-list.js
EhsanKia: The problem is that the list includes headers such as "MODERATORS" or "ADMINS" and I can't predict how many of those there are in the list.

The chatter/viewer thing really bugs me too, the issue is, it's called "viewers list" even though it's supposed to be "chatters list", but yeah I'm stuck between changing everything to chatter, or having it be inconsistent. Both solutions kinda suck.

night: Why are those in the list?

EhsanKia:
That's just the data structure I'm using. Trust me it's the cleanest way to do it. It's a linear pass that's happening anyway to filter so there's no extra cost to count anyway.

PR #1366, opened 2016-07-11
Option to hide Hosted and Recommended streams/videos

DennisMellegaard:
Couldnt find it in the current version of BTTV, so devided to add it in.

Basic funcitonality to unclutter your directory.following tabs. Really wish i could jump in before they print it out. But alas i did not find a way.

Settings:
- Enable/disable hiding of Hosted Streams
- Enable/disable hiding of Recommended Videos

Functionality:
- Hiding based on data-tt_content attribute

### DIFF
--- a/src/js/features/hide-hosts.js
+++ b/src/js/features/hide-hosts.js
@@ -1,0 +1,23 @@
+var debug = require('../helpers/debug');
+
+module.exports = function() {
+    if (bttv.settings.get('hideHostedStreams') !== true) return;
+
+    debug.log('Hiding hosted');
+
+    var hosted = $('*[data-tt_content="live_host"]');
+    if (hosted.length > 0) {
+        var hostedContainer = hosted[0].closest('.js-streams');
+        var hostedTitleContainer = $(hostedContainer).prev();
+
+        $(hostedTitleContainer).remove();
+        $(hostedContainer).remove();
+    } else {
+        var emptyGrid = $('.empty-grid');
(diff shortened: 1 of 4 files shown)


EhsanKia:
I don't think it's a good option to spam the settings page with a bunch of low impact settings like this. Before we know it, it'll be cluttered and unusable.

A maybe better solution is to hide a tiny "hide" button next to each section on the follower page itself, and have it remember the state?

night:
Let's not do any of this. As Ehsan said, we can't add options for everything or we'd have hundreds of options. Feature creep for customization is bad right now. Feature creep for actual features is okay. If you only want live channels, use the pre-existing "Default to Live Channels" option.

We plan to eventually release a new settings page. When that's in perhaps we can explore more customization of features (i.e. checkboxes/selections).

EhsanKia:
But it's inherently very bad design to put these in options.

There are 4 kinds of content on that page: followed channels, hosted channels, followed games, recommended videos.

Each user may be interested in ANY combination of the above. By switching to one of the top tabs as the current "Default to Live Channels" does, you're basically just allowing for ONE out of 15 combinations, and you're also wasting a configuration spot in the settings list.

The default follow page shows all 4 categories right there. My solution puts a tiny toggle button next to each so that you can collapse the section. It's very easy to toggle right there on the page, rather than go digging in settings, and allowed for people to pick and choose any of the 4 sections they want.

UX wise, this is a win all around. Users have more options, toggling is faster, and settings are less cluttered.

Issue #1364, opened 2016-07-11
Scrollback chat amount does not apply to chat replay


NBruderman:
The "Set Scrollback chat amount" is really useful to decrease lag in my place, but for some reason, the amount I've set on a live chat, which works perfectly, does not work on chat replay, which causes my frame drops. Can you fix it to work on both cases (Live and replay chat)?

night:
closing this since it cannot be fixed without modifying Twitch's logic, which we won't do.

PR #1371, opened 2016-07-12
Chat replay scrollback fix

MergHQ:
Hello,

This was requested in #1364 so thought I'd implement it. If there's anything I could improve then please let me know!


night:
does Twitch not manage the messages themselves?

MergHQ:
@night Nope, I set the scrollback value to 10. Worked for live chat but didn't affect replay chat.

night:
You should fix this to check once per loop, not in each iteration within the loop

EhsanKia:
But twitch definitely has a built in limit right? This will allow us to set a lower limit, but if you set it higher that the twitch limit, twitch will still delete the messages right? 

Also make sure while testing that you actually reach Twitch's limit to see if us deleting messages breaks anything with their limiter.

NBruderman:
@night What do you think?

night:
Have you tested this code? It doesn't work. Additionally it wouldn't count for the case where multiple lines are added at the same time. That and what Ehsan said about checking if Twitch errors on removed elements. Your fix also doesn't work for the case where the line limit is larger than what Twitch allows.

MergHQ:
I'll continue testing when I'm home. Currently unable to.

MergHQ:
Twitch limits chat replay messages to 100, so this will not work for limits above 100. It looks like removing elements won't affect the limiter twitch has.

night:
You should see if there is a way to increase Twitch's limit, and if so, use it for both increasing and decreasing the max count.

EhsanKia:
Researched it a bit. it's the vod-chat-service taking care of deleting lines. It's 100 when you're at the bottom, 200 when you're scrolled up. There's doesn't seem to be a way to change it since it's values in the closure. Can't really think of any way to get around it.

MergHQ:
@EhsanKia Could somehow merge the rechat requests to allow more messages? The site makes requests to rechat.twitch.tv to get the upcoming chat lines. We'd have to completely remake replay chat though.

EhsanKia:
Yeah, but rewriting the whole rechat system just to get a higher limit is kinda ridiculous. Don't forget you also need to handle being able to delete chat lines as mod and these other even handlers too. Just a huge pain and probably not worth it.

night:
Then I don't see why we really care about enforcing this limit for lower values.. I think the reason we made a configurable option in the first place was so you could set large amounts, not smaller amounts.

NBruderman:
Actually not. I opened the issue to apply lower values, as the stock values causes frame drops. Setting lower values fixes the problem, so I want to apply them in the replay too. I don't see a reason to increase it, as it's a replay, and you can just rewind back whenever you want. For me, it's a system resources issue.

EhsanKia:
100 lines causes frame rate drops? That doesn't sound very likely to be honest. What do you set it to on normal chats?

NBruderman:
Around 40... And in very active chats (like big tournaments) it's really effective. And it has the same effect in the replay chat. I opened this issue after I've tried to watch a CS GO Major replay.

EhsanKia:
That sounds really strange. Have you tried the new HTML5 player for turbos by any chance? It seems more like one of the bugs they had with the Flash player.

NBruderman:
I'll gladly try that. How to do so?

EhsanKia:
Well, it's only available to turbo users right now, unfortunately, but you simply turn it on in the player options:

NBruderman:
Well, I can tell you that my problem is well tested, and decreasing the history does solve it. Until the new player is available for testing, I just want to set the same amount on the replay chat. As and I say, the increasment factor is less important in the replay, as you can just rewind few seconds back, and watch it again.

src/js/features/chat-replay.js
```diff
@@ -36,6 +36,13 @@ ChatReplay.prototype.connect = function() {
                     this.messageParser(el);
                 }
             }
+
+            var chatLines = $('.chat-line');
+            if (bttv.settings.get('scrollbackAmount') < 100 && chatLines.length > bttv.settings.get('scrollbackAmount')) {
```
night: why are you getting this value three times in 2 lines? use a local variable..

src/js/features/chat-replay.js
```diff
@@ -36,6 +36,14 @@ ChatReplay.prototype.connect = function() {
                     this.messageParser(el);
                 }
             }
+
+            var chatLines = $('.chat-line');
+            var scrollBackAmount = bttv.settings.get('scrollbackAmount');
+            if (scrollBackAmount < 100 && chatLines.length > scrollBackAmount) {
```
night: this if statement seems superfluous?

revise your for loop to

```
for (var i = chatLines.length-1; i >= scrollBackAmount; i--)
```

night:
closing due to inactivity

Issue #366, opened 2015-04-22
Disable Offline Banner


ThorChayzarr:
Assuming it would be possible, I would like to see an option to disable the new "offline" banners that Twitch has implemented. They're finicky and I don't really like the look of them.

Cheers,
Thor

Yrlish:
I agree, I don't like it. It is annoying and just moving the player down. Also it doesn't remove itself when a streamer comes online so that makes it even more annoying.

night:
It seems Twitch has worked to fix the bugginess. Are you still experiencing issues?

I'm not sure that these are actually posing that much of an issue, since a link to the latest broadcast might entice people to watch a VoD.

ThorChayzarr:
I haven't been experiencing any real issues or bugs with it anymore, but it still seems a bit unnecessary to have, in my opinion, and I've talked to a lot of people who agree. Just my two cents, I guess, but it's cool if you don't find it to be a big enough issue to work on.

Yrlish:
I haven't been on Twitch the last few days so I don't know if there has been any changes. But I agree with what @ThorChayzarr says about it. The feature is unnecessary and an option to remove it in BTTV settings would be great.

Yrlish:
Any words on this? I still get annoyed by that message above the player. I'm not interested in that information.

A setting to turn this off would be amazing. Default it to shown, but let me be able to hide it.

Yrlish

Yrlish:
I've tried fixing this by myself multiple times. But I cannot setup the development environment unfortunately. Either Firefox or Chrome lets me use the self-signed certificate, seems that HSTS blocking it? But not sure how it works or why, but I'm sure it's the certificate.

Either, can someone fix this or help me set up the dev environment.

It shouldn't be too hard. Just an option in settings, defaulted to false and this code in clear-clutter.js should be enough, with the setting check around it of course. Couldn't test by myself due to above.
```
// if setting for "hide Watch <name>'s Last Broadcast!" is true
$('body').append('<style>.ember-view > .recent-past-broadcast { display: none !important; }</style>');
```

Yrlish

Yrlish:
I'm still annoyed by this banner thing, because it doesn't always go away. And if I'm interested in the VODs, I will go through the channel profile myself and watch them.

EhsanKia:
[...]

PR #1390, opened 2016-07-20
Option to hide Last Broadcast banner

Yrlish:
Finally solves #366

### DIFF
--- a/src/js/features/clear-clutter.js
+++ b/src/js/features/clear-clutter.js
@@ -13,4 +13,9 @@ module.exports = function() {
         removeElement('#nav_related_streams');
         $('body').append('<style>#nav_games, #nav_streams, #nav_related_streams, .js-recommended-channels { display: none !important; }</style>');
     }
+
+    // if setting "Hide Last Broadcast Banner" is true
+    if (bttv.settings.get('hideLastBroadcast') === true) {
+        $('body').append('<style>.ember-view > .recent-past-broadcast { display: none !important; }</style>');
+    }
 };
--- a/src/js/settings-list.js
+++ b/src/js/settings-list.js
@@ -429,6 +429,12 @@ module.exports = [
         }
     },
     {
+        name: 'Hide Last Broadcast Banner',
+        description: 'Hide the Last Broadcast banner above the player',
+        default: false,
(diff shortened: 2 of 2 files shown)


src/js/features/clear-clutter.js
```diff
@@ -13,4 +13,9 @@ module.exports = function() {
         removeElement('#nav_related_streams');
         $('body').append('<style>#nav_games, #nav_streams, #nav_related_streams, .js-recommended-channels { display: none !important; }</style>');
     }
+
+    // if setting "Hide Last Broadcast Banner" is true
+    if (bttv.settings.get('hideLastBroadcast') === true) {
```
night: this is not the correct location for this.

additionally, we do not place css within js any longer.

src/js/settings-list.js
```diff
@@ -429,6 +429,12 @@ module.exports = [
         }
     },
     {
+        name: 'Hide Last Broadcast Banner',
```
night: instead of making a setting in our menu, why not improve upon twitch's feature by adding a close button to their banner that remembers the setting

src/js/features/clear-clutter.js
Yrlish: Suggestions for better location? 

And for the CSS in js, I mainly did it because I saw it elsewhere and also to modify the minimal amount of files.

src/js/settings-list.js
Yrlish: That sounds like a good idea, but what if a user wants it back? If it's hidden forever, where would a good location for it be to make it visible again?

src/js/features/clear-clutter.js
night: > Suggestions for better location?

It should be in its own module.

> And for the CSS in js, I mainly did it because I saw it elsewhere and also to modify the minimal amount of files.

That's because we do have legacy code in some places.. BetterTTV has been around since 2011 so there's straggling old code around, unfortunately.

src/js/settings-list.js
night: Then have it remember it for the single page view. As long as the user can close it when it opens, there should be no reason why they can't close it instead of complaining it's there.

night:
closing due to inactivity

PR #1420, opened 2016-08-01
Adding Scrollbars to Whispers

Aaron128l:
Also added support to whisper emote menu.

### DIFF
--- a/src/css/betterttv-dark.css
+++ b/src/css/betterttv-dark.css
@@ -1968,21 +1968,21 @@ span.subscribe-text {
 }
 
 /* Twitch emote menu */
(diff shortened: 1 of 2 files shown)


night:
scrollbars weren't already a thing?

EhsanKia:
The problem is we have to add the custom scrollbar CSS for every new scrollbar. I think honestly we can just apply that to every single scrollbar on the site and be fine. The custom scrollbar looks good.

night:
oic.. well I don't think global css is the right idea since twitch might also use these scrollers on rails pages, which would break.. prefer to keep things whitelisted

night:
resolve the conflicts and I can merge

PR #1446, opened 2016-08-13
More blue buttons! 

ScaniaTV:
- Fixed all the other buttons that I found that were not blue.

- Also fixed the emote set button for turbo users being white with dark mode.

old

new


src/css/betterttv-blue-buttons.css
```diff
@@ -212,3 +213,55 @@ div#header_actions a.primary span {
 .recent-past-broadcast {
     border-left: 5px solid #2F58B9;
 }
+
+/** Emote set button for turbo users when selected. */
```
night: `**` is not needed for all your comments, just one `*`

ScaniaTV:
Fixed the comments in 8008078

night:
Had to revert this because there are too many bugs. You need to test better.

PR #1474, opened 2016-08-24
Filter repetitive

gabranches:
Hello, 
Big fan of the extension. Thanks for all your work!

I created a setting that allows a user to filter repetitive chat messages. It is also able to ignore random strings that are used to get around r9k mode. 

Here's a few examples:

**Example 1:**
Input: SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa 
Output: SPAM Kappa

**Example 2:**
Input: 123 SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa 
Output: SPAM Kappa

**Example 3:**
Input: SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa SPAM Kappa 123
Output: SPAM Kappa

**Example 4** (note: a message with 3 repetitions or less is ignored):
Input: Kappa Kappa Kappa
Output: Kappa Kappa Kappa

**Example 5:**
Input: Kappa Kappa Kappa Kappa 
Output: Kappa

I added the main interface function to **templates.js**, but I'm not sure if that is the best place for it. I also added the feature to **settings-list.js**.


aycgit:
i would suggest a option to let lit like display "Kappa (x5)"

night:
Do you have any profiling on resource usage in large channels with this enabled? We are hesitant to add any filtering to BetterTTV because of resource usage.

src/js/chat/templates.js
```diff
@@ -334,6 +334,11 @@ var message = exports.message = function(sender, msg, data) {
         msg = tokenizedMessage.join(' ');
     }
 
+    if (bttv.settings.get('filterRepetitive')) {
```
night: test

PR #1495, opened 2016-09-02
quick fix

smackedlol:
#emote-menu-button was using an !important tag overidding the chat-buttons-container. This just removes the important tag and keeps the same color.


night:
Can you explain what this fixes?

smackedlol:
It just changes the background color of the emote menu from rgb(30,30,30) to rgb(25,25,25).

PR #1509, opened 2016-09-09
replace deprecated gulp-browserify with browserify

KenanY:
The `gulp-browserify` module has been [deprecated](https://github.com/deepak1556/gulp-browserify/blob/db0f8c151458703ea2eedd30f651ae2e2e912475/README.md) for over a year now, in favor of simply using plain `browserify`. Since `browserify` already works with streams, it works perfectly alongside other gulp plugins like `gulp-header` and `gulp-uglify`, provided that we convert the stream to a `vinyl-buffer`.

This PR implements just that. I added the new dependencies under [`"devDependencies'`](https://docs.npmjs.com/files/package.json#devdependencies), since that's where build dependencies like these are intended to go. I can send another PR to move the rest of the `gulp` stuff there too, if that's desired.

Tested locally and everything appears to work, but let me know of any issues you encounter :)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,15 +1,16 @@
 var fs = require('fs'),
     gulp = require('gulp'),
     jade = require('gulp-jade'),
-    browserify = require('gulp-browserify'),
+    browserify = require('browserify'),
     header = require('gulp-header'),
     footer = require('gulp-footer'),
-    concat = require('gulp-concat'),
     del = require('del'),
     eslint = require('gulp-eslint'),
     uglify = require('gulp-uglify'),
     saveLicense = require('uglify-save-license'),
-    enviro = require('gulp-environments');
+    enviro = require('gulp-environments'),
+    source = require('vinyl-source-stream'),
+    buffer = require('vinyl-buffer');
 
 gulp.task('cleanup', function() {
     return del('build/**/*');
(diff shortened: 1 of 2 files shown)


night:
More than just browserify is deprecated. Perhaps you want to upgrade all the OOD deps. Additionally stuff in dev deps currently breaks our auto release machine.

KenanY:
@night Alright, I'll move them to `"dependencies"`. I can update the other dependencies in future PRs (in the interest of keeping each PR's changeset as minimal as possible).

KenanY:
Commit amended and force-pushed.

NBruderman:
@night Is this ready?

night:
my original feedback still stands.. I'd rather all deps be upgraded in a fell swoop than one of them

KenanY:
Alright, will do if that's what's desired.

KenanY:
Commit amended and force-pushed. Now contains the following changes:

  - replaced deprecated `gulp-browserify` plugin with `browserify`
  - removed dependency on `gulp-changed` since it is not being used anywhere
  - `gulp-eslint@^3.0.1`
  - `gulp-jade@^1.1.0`
  - `gulp-uglify@^2.0.0`
  - `lodash.throttle@^4.1.1`
  - `request@2.75.0`

KenanY:
Looks like the build failed because the `del` module, which I didn't touch here, wasn't installed for some reason. Can't reproduce locally with a `rm -rf node_modules/ && npm install` so might just be an issue with the cached `node_modules/` on travis.

PR #1579, opened 2016-10-01
Host Button Fix

Aaron128l:


src/js/features/host-btn-below-video.js
```diff
@@ -17,10 +17,10 @@ module.exports = function() {
     var $hostButton = $('#bttv-host-button');
 
     if (!$hostButton.length) {
-        $hostButton = $('<span><span></span></span>');
-        $hostButton.addClass('button').addClass('action');
+        $hostButton = $('<button><span></span></button>');
+        $hostButton.addClass('button').addClass('action button--hollow mg-l-1');
```
Aaron128l: Felt it was better to use Twitch's CSS. Otherwise it'd be hardcoded with `#bttv-host-button`.

PR #1634, opened 2016-10-10
mods cards, badges, logic and memes

teak:


src/css/betterttv.css
```diff
@@ -190,10 +190,11 @@ input.text:focus,select:focus,textarea:focus {
     background-color: #0c6f20;
 }
 
-.badges .twitch-premium-1.alpha {
+.badges .twitch-premium-1 {
```
night: The old fix is equivalent, and doesn't hardcode the color..

PR #1650, opened 2016-10-13
Add an option to disable Chat Freeze

HiDeoo:
This change allows users to enable or disable the Chat Freeze feature (enabled by default like it's the case at the moment).

_Reason behind this change:_ this feature is sometimes involuntary triggering for me when cmd-tabbing (Mac alt-tabbing) while moving at the same time the mouse off the window.
I can reproduce this issue 100% of the time on Mac. I can't say anything about Windows, I don't have one to test it.

Anyway, having an option to disable this feature (while keeping it enabled by default) could be a nice improvement.

### DIFF
--- a/src/js/features/chat-freeze.js
+++ b/src/js/features/chat-freeze.js
@@ -7,10 +7,10 @@ module.exports = function() {
     if (isLoaded) return shouldFreeze;
 
     $('body').on('keydown.chat-freeze', function(e) {
-        if (!(e.metaKey || e.ctrlKey) || !$('.chat-room:hover').length) return;
+        if (bttv.settings.get('chatFreeze') !== true || !(e.metaKey || e.ctrlKey) || !$('.chat-room:hover').length) return;
         shouldFreeze = true;
     }).on('keyup.chat-freeze', function(e) {
-        if (e.metaKey || e.ctrlKey) return;
+        if (bttv.settings.get('chatFreeze') !== true || e.metaKey || e.ctrlKey) return;
         shouldFreeze = false;
         helpers.scrollChat();
         $('.chat-room .chat-interface .more-messages-indicator').click();
--- a/src/js/features/chat-freeze.js
+++ b/src/js/features/chat-freeze.js
@@ -4,6 +4,8 @@ var shouldFreeze = false;
 var isLoaded = false;
 
 module.exports = function() {
+    if (shouldFreeze === true && bttv.chat.store.activeView !== true) shouldFreeze = false;
+
     if (isLoaded) return shouldFreeze;
 
     $('body').on('keydown.chat-freeze', function(e) {
--- a/src/js/features/chat-freeze.js
+++ b/src/js/features/chat-freeze.js
@@ -7,10 +7,10 @@ module.exports = function() {
     if (isLoaded) return shouldFreeze;
 
     $('body').on('keydown.chat-freeze', function(e) {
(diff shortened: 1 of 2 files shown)


voloshink:
So what's the issue being caused by accidentally freezing the chat when cmd tabbing

HiDeoo:
I want to be able to cmd tab to another program and always be sure that chat is not frozen. 

At the moment, I have to double check every time it's actually not frozen and I can't count the number of time I cmd tab, thinking chat is up-to-date on my second monitor but it's actually frozen.

I get that some people might want the chat to be accidentally frozen when cmd tabbing but I don't think it's the case for everyone, at least it's not my case so having an option and still keeping the default behaviour you choose is I think a good compromise.

voloshink:
To me a better solution is:

1) Showing the "more messages below" indicator when chat is frozen (no idea why we still haven't added that, i suggested that a while ago)

2) Fixing the bug where chat stays frozen if you tab out of the window

It seems like a better solution is actually fixing the bugs and making the feature better rather than making an option to disable a good feature just because it's buggy.

HiDeoo:
"Fixing" the feature was actually my first idea too, but I couldn't imagine a way to distinguish between the following two cases:

* People cmd tabbing and moving the mouse of the window at the same time and expecting chat to be frozen.
* People cmd tabbing and moving the mouse of the window at the same time and not expecting chat to be frozen.

If anyone has an idea on how to do that while not forcing a not expected behavior on some people, I'll be happy to make a new PR implementing that.

jonathandelgado:
+1 to either remove this or bind it to a more complex key combination to avoid this accidentally being triggered.

EhsanKia:
Fixing this is definitely a far better solution than adding an option.
I believe it's a fairly easily fix. I remember we had a system setup to detect if the window is in focus. So we simply need a check for the chat freeze to make sure the user has the window focused.

HiDeoo:
Unfortunately `bttv.chat.store.activeView` won't do the trick.

When this behaviour happens while cmd tabbing, the `keyup.chat-freeze` event is not even triggered since the window is no longer the active one and that's why we're staying in the 'frozen' mode.

If the idea is to never freeze the chat when the window is not focused, one solution could be to have another detection if the window is getting unfocused in this module and unfreeze the chat if frozen when it happens.

Another approach could be to reuse the current detection but that would need some refactoring. 

I'm not sure what would be the preferred way / better approach to do that in BetterTTV.

EhsanKia:
What if you put it here:
https://github.com/night/BetterTTV/blob/5d36d9b0ec63f9dbf58979f028354adf9957f93c/src/js/chat/helpers.js#L585

HiDeoo:
Nice catch @EhsanKia, I didn't know this was also called for every message received.

If the idea is to always unfreeze the chat if frozen when the window is not the active one, we can use `bttv.chat.store.activeView` there to manually disable it.

I pushed a fix reversing my previous changes and adding these new ones that fix the issue. Of course, this does not affect the chat being paused manually (scrolling).

PS: Should I make a new pull request since it's no longer an option?

EhsanKia:
Looks much better. You can keep the PR, if it really bugs you, you can rename it to reflect the actual change. Your change looks good to me, if you confirmed yourself that the bug doesn't happen anymore for you with this change.

HiDeoo:
Not at all, the question was much more about the preference of the maintainers of the repo and rules regarding PR.

And yes, I have heavily tested it today (on Mac), not happening anymore, as soon as the window is no longer active and if chat is frozen, it automatically unfreeze.

PR #1654, opened 2016-10-14
Add setting to disable localized names in chat

EhsanKia:


src/js/chat/handlers.js
```diff
@@ -523,6 +523,9 @@ var privmsg = exports.privmsg = function(channel, data) {
         from = data.bttvDisplayName;
     } else {
         from = helpers.lookupDisplayName(data.from);
+        if (bttv.settings.get('disableLocalizedNames') === true && from.toLowerCase() !== data.from) {
```
night: I'd prefer you modify `lookupDisplayName` to disable these special display names for this case (need to allow nicknames but block twitch display names)

EhsanKia: You are right that putting it inside there is cleaner, but I'm fairly sure my fix didn't break nicknames, as those happen in the other case of the if statement, I think?

EhsanKia: Nvm I guess I was wrong, moving it either way.

PR #1656, opened 2016-10-14
Fix sub badges not updating

EhsanKia:
Sub badges weren't properly updated when you switched between channels, so I separated sub badges from twitch badges

### DIFF
--- a/src/js/chat/helpers.js
+++ b/src/js/chat/helpers.js
@@ -710,21 +710,32 @@ exports.loadTwitchBadges = function() {
             store.__twitchBadgeTypes[badge] = badgeData;
         });
 
-        if (tmi().channel && tmi().channel.partner === true) {
-            $.getJSON('https://badges.twitch.tv/v1/badges/channels/' + tmi().channel._id + '/display', function(badges) {
-                if (!badges || !badges.badge_sets || !badges.badge_sets.subscriber) return;
-                Object.keys(badges.badge_sets.subscriber.versions).forEach(function(version) {
-                    var subBadge = badges.badge_sets.subscriber.versions[version];
-                    var cssLine = '.badges .badge.twitch-subscriber-' + version + ' { cursor: pointer;';
-                    cssLine += 'background-image: url("' + subBadge.image_url_1x + '"); }';
-                    $style.append(cssLine);
-                });
-            });
-        }
-
         $style.appendTo('head');
     });
 };
+
+
+exports.loadSubBadges = function() {
+    if (!tmi().channel || tmi().channel.partner !== true) return;
+    $('#twitch_badges').remove();
+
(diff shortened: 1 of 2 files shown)


src/js/chat/helpers.js
```diff
@@ -710,22 +710,33 @@ exports.loadTwitchBadges = function() {
...
-                });
-            });
-        }
+        $style.appendTo('head');
+    });
+};
+
 
+exports.loadSubBadges = function() {
+    if (!tmi().channel || tmi().channel.partner !== true) return;
```
night: can you assign `tmi()` to a variable instead of calling this function repeatedly? not sure how this was never noticed prior

src/js/chat/helpers.js
```diff
@@ -710,22 +710,33 @@ exports.loadTwitchBadges = function() {
...
+            var subBadge = badges.badge_sets.subscriber.versions[version];
+            var cssLine = '.badges .badge.twitch-subscriber-' + version + ' { cursor: pointer;';
+            cssLine += 'background-image: url("' + subBadge.image_url_1x + '"); }';
+            $style.append(cssLine);
+        });
         $style.appendTo('head');
     });
 };
 
+
```
night: extra new line?

src/js/chat/helpers.js
EhsanKia: Honestly the usage of `tmi()` in general in that whole file really needs to be cleaned up. Some functions call it 20+ times...

src/js/chat/helpers.js
EhsanKia: The function above had 2 new lines which threw me off, fixed it to 1 new line between each declaration.

src/js/chat/helpers.js
```diff
@@ -714,14 +713,14 @@ exports.loadTwitchBadges = function() {
...
-    if (!tmi().channel || tmi().channel.partner !== true) return;
+    var tmiChannel = tmi().channel;
+    if (!tmiChannel || tmiChannel.partner !== true) return;
     $('#subscriber_badges').remove();
 
-    $.getJSON('https://badges.twitch.tv/v1/badges/channels/' + tmi().channel._id + '/display', function(badges) {
+    $.getJSON('https://badges.twitch.tv/v1/badges/channels/' + tmiChannel._id + '/display', function(badges) {
         if (!badges || !badges.badge_sets || !badges.badge_sets.subscriber) {
-            debug.log('Failed to load Subscribers badges for ' + tmi().channel.id);
+            debug.log('Failed to load Subscribers badges for ' + tmiChannel.id);
```
night: `.id` and `._id` are used?

EhsanKia: Yeah, `_id` is the integer userid of the account, `.id` is the string user. Though I could use `.name` for logging which is the display_name of the channel... ¯\\\_(ツ)_/¯

A++ naming convention by Twitch

PR #1671, opened 2016-10-19
Regular Expression Highlighting

rue92:
Hey there,

Same old deal: PR to add regular expression highlighting for #267 after my last PR was closed (it's cool, I was lazy about getting it done :P). Anyway, I added the safe-regex package as suggested to deal with potentially browser-locking regexes. Curiously, I couldn't get Chrome or Firefox to lock up on an unsafe-regex like `(a+){10}` without it anyway, but, there are a multitude of bad regex's I agree should be avoided.

Hope this meets your expectations!

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "lodash.throttle": "^4.1.1",
     "request": "2.75.0",
     "resizable": "^1.2.0",
+    "safe-regex": "^1.1.0",
     "twemoji": "^2.2.0",
     "twitch-chat-emotes": "github:night/Userscript--Twitch-Chat-Emotes#1fcd73bdfe491235506cecf3f32f56c2e06a31ff",
     "uglify-save-license": "^0.4.1",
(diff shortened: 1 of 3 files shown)


thinkpad4:
Now when this is disabled highlighting will function as it always has, right Rue92?

rue92:
Yup, that's the intention. Unless I've screwed up something which I don't
believe I did. :)

thinkpad4:
You rock Rue92

night:
I will review this over the weekend. Sorry for the delay.

night:
so an edge case here is when people place links in their highlight list. this breaks for those users.

rue92:
What do you think about backslashes or pipes then? Alternatively maybe we tack on square brackets, e.g. [/regex/]?

EhsanKia:
I think we should allow only a single regex expression, as in, starts and ends with /
I think the current implementation allows for any number of regexes inside a single highlight expression?

Also, once that's in, I think we can even remove the option from settings menu.
I don't see anything inputting a highlight expression starting and ending with /

night:
This PR is kinda dead again. Please address the issue with it and resubmit it again, I guess.

PR #1675, opened 2016-10-19
Add highlighting to the JS code block in README

tnws:
Using GFM, I simply added a little bit of highlighting to a code block in the README.


night:
I don't think we need syntax highlighting for a one line console snippet.

Issue #1777, opened 2016-11-20
Mod card views/follows data vertically misaligned with icons


AlcaDesign:
![image](https://cloud.githubusercontent.com/assets/7132646/20465900/db2fc7ae-af2d-11e6-8886-7ab93238440c.png)

I've had this issue for a while and I don't think anyone else has complained. Could be just me.

Issue #1772, opened 2016-11-19
Theatre mode and fullscreen toggle


Jonty800:
Hi,

When I'm in theatre mode and enter full screen (double click or the button) the fullscreen view is not correct. I'm using a MacBook Pro and I cannot replicate this if I disable BetterTTV.

Note the large black bar on the right hand side - the same width as the chat element

![Image of Bug](http://i.imgur.com/9XQkpny.jpg)

EhsanKia:
Can confirm the bug is back again...

HKS218:
Same here, only way to bring back Theatre/full screen is to disable BetterTTV :(

EhsanKia:
You're still having this issue? Also I don't get what you mean, the only time where this is an issue is if you are in theatre and then enter fullscreen. The simple solution is to leave theatre mode BEFORE entering fullscreen and it shouldn't be an issue.

Jonty800:
@EhsanKia Yes, this is a bug which is only caused when entering fullscreen mode while already in theatre mode.

There is another simple solution: fixing the bug.

EhsanKia:
I have been chasing the bug, but it's hard to fix a bug I can't reproduce.
Actually, I may have an idea. How do you enter fullscreen?
I think I have an idea that may fix this.

Jonty800:
Wow I didn't realise - this happens only when double clicking the video to enter fullscreen. When clicking the fullscreen icon, it's fine

EhsanKia:
Ok, yeah, this explains why I couldn't reproduce it. Alright should be able to fix this tonight.

Torixz:
A little after your post @EhsanKia 
I started having issues with the same, just that now i cant go to fullscreen mode when clicking the button. it works fine when bttv is disabled. But here is the strange part, now i can only go fullscreen with the stream when in theater mode and double clicking the video window.

HKS218:
My pc on windows 10 has black bars issue going from theatre mod to fullscreen (only by double clicking the video, fullscreen button is working)

On my second pc on Windows 7 i get the opposite, theatre mod is working both way, but double clic or full screen button on no-theatre mod give me this issue : https://github.com/night/BetterTTV/issues/1810

EhsanKia:
@Torixz @HKS218 yes, this is a different issue that happened after the miniplayer update yesterday, and we'll use #1810 to to track it.

night:
Should be fixed now.

Issue #1771, opened 2016-11-19
Twitch scrolls down when exiting fullscreen if channel header is disabled


IndubitablyBengt:
When exiting fullscreen mode with the new channel headers, Twitch scrolls down:
https://streamable.com/8lhb
When this is turned off, you can see that when exiting fullscreen, Twitch is programed to scroll down to hide the banner:
https://streamable.com/o0cb
As you can see, for a moment it is at the true top of the page before it scrolls itself back to hide the banner.
So, when the banner is disabled, this autoscroll is unnecessary and simply scrolls the player up out of view.

night:
🤔 that's kinda lame. @EhsanKia want to take a look?

EhsanKia:
I can replicate it. It's a bit annoying, though not the end of the world. I'll see if there's a clean fix...

PR #1780, opened 2016-11-21
Fix icon alignment and fullscreen issues

EhsanKia:
This should fix the icon alignment in #1777 

It'll also force the window to scroll up when you leave fullscreen fixing #1771 

I think I also found the issue with #1772 , it seems like a jquery bug with .data(), it caches the results and doesn't update when they change...


EhsanKia:
Also fix cheer issue (They changed their bits CDN to amazon).
Fixes #1783

PR #1782, opened 2016-11-21
Add change chat font family and size feature

HON95:
Added "Set font" and "Set font size" to chat options. If default font family and size is used, no styles will be added or changed.

If accepted, it should solve [Font Settings #57](https://github.com/night/BetterTTV/issues/57).

### DIFF
--- a/src/css/betterttv.css
+++ b/src/css/betterttv.css
@@ -980,10 +980,9 @@ input, textarea, select {
 
 #chat_line_list .line p .chat_line, .ember-chat .chat-messages .chat-line {
     word-break: break-word;
-    padding-top: 5px;
+    padding: 0.4em 0.8em;
     margin-top: inherit;
-    padding-left: 10px;
-    padding-right: 10px;
+    line-height: 1.3em;
 }
 
 #chat_line_list li .nick, .ember-chat .chat-messages .from {
--- a/src/css/betterttv.css
+++ b/src/css/betterttv.css
@@ -980,8 +980,10 @@ input, textarea, select {
 
 #chat_line_list .line p .chat_line, .ember-chat .chat-messages .chat-line {
     word-break: break-word;
-    padding: .4em .8em;
+    padding-top: 5px;
     margin-top: inherit;
+    padding-left: 10px;
+    padding-right: 10px;
 }
 
 #chat_line_list li .nick, .ember-chat .chat-messages .from {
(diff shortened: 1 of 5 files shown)


HON95:
It now doesn't set font family if just font size is set and doesn't set font size and padding if just font family is set.

HON95:
Now chat line padding scales implicitly with font size. The calculated padding at font size 13.33333px (BTTV default) is aproximately the same as the original BTTV padding.

src/css/betterttv.css
```diff
@@ -980,10 +980,8 @@ input, textarea, select {
 
 #chat_line_list .line p .chat_line, .ember-chat .chat-messages .chat-line {
     word-break: break-word;
-    padding-top: 5px;
+    padding: .4em .8em;
```
AlcaDesign: Should use "`0.#em`" instead of "`.#em`" to match everything else.

night:
We do not use ems, we use pixels. We've pushed changes containing ems in the past, and they always result in obscure UI bugs.

HON95:
Okay, now it waits until betterttv.css is loaded, then saves the calculated style values for a chat line as defaults. It also scales line height, so that it's readable at 50px font size. I also removed the !important suffixes from my CSS rules.

HON95:
I tried scaling icons (badges and emotes), but that just looked horrible.

EhsanKia:
Are you sure night? em's can be pretty tricky, but when used subtly and carefully it can be very useful, especially in this case. The only other option is scaling things with js which seems a bit dirty.

night:
If there's adequate testing on different DPIs and computers, maybe we can use ems here. All I know is I accepted a PR in the past with them, and people on windows started complaining instantly upon release

HON95:
I'll go back to ems and test that, then.

Is #chat_line_list used for anything? Searched for it in a loaded page, found nothing. Searched for it [on the this repo](https://github.com/night/BetterTTV/search?q=chat_line_list), found only selectors in CSS sheets.

EhsanKia:
If you need help, I've got a high dpi device and a windows device I can test on.

HON95:
I've tested it at font size default and 40px on Chrome, Firefox and Opera on Windows 10 at resolutions 0x0 to 1920x1080 and 100% and 175% of default Windows DPI for 1920x1080. No visual bugs found. I'll try on a Linux VM soon. Anything else I should try?

EhsanKia, it would be great if you could test it on your high DPI device.

HON95:
Firefox on Manjaro Linux at resolutions up to 1920x1080 tested, everything OK. I didn't test different DPIs.

PR #1804, opened 2016-12-01
Remove margin from pinned cheers

AlcaDesign:
image

Mentions under the recent cheer become visible around the edges. Removing the margin set by Twitch in the `.sticky-message--pinned-cheers` rule will resolve this issue.


night:
Did you test this on light and dark theme? Did you test this with multiline bits messages?

Issue #963, opened 2016-02-18
Keyboard shortcuts - Not an issue, merely a request


lrsen:
I propose to make some keyboard shortcuts to the player if at all possible. Like fullscreen, and play/stop. This can also work in VODs.

night:
Merging with #834

PR #1831, opened 2016-12-05
Player Keyboard Shortcuts

### LINKED ISSUE #834: Keyboard Shortcuts
Hello!

I would like to be able to mute the stream when I press the M button or pause the stream when I pause the P button or something.

If you're reading the comments on YouTube and you press "K" while the video is playing, it will be paused! When you're watching a video and press the M button, it will be muted. I'd like to see this with BetterTwitchTV and I think it wouldn't be too hard to implement!

---

Mute/Volume/Pause/Play/Fullscreen Keyboard Shortcuts

joelhoisko:
Resolves #834 (and #963)
Adds simple shortcuts in the style of Youtube to Twitch, namely play/pause(k), fullscreen(f) and mute(m). Respects all the inputfields/textareas on the screen(I think? `$('input, textarea')`) meaning if the user is focused on the chat text input it won't fire.

Also it's super "hacky" as it relies on the Twitch -players different buttons to work and the buttons classnames, namely `js-control-fullscreen/playpause/volume` to access the elements. I'm pretty sure there is a better way, but I'm not familiar enough with all this stuff to find the right way.

I don't think it conflicts with the Moderator Keybindings, as the keys are different, but I haven't tested it.

First time really playing around with this kind of jQuery etc stuff, so feel free to point me in the right direction with all this stuff.


EhsanKia:
I haven't looked at the code in detail but quick overview, here's a few tips that could improve this:
- Twitch uses [Mousetrap](https://craig.is/killing/mice) for their own hotkeys (alt+t, alt+r, alt+l, etc). Maybe you could just use that? Looking at the documentation, it automatically handles input fields too.
- For the player, here's another possible hack, I'll let you choose if it's cleaner or not.
```javascript
var playerServie = App.__container__.lookup('service:player');
var player = playerService.playerComponent.player;
player.muted = true;
player.pause();
player.play();
player.fullscreen = true;
```
- For the setting, I think you can just use 
```javscript
toggle: playerKeyboardShortcuts,
load: playerKeyboardShortcuts
```
And not actually insert anything in `main.js`. Then inside the handler, try getting the player object. If it exists, to the action, if not, then assume there's no active player and ignore the hotkey.

Hope this helps.

joelhoisko:
Thank you Ehsan for the tips! Implemented most of them!

I actually tried using Mousetrap earlier, but nonetheless I tried using it now again and it didn't correctly handle the chat `textarea` for some reason so I decided to stick with the jQuery `on()` method.

I changed the `document.getElementsByClassName` to the `App.__container__` method and it seems to work nicely and it feels a bit more "robust" as it isn't tied to the classnames etc. A quick Google search did turn up  some [complaints](http://stackoverflow.com/a/17839306) about using it, but also people complaining that there isn't a [better way](https://hiddentao.com/archives/2013/08/27/ember-js-is-stupidly-easy-stupidly-difficul/) or something (btw I have no experience about Ember, so feel free to dismiss everything I said).

I changed the settings too, didn't realise it had a  `load:` attribute, and removed references to the function from `main.js`

One problem which I tried to tackle, is that it doesn't work on the frontpage's player as `App.__container__.lookup('service:player').playerComponent` can't be found/undefined etc. I tried to fix this with using the old `document.getElementsByClassName` method, but for some weird reason (maybe hovering? `display: none` or something?) I couldn't get it to work from BetterTTV extension, only through manually inputting ``document.getElementsByClassName('js-control-fullscreen/playpause/volume').click()`` to the Javascript console.

Again, thank you for the tips!

EhsanKia:
The frontpage player is inside an iframe. And to make it worse it's from a different subdomain (player.twitch.tv) so there's CORS issues. I don't think there's any way around it? Maybe @night might know more.

thinkpad4:
So was Alt chosen so that if the mouse is over the player and you type k, f or m it won't mess with the player when you don't want it to?

joelhoisko:
@thinkpad4 if you press k, f or m anytime when you aren't focused on an input (like chat textbox, twitch search etc) it pauses/plays/mutes/fullscreens the player. There are no extra modifying keys involved, simply pressing k/f/m anywhere on the screen activates it.

src/js/features/player-keyboard-shortcuts.js
```diff
@@ -0,0 +1,47 @@
...
+                } else if (keyPressed.keyCode === keyCodes.f) {
+                    // 'f', toggle fullscreen
+                    player.fullscreen = !player.fullscreen;
+                } else if (keyPressed.keyCode === keyCodes.m) {
+                    // 'm', toggle mute
+                    player.muted = !player.muted;
+                }
+            }
+        } catch (err) {
+            // catch the typeError if the playerComponent.player doesn't exist and then just silently fail
```
night: can we log an debug line here if this happens

joelhoisko:
I changed the code, but I found out that the `App.__container__.lookup('service:player')` returns `undefined` so I went back to the `document.getElementBy` way. Maybe something changed on Twitch? This also means that I dropped the check for frontpage.

EhsanKia:
Yeah Twitch keeps renaming things. If you're still interesting in using it, it's `App.__container__.lookup('service:persistent-player');`

If not it's fine as long as it works I guess.

src/js/features/player-keyboard-shortcuts.js
```diff
@@ -0,0 +1,21 @@
+var keyCodes = require('../keycodes');
+
+module.exports = function() {
+    function handleKeyEvent(keyup) {
+        if (!$('input, textarea, select').is(':focus')) {
+            if (keyup.keyCode === keyCodes.k) {
+                document.getElementsByClassName('js-control-playpause-button')[0].click();
```
night: small nit, I'd prefer you to grab the player by id before traversing its children for the classes of the buttons.. Twitch uses `player` for the player's id..

`$('#player').find('.js-control-playpause-button').click()`

etc..

EhsanKia:
hey @joelhoisko, I think if you fix that small issue, I looked through this and it looks good to go.
Especially since it's behind a setting that off by default, I think we can push this.

joelhoisko:
Hi!
Sorry, got a little busy with life. 
I impemented the small fixes, more elegant this way. I'm just a noob with jQuery...
I tried to also tackle the issue with the frontpage player, but couldn't get it to work properly. The frontpage players buttons only seem to react to `$('#video-playback').find('.js-control-volume').click();` after some hovering/clicking so not sure what's going on in there.

EhsanKia:
The frontpage issue is pretty minimal in my opinion. I don't think majority of people ever use the front page, even less so actually watch a stream there. So I wouldn't say support for that is really necessary.

thinkpad4:
Can't wait for this to be pushed live, been wanting to check it out

d87:
F shortcut conflicts with standard in all browsers Cltr+F keybind for search. Same thing with Ctrl+M actually but it's not bothersome. Need to add check that modifiers are not pressed?

EhsanKia:
yeah, adding `!keyup.ctrlKey` should fix that.

PR #1843, opened 2016-12-08
Fixes #1842

### LINKED ISSUE #1842: Emotes bug
After I subscribed to a channel, betterTTV emotes are not showing up. For example, people in the chat typed "LUL" but I didn't see the emotes that it used to be. I should mention that betterTTV worked well for me before

OurFlagIsMined:


EhsanKia:
Can you explain how this fixes the bug with emotes?
I'm not at home so I can't test it, but I can't see how that makes sense.
Do you have steps for replicating this bug, as I don't quite understand the issue.

EhsanKia:
Huh, well that line still works on my end, but your method is cleaner so 👍 
LGTM if you wanna push this @night

EhsanKia:
For what it's worth, this solution was better because it actually works in every scenario, whereas the error catch you put basically ignores the error.

night:
my assumption is the playercomponent is not mounted.. which means there's nothing to resize at the moment the error happens

OurFlagIsMined:
I just did as much testing as I could, in both Chrome and Firefox; the try/catch catches every time. I wasn't able to produce any scenario where `playerService.playerComponent` (from `playerService.playerComponent.ownerView.rerender()`) wasn't `null`, so Twitch may have changed something on their end.

`playerService.set` is how Twitch moved the player in https://web-cdn.ttvnw.net/emberapp-cde0747e8f32397d92ae01732fe80374.js. I could have simplified it to
`playerService.set('fullSizePlayerLocation', playerService.fullSizePlayerLocation);`
but it doesn't seem to matter anyway, since even with `.set` the values don't appear to make the player move any more (I'm not sure if a 2nd change was made today that stopped it from working, or if I was using it in the wrong tab; since my Chrome tabs crashed).

Interestingly, if the `playerService.playerComponent.ownerView.rerender();` line is removed entirely (and not replaced with a `.set` line), the `Disable Channel Header` option still works 100% in both Chrome and Firefox.

EhsanKia:
Yeah this is really confusing. Things have definitely been changing since 2 days ago when I made this patch, this morning when I was testing, and now...

Right now, as you say, rerender doesn't do anything anymore, yet somehow BTTV still works, I'm guessing we are changing the value before Twitch renders the player? But that's scary because it could run into race conditions.

The only thing that still works for me consistently in the console is `.set('fullSizePlayerLocation', {top: x, left: y})`. Everything else has stopped working. I think we should switch to this to avoid some weird race condition...

OurFlagIsMined:
Yep, `.set('fullSizePlayerLocation', {top: x, left: y})` is working again now for me too. Not sure why it worked yesterday morning then stopped working in the afternoon; I think it may have been working the whole time and just been an issue with the way I was loading `betterttv.js` later in the day.

PR #1849, opened 2016-12-09
Swap chat

EhsanKia:
This is **NOT** ready to be pulled yet.

The swap works, the chat works, but the hover popups on the other side bar does not work yet.


EhsanKia:
Ok this is ready now. It also includes a commit to fix search-panel dark css.

PR #1872, opened 2016-12-13
darken new dashboard

teak:


EhsanKia:
hey teak can you also add darkening for the automod options here https://www.twitch.tv/settings/channel

PR #1893, opened 2016-12-18
Calculate player position automatically

OurFlagIsMined:
**TL;DR calculate player-position (top+left) automatically (using Twitch's own method), and use `setPlayerHeight` if you want to offset it from that position (otherwise, leave it at 0 for automatic positioning). Or if you don't care to be able to set player position (since Twitch's method sets it to the correct position regardless of what the header height is set to), scrap everything else in the function and just use `playerService.playerComponent.ownerView.rerender()` (just update `service:persistent-player` to `service:persistentPlayer`, since it appears to have changed to that).**

Twitch keeps changing things (one reason for the commit).

The other reason, is that Twitch's method for calculating player position seems to work regardless of what the header height is (tested in as many different scenarios as I could think up).

`setPlayerHeight` was replaced with the same method that Twitch uses; and its input is now a modifier to the player height, as opposed to the height itself. Which means `setPlayerHeight` can now be called with `0` and the correct position will automatically be calculated; unless future situations arise where it may be necessary to offset it from it's default position.

The player-height error (player loads higher than it should be, then auto-corrects the next time something causes Twitch's `updatePlayerPosition` to be called) appears to be something on Twitch's end, since it can happen even if I don't load BTTV. It could be fixed by calling `setPlayerHeight(0)` after the player is loaded (or Twitch's `updatePlayerPosition()`), but if it's not being caused by BTTV, then it might be wise to just wait for Twitch to fix whatever's causing it (after which, everything should work properly again; including the changes made in this commit).

Side node:
`service:persistent-player` now appears to be `service:persistentPlayer`. Also `playerService.playerComponent.ownerView.rerender()` seems to render it to the correct position, even if `fullSizePlayerLocation` is set beforehand. In which case, my player-calculations aren't really needed at all, they could just be replaced with `playerService.playerComponent.ownerView.rerender()` if there isn't a need to ever offset the default player height in the future (and could be smart to do, or at least be a quick replacement, if Twitch changes player-position calculations again).

And:
And manually setting player-heights will become a headache; as long as nothing fundamental changes about the page layout, the auto-height-calculations should prevent it from ever becoming an issue.


OurFlagIsMined:
Things are pretty messy lately with all of Twitch's changes. I put a lot of testing and research into this and it seems pretty fundamentally sound. But use your best judgement; use it, modify it, or throw it away completely. Or use the research to make your own revision.

night:
@EhsanKia can you review this?

EhsanKia:
Will do, a bit busy this week but will test it out.

night:
@EhsanKia ?

EhsanKia:
Eek, you can close this. I included a fix (Strongly inspired by this one) in my last PR (#1941)

PR #1970, opened 2017-01-13
Pinned cheer fixes

xgerhard:
- Makes the clickable size of the trashcan smaller, also moved it to the top-left since there is limited space on the right side if a cheer is expandable. Still not sure what the best placement is (http://imgur.com/a/j9JXr).

- Twitch sets a top: Xpx inline now, for pinned cheers, so we have to set .chat-messages to 0 when dismissing and when disabling bits. Since it's inline I guess we have to !important that.

### DIFF
--- a/src/css/betterttv-hide-bits.css
+++ b/src/css/betterttv-hide-bits.css
@@ -2,8 +2,8 @@
 	display: none;
 }
 
-.ember-chat .chat-messages .bits-tags__offset, .ember-chat .chat-messages {
-    top: 0 !important;
+.ember-chat .chat-messages.bits-tags__offset {
+    top: 0;
 }
 
 .ember-chat .sticky-message--pinned-cheers {
--- a/src/css/betterttv-hide-bits.css
+++ b/src/css/betterttv-hide-bits.css
@@ -3,7 +3,7 @@
 }
 
 .ember-chat .chat-messages .bits-tags__offset, .ember-chat .chat-messages {
-    top: 0;
+    top: 0 !important;
 }
 
 .ember-chat .sticky-message--pinned-cheers {
--- a/src/css/betterttv-hide-bits.css
+++ b/src/css/betterttv-hide-bits.css
@@ -2,7 +2,7 @@
 	display: none;
 }
 
-.ember-chat .chat-messages.bits-tags__offset {
+.ember-chat .chat-messages .bits-tags__offset, .ember-chat .chat-messages {
     top: 0;
 }
 
--- a/src/js/chat/helpers.js
+++ b/src/js/chat/helpers.js
@@ -117,7 +117,7 @@ exports.getCheerConfig = function(piece) {
 exports.dismissPinnedCheer = function() {
     try {
         var service = window.App.__container__.lookup('service:bits-pinned-cheers');
-        if (service.currentPinnedCheer) service.dismissLocalMessage();
+        if (service.topPinnedCheer || service.recentPinnedCheer) service.dismissLocalMessage();
     } catch (dismissError) {
         debug.log('Failed to dismiss cheer:', dismissError);
     }
--- a/src/js/chat/helpers.js
+++ b/src/js/chat/helpers.js
@@ -115,12 +115,11 @@ exports.getCheerConfig = function(piece) {
 };
 
 exports.dismissPinnedCheer = function() {
-    if (!window.App) return;
     try {
         var service = window.App.__container__.lookup('service:bits-pinned-cheers');
-        if (service.currentPinnedCheer && service.currentPinnedCheer !== null) service.dismissLocalMessage();
+        if (service.currentPinnedCheer) service.dismissLocalMessage();
     } catch (dismissError) {
-        debug.log('Failed to dismiss cheer: ' + dismissError);
+        debug.log('Failed to dismiss cheer:', dismissError);
     }
 };
 
--- a/src/js/chat/helpers.js
+++ b/src/js/chat/helpers.js
@@ -114,6 +114,16 @@ exports.getCheerConfig = function(piece) {
     return config;
 };
 
+exports.dismissPinnedCheer = function() {
+    if (!window.App) return;
+    try {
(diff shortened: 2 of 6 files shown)


src/js/chat/takeover.js
```diff
@@ -300,7 +300,10 @@ var takeover = module.exports = function() {
     // Dismiss pinned cheers
     $('body').off('click', '.pinned-cheers').on('click', '.pinned-cheers', function(e) {
         if (!e.target.classList.contains('pinned-cheers')) return;
-        if (e.target.offsetWidth - e.offsetX < 48 && e.target.offsetHeight - e.offsetY > 48) $('.pinned-cheers').hide();
+        if (e.target.offsetWidth - e.offsetX < 48 && e.target.offsetHeight - e.offsetY > 48) {
+            $('.pinned-cheers').hide();
+            $('.ember-chat .chat-messages').css('top', '0');
```
night: doesn't this break new pins being posted after?

src/css/betterttv-hide-bits.css
```diff
@@ -2,8 +2,8 @@
 	display: none;
 }
 
-.ember-chat .chat-messages.bits-tags__offset {
-    top: 0;
+.ember-chat .chat-messages .bits-tags__offset, .ember-chat .chat-messages {
+    top: 0 !important;
```
night: doesn't this break pins being posted?

xgerhard:
@night New pins = new pinned cheers rights? So the Top: 0px; doesn't break anything, it removes the reserved space behind a cheer. On page load Twitch reserves that space for the cheer, so the chat doesn't start behind it. When we dismiss it in BTTV, we can continue the chat from the top.

However, after you commented, I tested how it would look when a new cheer is posted and noticed when you dismiss a cheer, the whole pinned-cheers div is hidden, but it's never showing again on a new cheer. Not sure if there is a better place to detect a cheer, but I added a line to make the .pinned-cheer div show again in the [parseBits ](https://github.com/night/BetterTTV/blob/5292613e78f723214f2b71dcb0368257df27fdda/src/js/chat/templates.js#L75) function.

night:
This is just a messy solution, and isn't one I'm fond of. I'd rather there be a hook into Twitch's JS for removing the pinned message there than fighting with CSS on this. At least then it can be ran in a try/catch and worst case is it breaks and wont dismiss.

xgerhard:
Understandable. I'm fairly new to working with those hooks, so I looked at some other parts of the BTTV code and ended up with this. Is this the proper way to hook into Twitch JS, this looks like a way cleaner solution.
 
```javascript
// Dismiss pinned cheers
    $('body').off('click', '.pinned-cheers').on('click', '.pinned-cheers', function(e) {
        if (!e.target.classList.contains('pinned-cheers')) return;
        if (e.target.offsetWidth - e.offsetX < 48 && e.target.offsetHeight - e.offsetY > 48) {
            if (!window.App) return;
            try {
                window.App.__container__.lookup('service:bits-pinned-cheers').dismissLocalMessage();
            } catch (unpinError) {
                debug.log('Failed to unpin message: ' + unpinError);
            }
        }
    });
```

EhsanKia:
Looks okay to me, if you can confirm that it does indeed work, even when you're not a mod and they don't let you dismiss it

xgerhard:
Ok thanks, yeah the dismissLocalMessage function only changes the local currentPinnedCheer var to null without any conditions, I'll test this a bit more later today.

xgerhard:
Moved the dismiss to a function, so it can be called while disabling bits and dismissing a pinned cheers.
The function changes the local value of currentPinnedCheer inside the bits service to null. There are no extra checks in the Twitch dismiss function, should work for everyone, tested on mod/non-mod, logged in/out.
New cheers show up after dismissing and after disabling - enabling the bits.

EhsanKia:
Looks great!

src/js/chat/helpers.js
```diff
@@ -114,6 +114,16 @@ exports.getCheerConfig = function(piece) {
     return config;
 };
 
+exports.dismissPinnedCheer = function() {
+    if (!window.App) return;
+    try {
+        var service = window.App.__container__.lookup('service:bits-pinned-cheers');
+        if (service.currentPinnedCheer && service.currentPinnedCheer !== null) service.dismissLocalMessage();
```
night: you're essentially null checking twice, only need to do it once

src/js/chat/helpers.js
```diff
@@ -114,6 +114,16 @@ exports.getCheerConfig = function(piece) {
...
+    } catch (dismissError) {
+        debug.log('Failed to dismiss cheer: ' + dismissError);
```
night: might want to not convert the error to string so it's an error object we can get a stack out of in the console..

`: ', dismissError` instead

src/js/chat/helpers.js
```diff
@@ -114,6 +114,16 @@ exports.getCheerConfig = function(piece) {
     return config;
 };
 
+exports.dismissPinnedCheer = function() {
+    if (!window.App) return;
```
night: since everything is within a try catch this check seems superfluous

xgerhard:
Appreciate the feedback, should be good now.

xgerhard:
Unfortunately the new pinned cheer update broke it already, I'll look into that.

night:
when you fix this it should be good to merge

xgerhard:
Some notes for this last commit.
In this new update you can have 2 pinned cheers now, Top + Recent. Dismissing works fine, new cheers show up after. A thing I haven't figured out yet, when you dismiss a 'Top cheer' and someone will send a new cheer no matter what amount, the 'Top cheer' value will be set again, so you have to dismiss it again.

Since the top bar is bit smaller I moved the expand arrow to the bottom corner so our Dismiss button is clickable. The transform css rule is because when Twitch replaces a Recent cheer with a Top cheer they will be switched with an animation, which ends with some inline styling. Without the transform set to none, the expand button will overlap the dismiss button ([Example](http://imgur.com/a/G4GbV)). 
There might be a nicer solution for this, but will need some more testing.

Issue #1952, opened 2017-01-08
Option to disable/ignore twitchnotify messages.


gapbl4:
It's quite spammy and ruins offline chat experience when events like GDQ are hosted with all their subs. 
Unfortunately its not possible to "/ignore twitchnotify" , I just get "There was a problem ignoring this user". Can't ignore with bttv blacklist either. 

![](http://i.imgur.com/IqQqCPo.png)

gapbl4:
There's an option in FrankerFaceZ, but it's not compatible with BetterTTV.

![](http://i.imgur.com/a8Wp2R5.png)

night:
You can add twitchnotify in the beta/v7 to your blacklist to filter it.

PR #2030, opened 2017-02-02
notifications are checked for blacklisted words

### LINKED ISSUE #2027: Expand blacklisted keywords for system messages
It would be great if it would be possible to use blacklisted keywords with things like hosts and subs.

Maybe that is already possible and i just did something wrong, but what i tried to accomplish is to blacklist a specific user trying to spam chat by hosting and unhosting.

Any help is appreciated.

Best regards
~ceriath

ceriathstudios:
This fixes #2027 and #1952 

Basically i just moved the blacklist check in front of the notification check, which returns immediately, which caused notifications to be never checked for blacklisted words.

I tested it with the keywords "hosting" and "subscribed" - both worked.
Also when i added a username as a normal term (without "()") his subs/hosts were not shown, others were.

Since its not a big deal i would really appreciate an approvement :)

### DIFF
--- a/src/js/chat/handlers.js
+++ b/src/js/chat/handlers.js
@@ -452,6 +452,13 @@ var privmsg = exports.privmsg = function(channel, data) {
 
     if (data.style && ['admin', 'action', 'notification', 'whisper'].indexOf(data.style) === -1) return;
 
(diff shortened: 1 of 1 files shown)


EhsanKia:
Looks good to me.

PR #2078, opened 2017-02-19
Theatre Mode shortcut

kallebornemark:
Extended `Player Keyboard Shortcuts` setting by adding shortcut ("*T*") for Theatre Mode.

### DIFF
--- a/src/js/features/player-keyboard-shortcuts.js
+++ b/src/js/features/player-keyboard-shortcuts.js
@@ -9,6 +9,8 @@ module.exports = function() {
                 $('#player').find('.js-control-fullscreen').click();
             } else if (keyup.keyCode === keyCodes.m) {
                 $('#player').find('.js-control-volume').click();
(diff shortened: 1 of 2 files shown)


EhsanKia:
Doesn't alt+t already do theater mode? That's part of Twitch itself.

night:
It does. We need not reimplement what's implemented already.

kallebornemark:
Urm, so what is the point of this setting? All of the keybinds it includes are already available at alt+key.

night:
They definitely are not.. The only other alt key for the player is alt+x for clipping

kallebornemark:
Oh, I see my confusion now. Having the setting enabled also allows one to use its keybinds with the alt modifier, making me believe alt+key existed for all of the features (mute, pause, etc). And even though I would much prefer "T" over "alt+T", I fully understand the choice of not implementing it. Sorry about the fuzz!

PR #2103, opened 2017-03-04
Couple small fixes for old BTTV

EhsanKia:


src/js/chat/handlers.js
```diff
@@ -422,12 +420,6 @@ var privmsg = exports.privmsg = function(channel, data) {
         store.displayNames[data.from] = data.tags['display-name'];
     }
 
-    // filter blacklisted emojis
-    blacklistedEmoji.forEach(function(emoji) {
```
night: why is this removed

voloshink: bttvMessageTokenize already handles the blacklist, no?

EhsanKia: What does this do? There's another blacklist emoji removal inside templates.js:

https://github.com/night/BetterTTV/blob/207cc6dd945eb01e73c47156d1a3035fc58240ca/src/js/chat/templates.js#L211

This catches chat, whispers and replay, so I'm not sure what other edge cases there are that this needs to catch?

night: k.. well then this may be fine.. idc really since v7 will handle this all better by filtering usable ones

PR #2106, opened 2017-03-05
Tab completion and other small fixes

EhsanKia:
Adds tab completion.

Features:
- Supports users and emotes
- works in chat and  conversations
- let's you tab complete in the middle of the line

Missing:
- does not support the suggestion popout yet
- does not support custom ordering (emotes first, reply to mention first, etc)
- does not support @ usernames yet

There's also a few other small fixes in the patch:
- Fix BVL callback
- add check to autotheatremode to avoid extra spam in console
- small regex typo


src/modules/auto_theater_mode/index.js
```diff
@@ -22,7 +22,9 @@ class AutoTheaterModeModule {
             const container = App.__container__;
             const routeName = container.lookup('controller:application').get('currentRouteName');
             if (routeName !== 'channel.index.index' && ['videos', 'vod'].includes(routeName)) return;
-            if (container.lookup('service:persistentPlayer').playerComponent.player.theatre) return;
+
+            const playerComponent = container.lookup('service:persistentPlayer').playerComponent;
```
night: i feel like this is the wrong fix.. why is channel loading on chat.. that should be prevented tbh

src/modules/tab_completion/index.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    load() {
+        this.tabTries = -1;
+        this.suggestions = null;
+        this.textSplit = ['', '', ''];
+        this.userList = new Set();
+
+        // Chat
+        $('body').off('click focus', CHAT_TEXT_AREA).on('click focus', CHAT_TEXT_AREA, this.onFocus);
```
night: every call to `$('body')` is expensive.. store it or chain these..

```
$('body').off().on().off().on()
```

src/modules/tab_completion/index.js
```diff
@@ -0,0 +1,109 @@
...
+        $('body').off('keydown', CHAT_TEXT_AREA).on('keydown', CHAT_TEXT_AREA, e => this.onKeydown(e));
```
night: it's safe to `off` this without the callback because we're using our own jquery, but we shouldn't do this as it means no other module can listen to his event on that textarea

src/modules/tab_completion/index.js
```diff
@@ -0,0 +1,109 @@
...
+        if (e.ctrlKey) return;
+
+        const $inputField = $(e.target);
+        const $suggestions = $(CHAT_SUGGESTIONS);
+        if ($suggestions.length) $suggestions.remove();
+
+        if (keyCode === keyCodes.Tab) {
+            e.preventDefault();
+
+            // First time presssing tab, split before and after the word
```
night: sss

src/modules/tab_completion/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+    resetChannelData() {
+        this.userList = new Set();
+    }
+
+    onFocus() {
+        this.tabTries = -1;
+    }
+
+    onKeydown(e, includeUsers) {
```
night: KeyDown

src/modules/tab_completion/index.js
```diff
@@ -0,0 +1,109 @@
...
+                if (this.tabTries >= this.suggestions.length) this.tabTries = 0;
+                if (this.tabTries < 0) this.tabTries = this.suggestions.length + this.tabTries;
+                $inputField.val(this.textSplit[0] + this.suggestions[this.tabTries] + this.textSplit[2]);
+
+                const cursorPos = this.textSplit[0].length + this.suggestions[this.tabTries].length;
+                $inputField[0].setSelectionRange(cursorPos, cursorPos);
+            }
+        } else if (keyCode === keyCodes.Esc && this.tabTries >= 0) {
+            $inputField.val(this.textSplit.join(''));
+        } else if (keyCode !== keyCodes.Shift) {  // not shift
```
night: nit: double space before comment

src/modules/tab_completion/index.js
```diff
@@ -0,0 +1,109 @@
...
+            this.tabTries = -1;
+        }
+    }
+
+    getSuggestions(prefix, includeUsers = true) {
+        const suggestions = emotes.getEmotes().map(emote => emote.code); // Emotes
+        if (includeUsers === true) suggestions.push(...this.userList);   // Users
```
night: nit spacing before comment

src/modules/auto_theater_mode/index.js
EhsanKia: Honestly, I had this happening when I was browsing around twitch. But you're right, should try to get to the root of the issue.

src/modules/tab_completion/index.js
```diff
@@ -0,0 +1,108 @@
+const $ = require('jquery');
+// const debug = require('../../utils/debug');
+const keyCodes = require('../../utils/keycodes');
+const emotes = require('../emotes');
+// const settings = require('../../settings');
```
night: why are these deps commented out?

EhsanKia: Fixed

PR #2123, opened 2017-03-14
Addition of setting to enable column toggling

ClayBenson94:
This toggle allows users to toggle the visibility of the left and right columns of various twitch.tv pages

When in a stream, the right column is where chat is displayed, so toggling that column hides and shows the chat. On other pages that have a right column, this is where your followed streamers are listed.

The left column is where most settings and navigation links are held (like Games, Channels, BTTV Settings, etc.)

The reason I made this little tweak was I found myself constantly wanting to be able to hide the chat when I moved a stream to a different sized monitor or resized the window. This would make chat take up a little too much space, and I would hide it for more stream real estate. I'm lazy and like to use my keyboard, so why the heck not.

### DIFF
--- a/src/js/features/column-toggle-hotkey.js
+++ b/src/js/features/column-toggle-hotkey.js
@@ -1,0 +1,17 @@
+var debug = require('../helpers/debug');
+
+module.exports = function() {
+    if (!bttv.settings.get('columnToggleHotkey')) return;
+    if (!window.Ember || !window.App) return;
+
+    try {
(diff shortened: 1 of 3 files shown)


night:
We are not accepting new PRs for the old version of BetterTTV. You should release a feature on the v7 branch only. Additionally, we don't use Twitch's Mousetrap for binding to keys. You should use jQuery directly.

ClayBenson94:
Thanks, will do :)

PR #2125, opened 2017-03-14
Added column-toggle-hotkey module

ClayBenson94:
This is a re-do of [this](https://github.com/night/BetterTTV/pull/2123) pull request I made yesterday, but now on the appropriate v7 branch - instead of master. Additionally, it uses JQuery to listen for keypress events instead of Twitch's Mousetrap.

Let me know what you think! I had some considerations to make this a generic "Add navigation hotkeys" toggle, as I don't notice there being a toggle to enable global twitch hotkeys for navigation (maybe in a future version this could just be a toggle that adds a slew of hotkeys?)

### DIFF
--- a/src/modules/column_toggle_hotkey/index.js
+++ b/src/modules/column_toggle_hotkey/index.js
@@ -20,6 +20,11 @@ class ColumnToggleHotkeyModule {
             $(document).keydown(e => {
                 if (e.ctrlKey) {
                     if (e.which === 37) {
+                        // Check if you're in theater mode, don't click left one if you are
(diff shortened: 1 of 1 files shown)


ClayBenson94:
It seems as if this change messes with the ability to "UpArrow" to restore previously sent chat messages. Is there a better way to approach these key listeners to avoid something like this?

night:
Apologies for not realizing the point of this PR sooner. Twitch already has hotkeys for this: `alt+r` and `alt+l`

ClayBenson94:
Whoops! I honestly had no idea! Thanks for checking it out anyways :)

PR #2146, opened 2017-03-20
Improved chat settings dialogs

dougrk:
These changes improve the behavior and appearance of the dialog boxes that appear when users click on BetterTTV's chat settings.

Prior to these changes, the settings dialog was the built-in `prompt()` method, which stops the script until the user completes the dialog. This is undesirable behavior because it prevents both the chat and the player from working.

I introduced a new utility that provides a more robust dialog box controlled by jquery. 
promptexample1
The dialog remains on screen until the user clicks either the save or cancel button. Clicking on another chat option will replace the current prompt. When the user clicks save, a callback method is called, updating the settings exactly the same way it was done prior to the change. 

Not only does it solve the issue of `prompt()` stopping the script, it is also customizable. I copied some styles from the main BetterTTV settings panel to maintain a consistent appearance with the rest of the app.

I hope that you find these changes satisfactory and accept this pull request. 

-- Doug


night:
this is a decent idea, but I'm not sure if we want this. the end goal is to make these popups their own settings panel in the settings when it is redesigned

dougrk:
I agree that a redesign is a good solution, but with changes Twitch has made recently, I'm guessing that will be low priority for a while. This would be a good way to bridge the gap.

night:
I mean, we already have a design and implementation, it's just not baked into v7 yet

dougrk:
Alright, thanks. I guess that wraps this up then. I don't have mush else to add. The prompt as-is is pretty bad for UX, so I hope you guys release the redesign soon.

PR #2169, opened 2017-03-22
Anon chat and small error fixes

EhsanKia:


src/modules/anon_chat/index.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        const currentUser = twitch.getCurrentUser();
+        if (!currentUser) return;
+
+        const session = twitch.getCurrentTMISession();
+        if (!session) return;
+
+        try {
+            const prodConn = session._connections.aws || session._connections.prod || session._connections.main;
```
night: iirc there's only main now

src/modules/anon_chat/index.js
```diff
@@ -0,0 +1,67 @@
+const settings = require('../../settings');
+const watcher = require('../../watcher');
+const twitch = require('../../utils/twitch');
+const Raven = require('raven-js');
+
+const forcedURL = window.location.search && window.location.search.indexOf('bttv_anon_chat=true') > -1;
```
night: search should default to empty string.. also use .includes instead of indexOf

src/modules/anon_chat/index.js
```diff
@@ -0,0 +1,67 @@
...
+                prodConn._send('QUIT');
+            }
+        } catch (e) {
+            twitch.sendChatAdminMessage('BetterTTV: [Anon Chat] We encountered an error anonymizing your chat. You won\'t be hidden in this channel.');
+            Raven.captureException(e);
+        }
+    }
+
+    onMessage($el, data) {
+        if (this.ignoreNextDC && data.message.indexOf('unable to connect to chat') > -1) {
```
night: includes

src/modules/anon_chat/index.js
```diff
@@ -0,0 +1,67 @@
...
+class AnonChatModule {
+    constructor() {
+        settings.add({
+            id: 'anonChat',
+            name: 'Anon Chat',
+            defaultValue: false,
+            description: 'Join channels without appearing in chat'
+        });
+
+        this.ignoreNextDC = false;
```
night: should be stored outside of the class

src/modules/anon_chat/index.js
```diff
@@ -0,0 +1,67 @@
...
+        watcher.on('load.chat', () => this.load());
+        watcher.on('chat.message', ($el, emberView) => this.onMessage($el, emberView));
+        settings.on('changed.anonChat', () => this.load());
+    }
+
+    load(force) {
```
night: i'd much prefer an actual api for this rather than use load to toggle.. should be a .join and .part .. not hard since you can now easily store state outside the class

EhsanKia: Hmm, how's this implementation?

src/modules/anon_chat/index.js
```diff
@@ -0,0 +1,66 @@
...
+            description: 'Join channels without appearing in chat'
+        });
+
+        watcher.on('load.chat', () => this.load());
+        watcher.on('chat.message', ($el, emberView) => this.onMessage($el, emberView));
+        settings.on('changed.anonChat', () => this.load());
+    }
+
+    part() {
+        const currentUser = twitch.getCurrentUser();
```
night: pull all this logic out to a helper function that takes a nickname.. have this function send the message and call that

EhsanKia: third try's the charm?

src/modules/anon_chat/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+        this.changeUser(currentUser.name, 'Logging you into chat...');
+        this.enabled = false;
+    }
+
+    load() {
+        if (forcedURL || settings.get('anonChat')) {
+            this.part();
+        } else {
+            this.join();
```
night: needs a way not to emit the message.. otherwise going into any chat will echo logging into chat

PR #2173, opened 2017-03-22
Fix errors in V7

EhsanKia:
- Don't try to run anon-chat for every single user connecting to chat
- Set eventlisteners max handlers to 100
- Check for undefined values in a few places

### DIFF
--- a/src/modules/anon_chat/index.js
+++ b/src/modules/anon_chat/index.js
@@ -16,7 +16,7 @@ class AnonChatModule {
 
         this.enabled = false;
         watcher.on('load.chat', () => this.load());
-        watcher.on('chat.message', ($el, emberView) => this.onMessage($el, emberView));
+        watcher.on('chat.message', ($el, msgObj) => this.onMessage($el, msgObj));
         settings.on('changed.anonChat', () => this.load(true));
     }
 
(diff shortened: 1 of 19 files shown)


src/modules/anon_chat/index.js
```diff
@@ -46,16 +46,17 @@ class AnonChatModule {
...
-        } else {
+        } else if (force) {
             this.join();
         }
     }
 
     onMessage($el, data) {
-        if (ignoreNextDC && data.message.includes('unable to connect to chat')) {
+        if (!ignoreNextDC) return;
+        if (data && data.message && data.message.includes('unable to connect to chat')) {
```
night: this event only emits if data exists?

src/modules/anon_chat/index.js
```diff
@@ -46,16 +46,17 @@ class AnonChatModule {
...
+    load(force = false) {
         if (forcedURL || settings.get('anonChat')) {
             this.part();
-        } else {
+        } else if (force) {
             this.join();
         }
     }
 
     onMessage($el, data) {
```
night: `{message}` instead

src/modules/anon_chat/index.js
```diff
@@ -17,15 +17,15 @@ class AnonChatModule {
         this.enabled = false;
         watcher.on('load.chat', () => this.load());
         watcher.on('chat.message', ($el, emberView) => this.onMessage($el, emberView));
```
night: just fyi this is not an emberView it's the msgObject

EhsanKia:
Fixed.

PR #2190, opened 2017-03-25
V7 send message API

EhsanKia:
defines a proper api for when a message is sent that modules can listen for and act on.

### DIFF
--- a/src/modules/chat_commands/index.js
+++ b/src/modules/chat_commands/index.js
@@ -1,5 +1,4 @@
 const $ = require('jquery');
-const watcher = require('../../watcher');
 const twitch = require('../../utils/twitch');
 const twitchAPI = require('../../utils/twitch-api');
 const chat = require('../chat');
@@ -231,11 +230,9 @@ function handleCommands(message) {
 }
 
 class ChatCommands {
-    constructor() {
-        watcher.on('chat.send_message', sendState => this.handleSendMessage(sendState));
-    }
+    constructor() {}
 
-    handleSendMessage(sendState) {
+    onSendMessage(sendState) {
         const result = handleCommands(sendState.message);
         if (result === false) {
             sendState.message = '';
--- a/src/modules/chat_commands/index.js
+++ b/src/modules/chat_commands/index.js
@@ -232,6 +232,19 @@ function handleCommands(message) {
 
 class ChatCommands {
     constructor() {
+        watcher.on('chat.send_message', sendState => this.handleSendMessage(sendState));
+    }
+
+    handleSendMessage(sendState) {
+        const result = handleCommands(sendState.message);
+        if (result === false) {
+            sendState.message = '';
+            sendState.preventDefault();
+        }
+
+        if (typeof result === 'string') {
+            sendState.message = result;
+        }
     }
 }
 
--- a/src/modules/chat_commands/index.js
+++ b/src/modules/chat_commands/index.js
@@ -235,7 +235,6 @@ class ChatCommands {
     onSendMessage(sendState) {
         const result = handleCommands(sendState.message);
         if (result === false) {
-            sendState.message = '';
             sendState.preventDefault();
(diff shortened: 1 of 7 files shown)


src/modules/emotes/emojis.js
```diff
@@ -82,6 +83,15 @@ class Emojis extends AbstractEmotes {
                 this.emotes.set(code, emote);
             });
     }
+
+    convertEmojis(sendState) {
+        sendState.message =  sendState.message.split(' ').map(piece => {
```
night: this is gonna create a chicken & the egg problem.. emojis run before something that replaces text that needs a replacement

src/modules/chat_commands/index.js
```diff
@@ -234,46 +230,21 @@ function handleCommands(message) {
...
-        if (!emberView) return;
-
-        const newTwitchSendMessage = emberView._actions.sendMessage;
-
-        // check if we've already monkeypatched
-        if (newTwitchSendMessage === sendMessage) return;
+    handleSendMessage(sendState) {
+        const result = handleCommands(sendState.message);
+        if (result === false) {
+            sendState.message = '';
```
night: this shouldn't be necessary.. to cancel sending a message why not a `sendState.cancel() which takes the clearing of the value in on its own

src/watcher.js
```diff
@@ -195,7 +198,46 @@ class Watcher extends SafeEventEmitter {
             })
         );
 
+        let twitchSendMessage;
+
+        const emitSendMessage = sendState => this.emit('chat.send_message', sendState);
```
night: I honestly do not think this is the right solution to this problem. We need a separate module that functions as an ordered mutator for sending messages.

src/watcher.js
```diff
@@ -195,7 +198,46 @@ class Watcher extends SafeEventEmitter {
...
+
+        function bttvSendMessage() {
+            try {
+                let defaultPrevented = false;
+                const sendState = {
```
night: this should be a message send class that gets instantiated on message send, don't ya think?

src/modules/emotes/emojis.js
night: also extra space

src/watcher.js
```diff
@@ -13,6 +14,8 @@ const chatState = {
     r9k: 0,
     subsOnly: 0
 };
+const TEXTAREA_SELECTOR = '.textarea-contain';
+
```
night: extra new line

src/modules/chat_commands/index.js
EhsanKia: What if you want to abort the message send but not clear the textbox? Or are you proposing a third function that does both?

night: what is the use case of an abort? i'm not opposed, just curious

EhsanKia: Let's say anon chat blocks you from sending a message, we don't want to eat the user's message maybe? I guess once we add message history it's less of an issue, Also, maybe we want to transform the content of textbox for the user, but not send it yet.

night: then i suggest having 2 is optimal

src/modules/tab_completion/index.js
```diff
@@ -109,6 +116,27 @@ class TabCompletionModule {
         } else if (keyCode !== keyCodes.Shift) {
             this.tabTries = -1;
         }
+
+        // Message history
+        if (keyCode === keyCodes.UpArrow) {
```
night: we need a way to disable this

EhsanKia: Do we? It's pretty unobtrusive. Unlike the old one, it only goes to the previous if you're at the start of your message, and it also saves the edits you made the messages, so when you go back, you don't lose any data. The chances of it accidentally fucking you over are basically non-existent.

Issue #2218, opened 2017-04-02
[v7] Feature reimplementation request:  Double-Click Auto-Complete


Nareese:
Not one I personally use, but I have seen several people asking for it in discord...  Double clicking on a user's name in chat copies the name to the chat input area.

EhsanKia:
Also coming, hopefully this weekend.

Issue #2217, opened 2017-04-02
[v7] Feature reimplementation request:  modcard keybinds


Nareese:
Requesting re-implementation of modcard keybinds for (p)urge, (t)imeout, (b)an, (u)nban, and (i)gnore, (a)!permit, and (w)hisper.  v6 implementation at js/chat/takeover.js:581

EhsanKia:
Definitely coming, hopefully this weekend.

Issue #2214, opened 2017-04-01
[v7] Insert name by double-click is missing


CaptainFlint:
The function of inserting user name into chat editbox by double-clicking the name is missing in version 7.

EhsanKia:
Coming soon (it's in the list of features that is missing and is being added).

Issue #2210, opened 2017-03-31
Enabling left side chat hides the 'hide chat' arrow in theater mode


OnlyMason:
In theater mode, the 'hide chat' arrow is either hidden or removed when using left side chat.

PR #2222, opened 2017-04-02
Doubleclick Mention, Mod keybinds and small fixes

EhsanKia:
- Add doubleclick mention for both usernames and chat mentions (#2214, #2218)
- Add mod keybinds (#2217)
- Fix theatre mode left-chat button (#2210)

- Fix a few module names to follow our standards
- Better z-index for BVL


src/modules/mod_keybinds/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+const CHAT_TEXT_AREA = '.ember-chat .chat-interface textarea';
+
+class ModKeybindsModule {
+    constructor() {
+        this.load();
+    }
+
+    load() {
+        $('body').on('keydown', e => this.onKeyDown(e));
```
night: I don't want these attached to body.. also, can we just make a separate file under the mod card for keybinds, and then import and run this when a modcard is created? extending its class for current card info seems out of place.

PR #2286, opened 2017-04-20
V7 - minor fixes

Aaron128l:
- Realign Badges & Moderator Icons
- Fixes Deleted Messages

### DIFF
--- a/src/css/betterttv.css
+++ b/src/css/betterttv.css
@@ -23,10 +23,6 @@ body.channel_new {
     border-radius: 0px;
 }
 
-.chat-line .message .deleted {
-    color: #6441a4;
-}
-
 #metagame a {
     color: white !important;
 }
@@ -2348,5 +2344,10 @@ body.channel_new.columns.ember-application {
 }
 
 .chat-line .bttv-deleted {
+    opacity: 0.75;
+}
+
+.chat-line .bttv-deleted.bttv-click {
     opacity: 0.5;
 }
+
--- a/src/css/betterttv.css
+++ b/src/css/betterttv.css
@@ -2350,11 +2350,8 @@ body.channel_new.columns.ember-application {
     cursor: pointer;
 }
 
-.chat-line .bttv-deleted {
-    opacity: 0.75;
-}
-
 .chat-line .bttv-deleted.bttv-click {
     opacity: 0.5;
+    text-decoration: none;
 }
 
--- a/src/css/betterttv.css
+++ b/src/css/betterttv.css
@@ -153,12 +153,16 @@ input.text:focus,select:focus,textarea:focus {
    text-decoration: none;
 }
 
-.ember-chat .badges {
-    font-size: 0;
+.ember-chat .badges span:last-child .badge {
+    margin-right: 3px;
 }
 
 .ember-chat .mod-icons {
     margin-right: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/modules/chat_deleted_messages/index.js
```diff
@@ -38,11 +38,10 @@ function onClearChat(name, tags) {
                 });
 
                 if (!settings.get('showDeletedMessages')) {
-                    $message.addClass('bttv-click');
                     const $messageClone = $message.clone();
+                    $message.addClass('bttv-click');
                     $message.click(() => {
                         $message.replaceWith($messageClone);
-                        $message.removeClass('bttv-click');
```
night: why is this removed?

src/css/betterttv.css
```diff
@@ -2347,6 +2350,8 @@ body.channel_new.columns.ember-application {
     cursor: pointer;
 }
 
-.chat-line .bttv-deleted {
+.chat-line .bttv-deleted.bttv-click {
     opacity: 0.5;
+    text-decoration: none;
```
night: i prefer the underline

src/css/betterttv.css
```diff
@@ -157,8 +153,16 @@ input.text:focus,select:focus,textarea:focus {
    text-decoration: none;
 }
 
-.ember-chat .badges, .ember-chat .mod-icons {
-    font-size: 0;
+.ember-chat .badges span:last-child .badge {
```
night: the last time you messed with badges you caused a lot of alignment issues when there were:

1) no mod icons
2) no badges
3) no timestamps
4) mix-match 1-3

src/css/betterttv.css
night: also this breaks opacity of deleted messages with show deleted messages on

src/modules/chat_deleted_messages/index.js
Aaron128l: I switch the order of the clone creation so the `bttv-click` class is added **after** the clone is created. No need to remove the class, since it wasn't there when the clone was created.

src/css/betterttv.css
Aaron128l: This only affects the opacity of messages with `bttv-click`. With 'Show Delete Messages' enabled the class isn't added, showing the normal message with no opacity.

night: `.bttv-deleted` is added when a message is deleted.. you're removing its class

src/css/betterttv.css
Aaron128l: All alignment issues stemmed from `font-size: 0` on both the mod icons and badges. With it being removed, the chat line goes back to its former alignment. All vertical margin tweaking is because of the compensation for my mistake. This change is mainly for the left and right additional whitespace added with `span` in an `inline-block` display. 

While I didn't want to resort to margins. `font-size: 0` was a mistake on multiple fronts.

src/css/betterttv.css
Aaron128l: `.bttv-deleted` is always on a deleted message, yes. I'm only applying the opacity to the `<message deleted>` text applied with the `.bttv-click` class. 

I believe for features like Show Delete Messages a deleted message shouldn't be opaque. It's open for personal preference.

PR #2441, opened 2017-05-12
Fixing colon under username in chat

SW1FT:
This is how it currently looks like:
colonbug


SW1FT:
Ops, just noticed someone else already made a PR regarding this.

Issue #2404, opened 2017-05-02
Image hover does not work


meinkrahe:
Image hovering does not seem to work anymore whenever I hover over links that feature images or gifs such as puush or waa.ai.

Nareese:
Further info on this.  I tested this against an imgur link (http://i.imgur.com/fNjkUSb.png) and each time I mouse into the link, I get two of this error in the console, presumably from the api call.

`Unhandled promise rejection Object { status=200, data=undefined}`

I didn't really know what an appropriate link to test the link_resolver api call, but presumably it's the same.

PS:  While you're in there fixing it, can you also re-add the option to turn it off (was chatImagePreview in v6).

PR #2452, opened 2017-05-15
restores chatImagePreview setting

Nareese:
Restores the setting to disable image previews per my PS request in #2404.


night:
why

Nareese:
I guess my initial answer would be, why not?  It was a feature in v6 and has virtually no overhead to implement.  Some people just prefer that images not pop up if they over over links.  It defaults to on so people have to explicitly disable it.  Also, it can prevent spammers who spam offensive images from having them pop up on preview unexpectedly.

Feel free to reject it, I guess, if you feel strongly about it, and I'll just maintain it in my local copy.

thinkpad4:
I wouldn't mind having this setting back

Issue #2397, opened 2017-05-01
Emote autocompletion priority option removed?


ojji:
I had this option on which allowed me to type 'Wu' and tab-complete to the WutFace twitch emote. Now instead I have to cycle through usernames until I find the right selection, which is extremely annoying to me especially in crowded chats where most likely there are users with names similar to Twitch emotes. Did I miss something or this option got removed completely?

EhsanKia:
It will be back soon.

Nokanah:
Mine worked fine until today, now have all sorts of issues on twitch.  Have to turn off html5 just to watch anyone and my bttv does absolutely nothing, when I go to the settings screen I have no options or settings to modify.

Latias4Ever:
I'm having the same issues as Nokanah, my settings window is completely empty, and I can't autocomplete emotes, it just says "No matches".

Screenshot of the settings window: http://prntscr.com/f9v603

Issue #2390, opened 2017-04-30
Minor @ user bug.


Apolloftw:
Whenever you @_@ or the same with a number or letter such as @5@ and @A@, the second @ disappears. http://imgur.com/a/yjyxE I tested this without BetterTTV enabled and it shows the second @.

Issue #2333, opened 2017-04-25
"Shift+Tab" should work like "Tab" like in previous versions!


neokjo:
With the latest version you can't auto-complete with Shift-Tab anymore.
In Chat Clients like Chatty this is used for emote-autocomplete.
Now with the latest update Shift+Tab just do nothing.

Would be a small usefull change to bring it back in the latest version

Edit: Subemotes seems to work with Shift+Tab atm. Only Twitch Global and BTTV Emotes not Complete with Shift+Tab

EhsanKia:
I'm not sure what you mean. The behavior of shift tab is same as before, and also consistent with all other software. Tab goes to the next suggestion, and shift+tab goes to the previous suggestion. It simply reverses the order.

neokjo:
If you have disabled "Completion Tooltip" - "Tab" and "Shift+Tab" both should complete the emote/nickname. but at this moment, this only work for subemotes,

EhsanKia:
I see the bug. There's a small bug where if the list only contains one element, shift+tab won't properly work. Otherwise, if there are more than 1, tab goes to the first, shift+tab goes to the last.

PR #2477, opened 2017-05-23
Fix anon chat, tab completion and other small stuff

### LINKED ISSUE #2402: Anon chat isn't working properly
I usually use anon chat when on Twitch, and it usually lets me go into a stream invisible, but the last few days it hasn't been working.  It shows me in the chat list name and I have to click on anon chat to get it to work again.

EhsanKia:
Closes #2402, #2397, #2390, #2333


src/modules/better_viewer_list/index.js
```diff
@@ -255,6 +254,10 @@ class BetterViewerListModule {
         // Load viewers
         this.loadViewerList();
     }
+
+    capitalize(str) {
```
night: this doesn't really need to be on the class.. i'd drop it as a named function at the top

EhsanKia: done.

src/modules/better_viewer_list/index.js
```diff
@@ -89,6 +89,7 @@ class BetterViewerListModule {
     extractViewers(data) {
         const typeDisplays = ['STAFF', 'ADMINS', 'GLOBAL MODERATORS', 'MODERATORS', 'VIEWERS'];
         const userTypes = ['staff', 'admins', 'global_mods', 'moderators', 'viewers'];
+        const capitalize = str => str.charAt(0).toUpperCase() + str.slice(1);
```
night: but why? :notlikethis: i meant before the class is even defined

EhsanKia: oh oops, :haHAA:

src/modules/chat/index.js
```diff
@@ -12,7 +12,7 @@ const channelEmotesTip = require('../channel_emotes_tip');
 const legacySubscribers = require('../legacy_subscribers');
 
 const EMOTE_STRIP_SYMBOLS_REGEX = /(^[~!@#$%\^&\*\(\)]+|[~!@#$%\^&\*\(\)]+$)/g;
-const MENTION_REGEX = /@\b(.*)\b/;
+const MENTION_REGEX = /^@([a-zA-Z\d_]+)$/;
```
night: i think this breaks for i18n names.. should we just match until end or symbol?

EhsanKia: Err, are we doing @ mention for displaynames or just usernames? How does that even work? I mean, clicking on it doesn't work right? Do you just care about the slight bolding?

Issue #333, opened 2015-04-11
Several chat channels


jaa121:
Hello!
What you think about the possibility of adding a few chats on one page? Like this (i know it is ugly but I have only paint ;p):

![cc](https://cloud.githubusercontent.com/assets/11669792/7102385/56936272-e07f-11e4-9cd8-b31ba6f535f5.png)

DTittel:
If all chats remained loaded at the same time I would back this "Tabbed Look" and was actually going to look into it when the feature first came out. Because each chat reloads every time you switch chats with the current Twitch implementation doing this and having it acceptably functional would require basically rewriting and overriding the entire Twitch feature.

night:
Unfortunately while BetterTTV already has support for being in multiple rooms, Twitch's internal API doesn't support multiple channel rooms being joined at the same time (they only support multiple group chat rooms).

PR #2491, opened 2017-05-30
Small CSS changes for dark mode

jjkjiang:
Really small changes to the dark mode CSS, since it looked like the notification panels didn't have many changes in dark mode while the text was changed to white by other parts of the CSS, making it really hard to read.

[Here are the visual changes in an imgur album](http://imgur.com/a/2EuyN)

I used the same color as the other panels (#1e1e1e) for the background of the notification panels, made the unread notifications to be slightly brighter (#323232) and set the hover color to be the brightest (#464646). 

The Prime tab had a grey line (#333) for the top and white line to separate elements, I thought that was weird so I changed it to all be (#333).

### DIFF
--- a/src/css/betterttv-dark.css
+++ b/src/css/betterttv-dark.css
@@ -3590,3 +3590,7 @@ circle#Oval {
 .notification-center__subheader-unread-count>span {
     color: #D3D3D3;
 }
+
+.notification__empty{
+    background-color: #1e1e1e;
+}
--- a/src/css/betterttv-dark.css
+++ b/src/css/betterttv-dark.css
@@ -3594,3 +3594,7 @@ circle#Oval {
 .notification__empty{
     background-color: #1e1e1e;
 }
+
+.notification-center__core--empty {
+    background-color: #1e1e1e;
+}
--- a/src/css/betterttv-dark.css
+++ b/src/css/betterttv-dark.css
@@ -3523,3 +3523,70 @@ button.follow-button, .subscription-modal__right button, .subscription-modal__le
 .vod-chat__scroll-wrapper::-webkit-scrollbar-thumb{
     border-color: #201c2b;
 }
+
+.offer-list__container>.ember-tether>.border-t {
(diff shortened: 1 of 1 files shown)


src/css/betterttv-dark.css
```diff
@@ -3523,3 +3523,78 @@ button.follow-button, .subscription-modal__right button, .subscription-modal__le
...
+
+.notification-center__core-end {
+    background-color: #1e1e1e;
+}
+
+.notification__container:hover {
+    background-color: #464646;
+}
+
+circle#Oval {
```
night: this needs to be scoped properly

jjkjiang:
Whoops, I thought that was the way the deny button was supposed to be.

Image of updates

This is some of the changes, but I'd like to get input on what option would be best, the original twitch light mode without bttv looks more like the ones in the following [album](http://imgur.com/a/bn0cy). Especially the purple icon one, the only problem is that it's difficult to make out the purple icon with existing css on .button--hollow:hover, setting the color on hover to much higher than the original. 

I was going to make a specific selector just for the friend request deny box so I don't break anything else on the css (new to all this, realized why I need to after you mentioned that I needed to properly scope changes), so I dug through twitch's css files for reference and found out they use rgba(100,65,164,.05), which is pretty nice looking on the twitch regular site. However, it's surprisingly really hard to make out the hover color on a dark background, and even harder to make it out if we use any brighter colors for hovering. [What it looks like when hovering, turned off notification__container hover to make it clearer](http://imgur.com/ejDgtDX).

Which, if any of these, do you think I should use?

I'm also having trouble reproducing the same scrollbar I had in my original screenshots to fix. I realized how the scrollbar background was different from the rest so I went to try and fix it, but I've been unable to get it back. For some reason on several browsers and with/without bttv, I'm getting a standard scrollbar for the prime tab and notification tab instead of the one they use for the rest of their site. I'm stuck on how I can fix that part, do you have any advice?

PR #2524, opened 2017-06-17
update Changelog panel to use new API endpoint

Nareese:


src/modules/settings/index.js
```diff
@@ -146,7 +147,20 @@ class SettingsModule {
         $('body').append(panel);
 
         cdn.get('privacy.html').then(data => $('#bttvPrivacy .tse-content').html(data));
-        cdn.get('changelog.html', true).then(data => $('#bttvChangelog .tse-content').html(data));
+
+        const changelogPanel = $('#bttvChangelog .tse-content');
+        api.get('changelog').then(data => {
+            if (data.status === 200) {
```
night: if status was not 200, it would not hit this function.. please read up on promises https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

src/modules/settings/index.js
```diff
@@ -146,7 +147,20 @@ class SettingsModule {
...
+                changelogPanel.append('<h1>Changelog</h1');
```
night: This should be in the template, not here

src/modules/settings/index.js
```diff
@@ -146,7 +147,20 @@ class SettingsModule {
...
+                data.changelog.forEach(entry => {
+                    const pubdate = window.moment(entry.publishedAt).format('MMM D, YYYY');
+                    changelogPanel.append(`<h2>Version ${entry.version} (${pubdate})</h2> ${entry.body}`);
```
night: this should be a separate template.. version should be html escaped

Nareese: 2 probably stupid questions
1/ Should body be escaped as well?
2/ Does it matter if the escape happens in the template call or in the template itself (most other examples in the code seem to do it in the template itself)

night: template itself

src/modules/settings/index.js
```diff
@@ -89,6 +89,11 @@ const settingsPanelTemplate = () => `
     </div>
 `;
 
+const changelogEntryTemplate = (version, pubdate, body) => `
+    <h2>Version ${html.escape(version)} (${pubdate})</h2>
+    ${body}
```
night: should really wrap body in a tag.. a paragraph tags be fine, for example.

src/modules/settings/index.js
```diff
@@ -150,16 +155,10 @@ class SettingsModule {
...
-                data.changelog.forEach(entry => {
-                    const pubdate = window.moment(entry.publishedAt).format('MMM D, YYYY');
-                    changelogPanel.append(`<h2>Version ${entry.version} (${pubdate})</h2> ${entry.body}`);
-                });
-            } else {
-                debug.log(`Changelog bad status: ${data.status}`);
-                changelogPanel.append('<h2>Error fetching changelog.</h2>');
-            }
+            data.changelog.forEach(entry => {
+                const pubdate = window.moment(entry.publishedAt).format('MMM D, YYYY');
```
night: date should be computed within the template, and `pubdate` is not camelcase.. i would also prefer just `date` or the original key name.

PR #2543, opened 2017-06-29
Brighten default/affiliate dark star sub badge in dark mode

### LINKED ISSUE #2541: Affiliate sub badge invisible in dark mode
Like the title says, the Affiliate sub badge is invisible in dark mode, because it's almost the same shade of black/grey

Stoupedog:
Fixes #2541

### DIFF
--- a/src/css/betterttv-dark.css
+++ b/src/css/betterttv-dark.css
@@ -58,13 +58,18 @@ body {
     background: #111 !important;
 }
 
+/* brighten default dark star sub badge */
+.ember-chat .badges img.badge[src="https://static-cdn.jtvnw.net/badges/v1/19dd8673-124d-4f44-830c-b0f4f9d78635/1"] {
+    filter: brightness(6);
+}
+
 /* live dashboard */
 div#site_header {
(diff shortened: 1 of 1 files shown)


night:
how does default twitch render it on their dark chat?

Stoupedog:
They don't even change it at all, they use exactly the same image as light mode so you can barely see it.

image

src/css/betterttv-dark.css
```diff
@@ -58,13 +58,18 @@ body {
     background: #111 !important;
 }
 
+/* brighten default dark star sub badge */
+.ember-chat .badges img.badge[src="https://static-cdn.jtvnw.net/badges/v1/19dd8673-124d-4f44-830c-b0f4f9d78635/1"] {
```
night: match on `.badges a[href*="/subscribe"] img.badge` instead

Stoupedog: That brightens every subscriber badge including custom ones uploaded by partnered streamers

night: u r right, my bad

PR #2590, opened 2017-07-30
Ability to decrease/increase VOD playback speed with '<' and '>' keys

jdpurcell:
Speed keyboard shortcuts are a feature I miss from YouTube, and even the stock Twitch clip player has it now. This PR implements the feature while playing back VODs.

Initially I tried changing the player's playbackRate directly, and although it changes the playback speed, the player UI doesn't reflect the updated speed unless playback is stopped and restarted.

Instead, I accomplished this by simulating clicks in the player settings menu. The menu items are not present in the DOM when the settings menu is hidden, so it is necessary to click the settings button and then the speed submenu to be able to access the playback speeds.


night:
sorry for the delay on this. the reason it has been sitting is because of twitch's new beta site. they have recently integrated the player as a non-iframe into the beta site. once we have ported the old video player module to it I'll ask you to update it to work on the new-twitch branch instead.

night:
finally was able to get video player module ported over.. i took a look at your PR but Twitch no longer has speed modifications in their player, so this won't work now.

jdpurcell:
Hmm? I still have speed controls when watching VODs.

image

I tried out the code in a Greasemonkey script and it still works fine - the only part that's different is the code to locate the player element, but everything inside the stepPlaybackSpeed function is exactly the same.

night:
i see.. i was testing on a live stream..

src/modules/video_player/index.js
```diff
@@ -3,23 +3,56 @@ const watcher = require('../../watcher');
...
-            break;
-        case keyCodes.f:
-            $player.find('.js-control-fullscreen').click();
-            break;
-        case keyCodes.m:
-            $player.find('.player-button--volume').click();
-            break;
+    if (keyup.shiftKey) {
+        switch (keyup.keyCode) {
+            case keyCodes.Comma:
```
night: you want keycodes for `<` and `>` .. it should be `60` and `62`

src/modules/video_player/index.js
```diff
@@ -3,23 +3,56 @@ const watcher = require('../../watcher');
...
-        case keyCodes.k:
-            $player.find('.js-control-playpause-button').click();
-            break;
-        case keyCodes.f:
-            $player.find('.js-control-fullscreen').click();
-            break;
-        case keyCodes.m:
-            $player.find('.player-button--volume').click();
-            break;
+    if (keyup.shiftKey) {
```
night: shouldnt need to handle shift with correct keycodes

src/modules/video_player/index.js
```diff
@@ -3,23 +3,56 @@ const watcher = require('../../watcher');
 const settings = require('../../settings');
 const keyCodes = require('../../utils/keycodes');
 
+function stepPlaybackSpeed($player, faster) {
```
night: instead of this really tedious dom manipulation.. you can just get the current player from our twitch util

```
const currentPlayer = twitch.getCurrentPlayer();
if (!currentPlayer) return;
currentPlayer.player.setPlaybackRate(2);
```

jdpurcell:
@night I made the requested changes.

Checking for keycodes 60/62 didn't work initially because it was using the keyup event, that's why I checked for shift + comma/period originally. I changed it to use the keypress event instead, but then I realized the letters in keycodes.js are for the uppercase letters (which keyup always sends), so there was an additional fix needed for that. I hope you're okay with the way it turned out, but if you have any other suggestions let me know.

As for the code to change the playback speed, I think it's mostly self-explanatory. The only thing I wasn't sure about was the best way to check for VOD playback, because setPlaybackRate doesn't do its own validation for that. It will happily change the rate on live videos, which is obviously no good. I ended up checking whether currentPlayer.props.vodID is set, which seems to be the best way I could find.

PR #2593, opened 2017-07-31
Fixed textareas in channel feed (again)

dansa:
Twitch removed/changed some classes from the channel feed textareas since my previous PR.

I also took the liberty to embed the channel feed textarea into a card block with background/padding because it looks very out of place as-is.

Additionally I fixed two semicolon errors in the css that crashed my linter.

---
Channel post textarea
image

Fixed
image

---

Post Modal textarea
image

Fixed
image

### DIFF
--- a/src/css/betterttv-dark.css
+++ b/src/css/betterttv-dark.css
@@ -3466,8 +3466,8 @@ button.follow-button, .subscription-modal__right button, .subscription-modal__le
 
 .js-live-widget .form__input[type=checkbox]+label:before,
 .js-live-widget .form__input[type=radio]+label:before {
-    background-color: #292929
-;    border-color: #222;
+    background-color: #292929;
+    border-color: #222;
 }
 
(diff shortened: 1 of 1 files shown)


src/css/betterttv-dark.css
```diff
@@ -3650,14 +3650,20 @@ button.follow-button, .subscription-modal__right button, .subscription-modal__le
...
 .activity-card {
     background-color: rgb(28, 28, 28);
     box-shadow: inset 0 0 0 1px #333;
     color: #d3d3d3;
 }
 
-.activity-card__comments, .activity-comments {
-    background-color: #282828;
+.activity-create--card {
+    padding: 1.5rem;
```
night: have you verified this padding on high dpi and non-high dpi screens?

dansa: I do not have a high dpi device with a development environment set up at the moment, so no, it is however the same padding that twitch uses in their post cards below.

Issue #761, opened 2015-10-12
Global Ban Protection


bufutda:
A feature where BTTV doesn't let you send a chat message if that message would breach Twitch's rate limit and give you a global ban. This would be toggle-able, and off by default.

th0mk:
I'm not a 100% sure, but i think mods can spam up to 60 messages in 30 secs before getting globally banned. Might be an idea to change it for mods of a channel.

LeotomasMC:
This would possibly be hard to do, because BTTV has no way of knowing close you are to the limit, as far as I know, its easy to manage this when working with chat bots, because you usually only have one instance of a chat bot running, and the chat bot can keep track itself. However, BTTV just changes the Twitch website on a local level, so if you send messages on 2 clients (eg. 1 channel on your computer with BTTV, and one on a phone), if you send enough messages, on the non-BTTV client, the client with BTTV will be unable to know that you've sent too many. Either way, if your encountering this on one client, your probably spamming, or at least doing something very wrong. The client on twitch also has a built in thing so you dont send messages too fast, but that might be on the API's end as well (you can see this if you copy something on your clipboard, then send a messages, then immediately paste your clipboard contents, then press enter)

night:
While this may have been useful in the past, Twitch's chat prevents you from sending too quickly these days. This avoids the possibility of receiving a chat ban.

PR #2607, opened 2017-08-10
Feature: Global Ban Protection

Rhacco:
Long-time BetterTTV user and this is my first experience with JavaScript.

I implemented a global ban protection as requested by #761 and others. Please let me know if my concept (and code) is generally acceptable or what I can improve.

Missing from my side:
1. ~~Check if the user is a moderator (or in an even higher position) to increase the message limit, see `global_ban_protection/index.js:21`. I have no idea how to do this yet, any help is appreciated.~~
2. ~~Is the `while (true)` at `global_ban_protection/index.js:34` acceptable? It generates a warning.~~


Raccoh:
Turns out checking for moderator is not very hard. Tested it as moderator and owner/broadcaster, worked fine in both cases. Cannot test it as staff, admin or global mod.

For me, the code is finished now. Feedback still welcome, of course.

night:
i dont know if this will even be possible going forward with twitch's new beta site, so i wont review it yet

Raccoh:
Thanks for your comment.

Bit unfortunate that there are these big changes coming up with Twitch's new site. I guess I'll close both my PRs for now and test and rework everything with the new environment before opening them again later.

When do you expect your changes for the beta site to be merged?

night:
> When do you expect your changes for the beta site to be merged?

you'll see a new branch when im happy with the work ive done. for now im hacking away locally until it's ok. i have a lot working, but there's some race conditions due to how im hooking into react and knowing when things are loaded is non-trivial versus ember. soon™

src/modules/global_ban_protection/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+        this.msgLimit = -1;
+        this.msgTimeStamps = new Array();
+    }
+
+    onSendMessage(sendState) {
+        if (settings.get('globalBanProtection') === false) return;
+
+        if (this.msgLimit < 0) {
+            // Do this here for a reliable result, wasn't reliable in the constructor
```
night: it wouldn't be reliable in the constructor since it's loaded on page load. twitch is a single page app, so navigating between channel pages changes the value.

src/modules/global_ban_protection/index.js
```diff
@@ -0,0 +1,61 @@
...
+            if (twitch.getCurrentUserIsOwner() || twitch.getCurrentUserIsModerator()) {
+                this.msgLimit = modMsgLimit;
```
night: instead of setting this on every send message, add a getter to this class for message limit.

```
get maxMessagesPerSecond() {
    return (twitch.getCurrentUserIsOwner() || twitch.getCurrentUserIsModerator()) ? MODERATOR_MAX_MESSAGES_PER_SECOND : MAX_MESSAGES_PER_SECOND;
}
```

src/modules/global_ban_protection/index.js
```diff
@@ -0,0 +1,61 @@
...
+                this.msgTimeStamps.unshift(oldestMsgTimeStamp);
+                done = true;
+            }
+        }
+
+        // At this point, this.msgTimeStamps.length is the number of messages
+        // sent in the last timeLimit seconds. If there is room for another
+        // message, collect its time stamp. Else, prevent the global ban.
+        if (this.msgTimeStamps.length < this.msgLimit) {
+            this.msgTimeStamps.push(Date.now());
```
night: rate limits are generally not calculated like this, and I would argue this is not how Twitch is performing their calculation.

generally limits like this are windows of wall clock time. if you send your first message at 12:12:30 the rate limit expires at 12:13:00 if it's a 30s rate limit. you would want to store an expiration time when the first message is sent, and keep a count during that duration.

sample data model:

```
{
    expires: Date.now() + RATE_LIMIT_EXPIRATION,
    count: 0,
}
```

night:

PR #2610, opened 2017-08-11
Added a distinct background color for sent messages

Rhacco:
Implemented the chat enhancement requested by #61.

I chose a shade of green for both dark and normal theme respectively which looks good in my opinion.

Do you think it's necessary to make this optional, independent from Split Chat? Users will probably want this but it seems a bit tricky to get perfectly right.


night:
i dont know how i feel about green.. i would also prefer not to add this until the changes for beta site im working on merge, since css is being moved around entirely

src/modules/split_chat/index.js
```diff
@@ -10,6 +12,14 @@ class SplitChatModule {
             defaultValue: false,
             description: 'Easily distinguish between messages from different users in chat'
         });
+        watcher.on('chat.message', ($el, msg) => {
+            if (settings.get('splitChat') === false) return;
+
+            if (twitch.getCurrentUser().name === msg.user.userLogin) {
```
night: we should be matching on user id ideally.

src/modules/split_chat/index.js
```diff
@@ -10,6 +12,14 @@ class SplitChatModule {
...
+                $el.toggleClass('bttv-own-msg-bg');
```
night: lets go with `bttv-sent-message-bg`

src/modules/split_chat/index.js
```diff
@@ -10,6 +12,14 @@ class SplitChatModule {
             defaultValue: false,
             description: 'Easily distinguish between messages from different users in chat'
         });
+        watcher.on('chat.message', ($el, msg) => {
+            if (settings.get('splitChat') === false) return;
```
night: odd that it's only available when split chat is on. what's the reasoning behind that?

night:

Issue #2496, opened 2017-06-01
Option to hide "Reruns"


Mitsuma:
It is a new feature and I'm not sure if twitch has something planned.
Somehow I think Twitch might not give us the option so I want to request if it's possible to get an option to hide the new vodcasts as they currently clutter live section in your follower list.

It would be a great addition and I'm sure there are others that don't want some VoD playlist on their live feed.

Spliffs:
yes please! I do not want to see any vodcasts in my following list. If I wanted to watch vods, I know where they are. :D Silly Twitch LUL

Snizh-Fize:
any news on this feature? those vodcasts are a joke and i don't know why they are being treated like live channels...

colincreighton:
I would also love to see this on BTTV. Looks like it is possible with some undocumented API filtering https://discuss.dev.twitch.tv/t/vodcasts-not-reported-as-playlist-streams-in-api/10603

Andrey-Sisoyev:
I guess this is hard to implement, or against BTTV authors' views. I see no other reason to have this obviously needed feature missing. 
A possible workaround for those, who are still waiting for it: just block every streamer who streams vodcasts. There is an extension for Chrome for that: "Hide unwanted streams on Twitch"
You can also vote for this feature here: https://github.com/LinogeFly/hide-unwanted-streams-on-twitch/issues/18

ericchase:
I have implemented this myself with a userscript.

qvatch:
Or perhaps filter them to their own section "Live Channels", "Live VODs", "Live Hosts", etc.

dsnvwlmnt:
Anytime now... I acquired BTTV strictly for this. Turns out it's still not implemented. :(

SlidingHorn:
This feature is almost a must anymore...half of twitch seems to be VODs/replays lately, and it's extra annoying to visually "sort through" them

Andrey-Sisoyev:
@SlidingHorn there's a beautiful extension for Chrome "Unwanted Twitch" - it solves this feature, as well as many others. Already, I was able to find many great small channels thanks to the filtering it provides.

Mitsuma:
> This feature is almost a must anymore...half of twitch seems to be VODs/replays lately, and it's extra annoying to visually "sort through" them

You might just want to install FFZ in addition to BTTV. 
It actually offers to hide re-runs and I think vodcasts as well.

ericchase:
[...]

PR #2612, opened 2017-08-12
Added hide vodcast feature

hobbycoder7:
Feature requested here: #2496

Feature was requested to work in the "Following list" of the site. Option description is: 'Hide Vodcasts in your following list' to let the user know.

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -106,9 +106,9 @@ class Watcher extends SafeEventEmitter {
                         if (lastRoute.substr(0, 19) === 'directory.following') break;
                         this.emit('load.directory.following');
                         break;
-                    case 'directory.channels.all':
-                    case 'directory.following.channels':
                     case 'directory.game.index':
+                    case 'directory.following.channels':
+                    case 'directory.channels.all':
                         this.emit('load.directory');
                         break;
                     case 'communities.community':


src/watcher.js
```diff
@@ -99,10 +99,13 @@ class Watcher extends SafeEventEmitter {
                         this.emit('load.chat');
                         break;
                     case 'directory.following.index':
+                        this.emit('load.following');
```
night: you should use the existing event

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+    registerMutationObserver() {
+        let target = document.querySelector('.js-streams .infinite-scroll.tower'); // Overview tab (twitch.tv/directory/following/live)
+        if ( !target ) target = document.querySelector('.js-streams.qa-live-streams'); // Channels tab (twitch.tv/directory/following)
+        if ( target ) {
+            // hides vodcasts when infinite scrolling
+            this.observer = new MutationObserver(() => {
+                $('.qa-stream-preview .is-watch-party').closest('.qa-stream-preview:visible').hide();
+            });
+            const config = {
```
night: this is not a needed variable since you're using it only once.. it could be passed in directly.

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    unload() {
+        if ( this.observer ) {
+            this.observer.disconnect();
+            $('.qa-stream-preview .is-watch-party').closest('.qa-stream-preview:hidden').show();
+        }
+    }
+
+    registerMutationObserver() {
```
night: in general we do not put mutation observers anywhere but the watcher.js file.. this is because if someone wants to build something relying on the same/similar mutations we dont need multiple of the observer

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+        let target = document.querySelector('.js-streams .infinite-scroll.tower'); // Overview tab (twitch.tv/directory/following/live)
```
night: this are bad ways of figuring out the tab.. use the route name in the watcher and emit an event such that you can distinguish one from the other

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    load() {
+        if (settings.get('hideVodcasts') === false) return;
+        $('.qa-stream-preview .is-watch-party').closest('.qa-stream-preview:visible').hide(); // hide vodcasts on page load
+        this.registerMutationObserver();
+    }
+
+    unload() {
+        if ( this.observer ) {
```
night: spacing

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+            this.observer.disconnect();
+            $('.qa-stream-preview .is-watch-party').closest('.qa-stream-preview:hidden').show();
```
night: reused css classes should be consts atop the file

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+
+    registerMutationObserver() {
+        let target = document.querySelector('.js-streams .infinite-scroll.tower'); // Overview tab (twitch.tv/directory/following/live)
+        if ( !target ) target = document.querySelector('.js-streams.qa-live-streams'); // Channels tab (twitch.tv/directory/following)
+        if ( target ) {
+            // hides vodcasts when infinite scrolling
+            this.observer = new MutationObserver(() => {
+                $('.qa-stream-preview .is-watch-party').closest('.qa-stream-preview:visible').hide();
```
night: probably could just hide the parent instead of checking for visibility

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+const settings = require('../../settings');
+const watcher = require('../../watcher');
+
+class HideVodcastsModule {
+    constructor() {
+        settings.add({
+            id: 'hideVodcasts',
+            name: 'Hide Vodcasts',
+            defaultValue: false,
+            description: 'Hide Vodcasts in your following list'
```
night: should it be everywhere, not just in the following list

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,46 @@
...
+        });
+        settings.on('changed.hideVodcasts', value => value === true ? this.load() : this.unload());
+        watcher.on('load.following', () => this.load());
+    }
+
+    load() {
+        if (settings.get('hideVodcasts') === false) return;
+        $('.qa-stream-preview .is-watch-party').closest('.qa-stream-preview:visible').hide(); // hide vodcasts on page load
```
night: comment is superfluous

src/modules/hide_vodcasts/index.js
night: this is also the same line as in the load.. why not make a helper function instead of copy/pasting the same line

src/watcher.js
hobbycoder7: The reason for adding the new event is because if you switch from another tab back to the index tab the event wont fire:
```
// Switching between tabs in following page
if (lastRoute.substr(0, 19) === 'directory.following') break;
```
It makes the event unreliable for this modules and I didn't want to mess with the existing event and possibly break a module that relies on it.

src/modules/hide_vodcasts/index.js
hobbycoder7: The reason for visibility checking is so that when the feature is enabled/disabled streams can be hidden or shown without having to force a page refresh.

hobbycoder7:
I've updated and committed the changes.

src/watcher.js
```diff
@@ -301,6 +313,29 @@ class Watcher extends SafeEventEmitter {
...
+                if (watcher) watcher.disconnect();
+                watcher.observe(element, {childList: true, subtree: true});
+            }, 200);
+        };
+
+        directoryWatcher = new window.MutationObserver(() => {
+            this.emit('directory.changed');
+        });
+
+        const infiniteScrollSelector = '.js-streams .infinite-scroll.tower';
```
night: constants go atop the file

src/watcher.js
```diff
@@ -99,13 +101,23 @@ class Watcher extends SafeEventEmitter {
                         this.emit('load.chat');
                         break;
                     case 'directory.following.index':
-                        this.emit('load.following');
+                        this.emit('load.following.overview');
                         // Switching between tabs in following page
                         if (lastRoute.substr(0, 19) === 'directory.following') break;
                         this.emit('load.directory.following');
                         break;
+                    case 'directory.channels.all':
```
night: these all seem to do the same thing, why are they not all grouped together as a simplistic `load.directory`

src/modules/hide_vodcasts/index.js
```diff
@@ -2,44 +2,42 @@ const $ = require('jquery');
...
+        watcher.on('load.following.overview', () => this.delayedHide());
     }
 
-    load() {
-        if (settings.get('hideVodcasts') === false) return;
-        $('.qa-stream-preview .is-watch-party').closest('.qa-stream-preview:visible').hide(); // hide vodcasts on page load
-        this.registerMutationObserver();
+    delayedHide() {
+        // timeout used because events sometimes fire before elements are created
+        setTimeout(() => {
```
night: this should not be needed

src/watcher.js
```diff
@@ -301,6 +313,29 @@ class Watcher extends SafeEventEmitter {
...
+            setTimeout(() => {
+                const element = $(selector)[0];
+                if (!element) return;
+                if (watcher) watcher.disconnect();
+                watcher.observe(element, {childList: true, subtree: true});
+            }, 200);
+        };
+
+        directoryWatcher = new window.MutationObserver(() => {
+            this.emit('directory.changed');
```
night: you should be watching for specific elements being created (i.e. creation of the vodcast streams), not firing on every change

src/watcher.js
```diff
@@ -301,6 +313,29 @@ class Watcher extends SafeEventEmitter {
 
         this.on('load.vod', () => observe(vodChatWatcher, $('.vod-chat')[0]));
     }
+
+    directoryObserver() {
+        const observe = (watcher, selector) => {
+            // timeout used because events sometimes fire before elements are created
+            setTimeout(() => {
```
night: you should not need a timeout here, if the element does not exist you've likely not selected the wrapping element properly

hobbycoder7:
Updated.

night:
sorry for the delay on this. as you know, twitch released a new beta site, which is why this has been sitting. if you update it to work on the new-twitch branch I can consider merging this finally.

hobbycoder7:
Looking at the beta site, they removed the way I was checking to see if the stream was vodcasting or not. They removed or haven't implemented the "vodcast" badge that's on the current site.

I'll keep an eye out on the beta site to see if they implement the badge and update then.

hobbycoder7:
Please close this. I've updated the code for new-twitch branch: #2791

PR #2623, opened 2017-08-14
Added 'clickToMute' setting

LyleE:
Added a new setting that is a slight variation of "Click to Play/Pause Stream" that mutes/unmutes instead. Because both of these settings use the same input (clicking on the player), I've made them mutually exclusive; enabling one will disable the other.

### DIFF
--- a/src/modules/video_player/index.js
+++ b/src/modules/video_player/index.js
@@ -24,7 +24,7 @@ function handleKeyEvent(keyup) {
 }
 
 let clicks = 0;
-function handlePlayerClick() {
+function handlePlayerClickToPause() {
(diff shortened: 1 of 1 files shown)


night:
thanks for opening a pr. unfortunately i dont think this would make a good addition. clicking to mute is not a paradigm known for video players, while clicking to play/pause is. also not a fan of mutually exclusive settings

LyleE:
Thanks, that's reasonable! Not a problem - I'll just have this [standalone script](https://gist.github.com/LyleE/563e96f5e5e101a44e3e6e1f658253ce) running instead.

PR #2625, opened 2017-08-16
Re-enable panel list styling

AlcaDesign:
Before:
image

After:
image

The selector is the same that Twitch uses which is fairly specific. An alternative may use "`#channel_panels`" instead of "`.panel-formatting`" but this allows for matching against other panels that may exist around the site.

This line from the official CSS is what disables it:
```
ol, ul, li {
	list-style: none;
}
```

despite this other styling they have that doesn't make it appear:
```
.panel-formatting .panel .description ul li {
	list-style-type: disc;
}
```


night:
this looks like a twitch bug.. and it looks like they are fixing it on the beta site. not worth fixing

Issue #2606, opened 2017-08-09
Follow command


biwegas:
it would be cool if you could make a /follow command it chat and it makes you follow the current channel if it's possible.

Stoupedog:
duplicate of #2019

biwegas:
thats not exactly a issue for me i think it would be a good feature for people who doesn't use mouse often and prefers browsing/typing with keyboard or has limited accesibility on mouse usage (like with mouse pad is not easy to get used to it fast) also it would be more convenient for example if you use it while it is in followers only mode so you quickly type /follow and can use chat afterwards

biwegas:
@night you probably didn't noticed but other guy which asked about the /follow command his intentions was way different... Nothing is broken on twitch side with follow button i want to read you what i commented on my other message. Hope you understand it's more convenient for some people to use command (there is some cases) unless ofcourse you don't really have time or will to do it. Everything is understandable from your side if you are not willing to do it.

Issue #2019, opened 2017-01-30
An option to follow the streamer with /follow command


XxBonf1r3xX:
Would that be possible to add? I hate to click on Follow button because 99% of time it's broken and it will unfollow the streamer after a while by itself. I think Twitch will register one click and after a while it will register second click to unfollow the streamer.

With this command we would not have that issue because we can't unfollow the streamer with /follow command, to unfollow we would have to use /unfollow command.

EhsanKia:
The follow button works just fine for me and never unintentionally unfollows anyone. If you're experience issues with it, it makes more sense to try and fix the button rather than add a different way of following channels.

PR #2635, opened 2017-08-19
Added /follow and /unfollow commands #2606 #2019

biwegas:
After @night closed issue #2606 and @Stoupedog made it seem like i'm having the same issue that is in #2019 I figured out i can take couple nights and learn JS and add /follow and /unfollow commands to channel. My biggest argument is there is more convenient for most people I asked around my friends and they agreed that it would be really useful to have those commands. Second argument is that some people have automatically turned on Theater mode (Function of BTTV!) and they may like to follow by just typing in command. Well at the end it's quality of life thing so it's not number one priority but imo it could help to alot of people out of laziness to not follow the streamer and support more streamers that not get much support.

### DIFF
--- a/src/css/betterttv.css
+++ b/src/css/betterttv.css
@@ -2469,8 +2469,7 @@ body.channel_new.columns.ember-application {
     display: none;
 }
 
-.chatReplay .chat-line .colon,
-.conversation-chat-line .colon {
+.chatReplay .chat-line .colon {
     margin-left: -3px;
 }
 
--- a/src/modules/chat_commands/index.js
+++ b/src/modules/chat_commands/index.js
@@ -26,6 +26,8 @@ const HELP_TEXT = `BetterTTV Chat Commands:
 /u — Shortcut for /unban
 /uptime — Retrieves the amount of time the channel has been live
 /viewers — Retrieves the number of viewers watching the channel
+/follow - follows current channel you watching.
+/unfollow - unfollows current channel you watching.
 Native Chat Commands:`;
 
 function secondsToLength(s) {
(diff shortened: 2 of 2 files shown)


AlcaDesign:
Here's _some_ problems:

- You changed all line endings from LF (`\n`) to CR+LF (`\r\n`)
- The command help text is in alphabetical order
- You didn't capitalize the command descriptions and they have periods while the rest do not.
- Your responses aren't quite right and they're **missing periods**. Here's some suggestions:
  - Wording
    - `` `You already following ${channel.displayName}` ``
    - `` `You are already following ${channel.displayName}.` ``
  - Wording
    - `` `You followed ${channel.displayName}` ``
    - `` `You are now following ${channel.displayName}.` ``
  - Period
    - `` `You unfollowed ${channel.displayName}` ``
    - `` `You unfollowed ${channel.displayName}.` ``
  - Period
    - `` `You are not following ${channel.displayName}` ``
    - `` `You are not following ${channel.displayName}.` ``
- You should save `` ``users/${currentUser.id}/follows/channels/${channel.id}` `` to a variable and reuse it. Here's an example: https://github.com/night/BetterTTV/blob/e4cfaad999a303fdef5e4e7ce3b79ad2adf82336/src/modules/chat_moderator_cards/index.js#L132-L141

I haven't tested it for myself, but I imagine there's an issue with the Promise logic.

biwegas:
Fixed *some* problems which @AlcaDesign pointed out except one with **Promise logic** *(Would be great to know what should i do about that)*

night:
sorry for the delay on this. the reason it has been sitting is because of twitch's new beta site. once we have commands support added to it I'll ask you to update it to work on the new-twitch branch instead.

biwegas:
okay will keep that in mind

night:
can you rebase this change off master? the commands stuff was re-added

biwegas:
May take some time but ill do it

night:
closing due to inactivity

PR #2646, opened 2017-08-27
Add Option to hide Recommended Friends

SexualRhinoceros:
This really annoys me and was a simple fix.


SexualRhinoceros:
just wanna acknowledge that fixing `betterttv-hide-friends.css` would also fix this. Wasn't sure if the hide all friends should be the catch all for this and still have a hide recommended friends. I personally prefer this set up but unsure what everyone else's use case would be.

PR #2647, opened 2017-08-27
Add two new options...

SexualRhinoceros:
Adds an option to hide offline followed channels. This will allow people
who follow a lot of streamers and need to always use the "Load More"
option to not have a scroll bar on their side bar. With this enabled,
following 100+ people is less of a chore.

Also adds an option to hide Recommended Friends. There was a bug in
`betterttv-hide-friends.css` causing it not no longer be removed but I
prefer having a seperate option as I heavily moderate who I add as a
friend making that section useless for me.

The former option is more useful than the latter which arose from a
simple bug so feel free to rip one and not the other.

### DIFF
--- a/src/css/betterttv-hide-offline-channels.css
+++ b/src/css/betterttv-hide-offline-channels.css
@@ -1,0 +1,3 @@
+.sc-item__followed-channel--offline {
+    display: none !important;
+}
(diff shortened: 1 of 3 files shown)


night:
hello. can you port this to the new-twitch branch. I do not wish to add new functionality to old twitch moving forward.

SexualRhinoceros:
sure, I'll address this later tonight

SexualRhinoceros:
see https://github.com/night/BetterTTV/pull/2742

PR #2650, opened 2017-08-28
update dependencies

NBruderman:
Only small numerical updates, I left out the 7.x.x update of "gulp-babel", as requested by @night


night:
i will update deps at some point. that point is not now

PR #2668, opened 2017-09-07
Fixes click to play/pause bug not resuming on click

ConnorWalsh7:
-Clicking on screen would pause the player but clicking screen to resume would do nothing
-Added a second event handler for the paused overlay screen
-Replaced .data() with .attr() because .data() caches results, so when it checks if the player is paused, it will read result from cache instead of getting up to date value
-.attr() only returns strings so changed isPaused check to a string compare 
-Created two separate methods for playing and pausing because I ran into an issue where one method for both would cause pressing the play button in the middle of screen to immediately pause after resuming on the first click, but work as intended after that

### DIFF
--- a/src/modules/video_player/index.js
+++ b/src/modules/video_player/index.js
@@ -24,7 +24,7 @@ function handleKeyEvent(keyup) {
 }
 
 let clicks = 0;
-function pausePlayer() {
+function handlePlayerClick() {
     clicks++;
     setTimeout(() => {
         if (clicks === 1) {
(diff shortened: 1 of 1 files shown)


night:
sorry for the delay. I do not want there to be two separate methods for this, and I don't think there needs to be two. should investigate why the screen is immediately pausing instead

ConnorWalsh7:
@night No worries! I fixed it up to be one method. I think the immediate pausing was just an environment issue, because I haven't been able to reproduce it, and my new update was tested in chrome and firefox and working fine

night:
looking back on this, it's actually intentionally this way.. if you click to pause you can resume by clicking the big play button on screen. optionally press k to play/pause.

Issue #2701, opened 2017-09-30
Font family applied incorrectly


d07RiV:
The font prompt suggests putting in monospace as one of the options, but doing so results in `font-family: "monospace"` style being applied, which is not a valid font family name (should be without quotes). Perhaps exceptions should be made for generic font names (serif, sans-serif, monospace, cursive, fantasy, system-ui)?

night:
pull requests welcome

d07RiV:
Okay hope I didn't mess up your code style too much.

PR #2702, opened 2017-10-01
Fix CSS for generic font families

d07RiV:
Issue #2701

### DIFF
--- a/src/modules/chat_font_settings/index.js
+++ b/src/modules/chat_font_settings/index.js
@@ -12,7 +12,7 @@ Leave the field blank to use the default.`;
 
 const STYLE_ID = 'bttv-font-size';
 
-const encodeFontFamily = (fontFamily) => {
+const encodeFontFamily = fontFamily => {
     if (['serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui'].indexOf(fontFamily) >= 0) {
         return html.escape(fontFamily);
     } else {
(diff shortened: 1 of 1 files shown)


src/modules/chat_font_settings/index.js
```diff
@@ -12,9 +12,17 @@ Leave the field blank to use the default.`;
 
 const STYLE_ID = 'bttv-font-size';
 
+const encodeFontFamily(fontFamily) => {
```
night: this is invalid JavaScript

src/modules/chat_font_settings/index.js
```diff
@@ -12,9 +12,17 @@ Leave the field blank to use the default.`;
...
+    if (['serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui'].indexOf(fontFamily)) {
+        return html.escape(fontFamily);
+    } else {
+        return '"' + html.escape(fontFamily) + '"';
```
night: you should use template literals

src/modules/chat_font_settings/index.js
```diff
@@ -12,9 +12,17 @@ Leave the field blank to use the default.`;
 
 const STYLE_ID = 'bttv-font-size';
 
+const encodeFontFamily(fontFamily) => {
+    if (['serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui'].indexOf(fontFamily)) {
```
night: indexOf returns `-1` when it doesn't find anything .. this evaluates to true when it doesn't find anything

night: there should also be a constant made of all generic names

src/modules/chat_font_settings/index.js
night: the logic can also be improved by always escaping the passed font family and then choosing to wrap it in quotes if it is not a generic name.

d07RiV:
Whoops that's embarassing, fixed those. Are those compiled into ES5 btw?

PR #2715, opened 2017-10-06
White VOD chat username theatre mode

sparcut:
Fixes problem in VOD chat replay (theatre mode) where usernames are white instead of colour.

No good:
chrome_2017-10-06_19-56-53

### DIFF
--- a/src/css/betterttv-dark.css
+++ b/src/css/betterttv-dark.css
@@ -2250,7 +2250,7 @@ span.subscribe-text {
     box-shadow: 0 0 3px 1px #333;
 }
 
-.emoticon-selector .emoticon-selector-box input:focus, , .emote-picker-and-button .tw-input {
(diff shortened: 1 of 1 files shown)


sparcut:
`better fix` commit readds original style.

Instead adds custom rule for
```css
.video-chat__message-author span { ... }
```

src/css/betterttv-dark.css
```diff
@@ -3807,6 +3807,10 @@ button.follow-button, .subscription-modal__right button, .subscription-modal__le
     color: #d3d3d3;
 }
 
+.video-chat__message-author span {
+	color: inherit;
```
night: we use spaces not tabs

sparcut: That's strange, my editor should not be set to tabs... I use spaces.

Will fix now.

sparcut:
@night Fixed tab and editor also found a few semantic errors which I included in the second commit.

PR #2720, opened 2017-10-08
[new-twitch] Remove BetterTTV Dark Theme Setting

Aaron128l:
With the addition of Twitch's Dark theme being available to [everyone](https://i.aaron128l.com/uFisu1d.png). There is no purpose to have a BetterTTV setting to enable it ourselves. Problems arise when one setting is enabled while another gets disabled creating different states.

With the removal of this setting, I supplement it with the `twilight.theme` under localstorage. This Storage object changes in accordance with the Twitch's Dark Mode toggle whose values are  `0 = Light, 1 = Dark`.

I've also replaced the only location that uses the `darkenedMode` settings. The chat colors are now calculated using `twilight.theme`.

I believe I'm thinking about this correctly but I may be mistaken in either my reasoning or my implementation.


src/modules/chat/index.js
```diff
@@ -19,7 +19,7 @@ const badgeTemplate = (url, description) => `
         <img alt="Moderator" class="chat-badge bttv-chat-badge" src="${url}" alt="" srcset="" data-a-target="chat-badge">
         <div class="tw-tooltip tw-tooltip--up tw-tooltip--align-left" data-a-target="tw-tooltip-label" style="margin-bottom: 0.9rem;">${description}</div>
     </div>
-`;
+`.replace(/\n^(?!<)\s+(?=<)/gm, '');
```
night: ?

src/watcher.js
```diff
@@ -216,6 +217,18 @@ class Watcher extends SafeEventEmitter {
         this.on('load.chat', () => observe(chatWatcher, $('.chat__container')[0]));
     }
 
+    darkObserver() {
+        const store = twitch.getConnectRoot()._context.store;
+        let currentState = store.getState().ui.theme;
+        store.subscribe(() => {
+            const newState = twitch.getConnectRoot()._context.store.getState().ui.theme;
```
night: you don need to regrab the store (which is expensive) .. just call `store.`

src/modules/chat/index.js
Aaron128l: Whoops, this was in the wrong place in general. However there does exist a bug with badges where extra space will create whitespace in a span. This will be posted in another PR.

src/watcher.js
```diff
@@ -216,6 +217,18 @@ class Watcher extends SafeEventEmitter {
...
+            if (currentState !== newState) {
```
night: I would just

```
const Themes = {
    LIGHT: 0,
    DARK: 1
};

const isDarkMode = store.getState().ui.theme === Themes.DARK;
if (settings.get('darkenedMode') !== isDarkMode) {
    settings.set('darkenedMode', isDarkMode);
}
```

src/watcher.js
```diff
@@ -216,6 +217,18 @@ class Watcher extends SafeEventEmitter {
         this.on('load.chat', () => observe(chatWatcher, $('.chat__container')[0]));
     }
 
+    darkObserver() {
```
night: you can just do something similar to `chat_state/index` instead of putting it here.. this file is mainly for watching dom changes.

src/modules/global_css/index.js
```diff
@@ -28,14 +17,22 @@ class GlobalCSSModule {
...
+        settings.on('changed.darkenedMode', isDark => this.darkChange(isDark));
+        watcher.on('dark.change', isDark => this.darkChange(isDark));
+
+        this.darkSet(settings.get('darkenedMode'));
+    }
+
+    darkSet(isDark) {
+        const connectRoot = twitch.getConnectRoot();
+        if (!connectRoot) return;
+        connectRoot._context.store.dispatch({type: 'core.ui.THEME_CHANGED', theme: (isDark ? 1 : 0)});
```
night: type should really be a constant atop the file.. theme should use a constant too (example in my other comment).

src/modules/global_css/index.js
```diff
@@ -28,14 +17,22 @@ class GlobalCSSModule {
...
-        settings.on('changed.darkenedMode', () => this.loadDark());
-        this.loadDark();
+
+        settings.on('changed.darkenedMode', isDark => this.darkChange(isDark));
+        watcher.on('dark.change', isDark => this.darkChange(isDark));
+
+        this.darkSet(settings.get('darkenedMode'));
+    }
+
+    darkSet(isDark) {
```
night: this name is funky. i think you want `setTwitchDarkMode`

PR #2742, opened 2017-10-30
Added a few options for the sidebar

SexualRhinoceros:
ported to new-twitch branch from
https://github.com/night/BetterTTV/pull/2647/files.

Also jquery is cool
and already imported so why not add an option to auto expand followed
channels as well.

The hiding of offline channels that are followed is
pretty annoying now though since you have to use javascript to get the
parent elements. I'm not a fan of the approach I used as it can bug out
when enabled before expanding the "Load More" option but will auto correct on frame resizes
(another thing I'm not a fan of).

### DIFF
--- a/src/modules/hide_sidebar_elements/index.js
+++ b/src/modules/hide_sidebar_elements/index.js
@@ -11,15 +11,39 @@ class HideSidebarElementsModule {
             description: 'The left sidebar is too cluttered, so you can remove recommended channels'
         });
         settings.add({
+            id: 'autoExpandChannels',
+            name: 'Auto Expand Followed Channels List',
+            defaultValue: false,
+            description: 'Automatically clicks the "Load More" option for you'
+        });
+        settings.add({
+            id: 'hideRecommendedFriends',
+            name: 'Hide Recommended Friends',
+            defaultValue: true,
+            description: 'Hides the Recommended Friends section so you have more room for activities!'
+        });
+        settings.add({
+            id: 'hideOfflineFollowedChannels',
+            name: 'Hide Offline Followed Channels',
+            defaultValue: false,
+            description: 'Hides all offline followed channels for those who follow a ton of channels'
+        });
+        settings.add({
             id: 'hidePrimePromotion',
             name: 'Hide Prime Promotions',
             defaultValue: false,
             description: 'Hides the "Free With Prime" section of the sidebar'
         });
         settings.on('changed.hideFeaturedChannels', () => this.toggleFeaturedChannels());
+        settings.on('changed.autoExpandChannels', () => this.toggleAutoExpandChannels());
+        settings.on('changed.hideRecommendedFriends', () => this.toggleRecommendedFriends());
+        settings.on('changed.hideOfflineFollowedChannels', () => this.toggleOfflineFollowedChannels());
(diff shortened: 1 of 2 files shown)


src/modules/hide_sidebar_elements/index.js
```diff
@@ -28,6 +52,29 @@ class HideSidebarElementsModule {
...
+    }
+
+    toggleOfflineFollowedChannels() {
+        var resizeTimer = null;
+        function hideOfflineUser() {
+            if (settings.get('hideOfflineFollowedChannels') === true) {
+                $('.side-nav-card__avatar--offline').parent().css('cssText', 'display: none!important');
+            }
+        };
+        $(window).bind('resize', function() {
```
night: not a fan of this solution. is there not a way to do this using css alone? if so, can you explain?

SexualRhinoceros: in the old layout, yes. but from what I can gather, the best way to determine offline status thru css is 2 layers too deep meaning some jquery can do the trick at finding the desired element. 

This would be the end of it if resizing the window didn't cause every single element to delete and redraw itself clearing out the old css overrides.

From what I've found, the best solution is to wait for [CSS4 Selector Overrides](https://caniuse.com/#feat=css-has) to be supported by literally any single browser and not have to rely on Javascript to do this unless a better Javascript solution exists that I don't know of yet

night: couldn't you just apply a css class that performs a display none?

SexualRhinoceros: no since if you apply it to the class which has the flag for offline, it'll only hide the profile picture. you'd need the parent of the class with the offline flag and there isn't a supported parent selector in css (yet)

[this is the layout for offline user](https://i.imgur.com/rfTIn4X.png) while [this is the layout for an online user](https://i.imgur.com/um5jloc.png).

The only discernable difference is the --offline flag in the avatar element :/

night: I meant use js to apply the class.. it shouldn't cause you issues on resize then

SexualRhinoceros:
Okay, after sitting here looking, I noticed that offline channels have an href of `*channel name*/videos/all` while online channels go straight to the live channel. This should provide the functionality desired with no extra JS at all

src/modules/hide_sidebar_elements/index.js
```diff
@@ -28,6 +52,20 @@ class HideSidebarElementsModule {
...
+            $('.side-nav-load-more__button').trigger('click');
+        }
+    }
+
+    toggleRecommendedFriends() {
+        $('body').toggleClass('bttv-hide-recommended-friends', settings.get('hideRecommendedFriends'));
+    }
+
+    toggleOfflineFollowedChannels() {
+        $('body').toggleClass('bttv-hide-followed-offlines', settings.get('hideOfflineFollowedChannels'));
```
dansa: I believe this is supposed to be `bttv-hide-followed-offline` and not `bttv-hide-followed-offlines` ?

SexualRhinoceros: yes, I'm bad at words

PR #2791, opened 2017-11-22
[new-twitch] Added hide vodcast feature

hobbycoder7:
Feature requested here: #2496

The new twitch site added the Vodcast pill to the /following and /following/live directories so I've updated the code for the new site.


src/watcher.js
```diff
@@ -266,6 +271,29 @@ class Watcher extends SafeEventEmitter {
...
+
+                    const vodCasts = $el.find('.tw-pill');
+
+                    if (vodCasts.length > 0) {
+                        this.emit('directory.vodcast');
+                    }
+                }
+            })
+        );
+        this.on('load.directory.following', () => observe(directoryWatcher, $('.twilight-main')[0]));
```
night: twilight-main is not scoped enough .. a page change should break the observer such that it isn't constantly watching this dom node.

src/watcher.js
```diff
@@ -135,6 +137,9 @@ class Watcher extends SafeEventEmitter {
                 case routes.DIRECTORY_FOLLOWING:
                     this.waitForLoad('following').then(() => this.emit('load.directory.following'));
                     break;
+                case routes.DIRECTORY_FOLLOWING_LIVE:
+                    this.waitForLoad('following').then(() => this.emit('load.directory.following'));
```
night: this can cause an infinite loop.. this calls a function that clicks a link to load this ...

src/watcher.js
```diff
@@ -266,6 +271,29 @@ class Watcher extends SafeEventEmitter {
...
+
+        directoryWatcher = new window.MutationObserver(mutations =>
+            mutations.forEach(mutation => {
+                for (const el of mutation.addedNodes) {
+                    const $el = $(el);
+
+                    const vodCasts = $el.find('.tw-pill');
+
+                    if (vodCasts.length > 0) {
+                        this.emit('directory.vodcast');
```
night: this should pass an element .. not rely on `this`

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,31 @@
...
+            description: 'Hide Vodcasts from directory listings'
+        });
+        settings.on('changed.hideVodcasts', value => value === true ? this.hide() : this.show());
+
+        watcher.on('directory.vodcast', () => this.hide());
+    }
+
+    hide() {
+        if (settings.get('hideVodcasts') === false) return;
+        $(VODCAST_SELECTOR).closest(CARD_SELECTOR).parent(':visible').hide();
```
night: it would be nice if we can avoid a mutation observer.. i dont see vodcast pills so either they disabled them or it isnt fully rolled out yet.. without seeing them myself i cant tell you if this is the best way

src/watcher.js
hobbycoder7: Added followingLive to loadPredicates. Was going to stick with just following but in the debug it would show following when loading following/live. Wasn't sure if this was needed though since it's the exact same as following just renamed.

Added load.directory.following.live for DIRECTORY_FOLLOWING_LIVE route.

src/watcher.js
hobbycoder7: Changed vodCasts to $vodCasts and added that to this.emit: this.emit('directory.vodcast', $vodCasts);

src/watcher.js
hobbycoder7: Found and changed a closer element that's removed when you navigate away from the directory area: 'main .scrollable-area .pd-3'

src/modules/hide_vodcasts/index.js
hobbycoder7: The mutation observer was used since there's a loading icon before live channels are displayed and it's what I had used in the pre new-twitch code.

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,32 @@
...
+        });
+        settings.on('changed.hideVodcasts', value => value === true ? this.hide() : this.show());
+
+        watcher.on('directory.vodcast', () => this.hide());
+        watcher.on('load.directory.following', () => this.hide());
+    }
+
+    hide() {
+        if (settings.get('hideVodcasts') === false) return;
+        $(VODCAST_SELECTOR).closest(CARD_SELECTOR).parent(':visible').hide();
```
night: this selector is dangerous .. if twitch changes their css to remove the parent element you could end up hiding the entire directory. you should select the parent by class and check if that is visible or hidden.

src/watcher.js
```diff
@@ -8,11 +8,13 @@ let router;
 let currentPath = '';
 let chatWatcher;
 let vodChatWatcher;
+let directoryWatcher;
 let currentChatReference;
 let channel = {};
 
 const loadPredicates = {
     following: () => !!$('.following__header-tabs').length,
+    followingLive: () => !!$('.following__header-tabs').length,
```
night: i dont understand why this is a separate predicate

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,32 @@
...
+        settings.add({
+            id: 'hideVodcasts',
+            name: 'Hide Vodcasts',
+            defaultValue: false,
+            description: 'Hide Vodcasts from directory listings'
+        });
+        settings.on('changed.hideVodcasts', value => value === true ? this.hide() : this.show());
+
+        watcher.on('directory.vodcast', () => this.hide());
+        watcher.on('load.directory.following', () => this.hide());
```
night: this should not be needed

src/modules/hide_vodcasts/index.js
```diff
@@ -0,0 +1,32 @@
...
+    constructor() {
+        settings.add({
+            id: 'hideVodcasts',
+            name: 'Hide Vodcasts',
+            defaultValue: false,
+            description: 'Hide Vodcasts from directory listings'
+        });
+        settings.on('changed.hideVodcasts', value => value === true ? this.hide() : this.show());
+
+        watcher.on('directory.vodcast', () => this.hide());
```
night: `.hide` should hide a single vodcast emitted from the watcher, not target every vodcast on the directory

src/watcher.js
```diff
@@ -135,6 +138,9 @@ class Watcher extends SafeEventEmitter {
                 case routes.DIRECTORY_FOLLOWING:
                     this.waitForLoad('following').then(() => this.emit('load.directory.following'));
                     break;
+                case routes.DIRECTORY_FOLLOWING_LIVE:
+                    this.waitForLoad('followingLive').then(() => this.emit('load.directory.following.live'));
```
night: why does this only target the following pages? are vodcasts on the main directory pages?

src/watcher.js
```diff
@@ -266,6 +272,30 @@ class Watcher extends SafeEventEmitter {
...
+                    const $vodCasts = $el.find('.tw-pill');
+
+                    if ($vodCasts.length > 0) {
+                        this.emit('directory.vodcast', $vodCasts);
+                    }
+                }
+            })
+        );
+        this.on('load.directory.following', () => observe(directoryWatcher, $('main .scrollable-area .pd-3')[0]));
+        this.on('load.directory.following.live', () => observe(directoryWatcher, $('main .scrollable-area .pd-3')[0]));
```
night: these both have the same anonymous function..

src/watcher.js
```diff
@@ -266,6 +272,30 @@ class Watcher extends SafeEventEmitter {
...
+        const observe = (watcher, element) => {
+            if (!element) return;
+            if (watcher) watcher.disconnect();
+            watcher.observe(element, {childList: true, subtree: true});
+        };
+
+        directoryWatcher = new window.MutationObserver(mutations =>
+            mutations.forEach(mutation => {
+                for (const el of mutation.addedNodes) {
+                    const $el = $(el);
```
night: added nodes only returns added nodes, what about initial load for the case where betterttv loads after the elements have rendered?

src/watcher.js
hobbycoder7: I've only seen vodcasts on following pages and only in the Overview and Channels tab.

src/watcher.js
hobbycoder7: I'm not sure. I had it just hide on load to fix that.

night:
please rebase your changes onto master

Qrokqt:
There's a nice class you could use to check now, something like `.stream-type-indicator:not(.stream-type-indicator--live)`

Issue #2787, opened 2017-11-20
Emote menu is almost empty


Honkizz:
The emote menu has only 3 emotes ClappyHype, ClappyDerp and Clappy.

night:
Not able to reproduce this issue. Try disabling third party extensions and make sure you’re logged into Twitch

xgerhard:
I do see this happening sometimes, sometimes a hard refresh fixes it.
It loads the BTTV channel emotes, then the new clips emotes as Honkizz describes, the rest fails/is empty.
No errors in console, according to the network tab, all emotes are loaded in, but for some reason not added to the menu.

To reproduce this seems to work for me on Chrome (With only the BTTV extension enabled):
- Open incognito tab
- Go to a random channel
- Sign in with Twitch
- Enable BTTV emote menu
- Close settings
- Open BTTV emote menu

Hopefully this helps to resolve it.

BradyThe:
I had the same problem with my default browser. The emote menu looked like this:
![firefox_2017-11-21_22-37-22](https://user-images.githubusercontent.com/8181555/33099248-863db298-cf10-11e7-8859-6c71b3d7624b.png)

I precautionally cleared all cookies from Firefox, and to eliminate any incompatibilities or override I tested it on different browsers:

Edge (never used before)
![applicationframehost_2017-11-21_22-54-23](https://user-images.githubusercontent.com/8181555/33099529-76a8b3b8-cf11-11e7-8536-6acb2ebcd5ee.png)

Chrome (used before, different profile)
![chrome_2017-11-21_22-55-16](https://user-images.githubusercontent.com/8181555/33099547-86f69f82-cf11-11e7-85bd-e03a52d3523c.png)

BradyThe:
I had the same problem with my default browser (Firefox). The emote menu looked like this:
![firefox_2017-11-21_22-37-22](https://user-images.githubusercontent.com/8181555/33099571-98faf3d6-cf11-11e7-85fe-7a4b780846f6.png)

I precautionally cleared all cookies from Firefox, and to eliminate any incompatibilities or override I tested it on different browsers with a Twitch account that makes use of BetterTTV channel emotes:

Edge (never used before)
![applicationframehost_2017-11-21_22-54-23](https://user-images.githubusercontent.com/8181555/33099783-661b75fc-cf12-11e7-8642-e2903b22fe3b.png)

Chrome (used before, different profile)
[...]

PR #2792, opened 2017-11-22
[new-twitch] Emote Menu fix

hobbycoder7:
Fix for #2787

This worked for me on Chrome. Before this fix I would only get 3 BTTV emotes showing in the menu. All I did was move the injection below the emote registration code.

Note: opening the menu too fast will cause no emotes to appear. Closing and reopening the menu fixes that but maybe a delay should be added until the emote button appears?


night:
If the injection happens after then the emote menu will not show our emotes..

hobbycoder7:
Ah!

It looks like it's a problem with the emote addon itself :(

https://twitter.com/ryanbchatham/status/930673440247468032

Issue #2676, opened 2017-09-14
top cheer option to close wont close


nufohdz88:
top cheer option to close wont close on twitch

mattj456:
in the meantime, /localunpin will do the same thing :)

PR #2796, opened 2017-11-24
[new-twitch] Dismiss pinned cheer on click

hobbycoder7:
Just copied over the old code and updated the class names.

#2676


night:
did you even test this? there's no icon rendered where you're marking a clickable area

hobbycoder7:
I was unable to find in the old code where an icon was added so just made it close on clicking the background area.

night:
Original css for that is https://github.com/night/BetterTTV/blob/master/src/css/betterttv.css#L2306

Issue #2716, opened 2017-10-07
host button


Roseii:
The host button isn't showing on my BTTV settings like it was last night

Roseii:
It isn't showing on my twitch when I'm watching someone and it doesn't show on the list to enable for twitch in the settings

night:
#2605

PR #2798, opened 2017-11-24
[new-twitch] Host Button

hobbycoder7:
To be fair, I don't know how this looked before so I just copied the old code. Also since I removed the TMI stuff it doesn't update the hosting state.

I don't have the ability to host myself so this really needs testing by someone who can. It just tells me I don't have the permission to do that. Which is the same message I get when I manually try to use /host channelName in chat.

#2716

### DIFF
--- a/src/modules/host_button/index.js
+++ b/src/modules/host_button/index.js
@@ -1,0 +1,67 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+const watcher = require('../../watcher');
+const twitch = require('../../utils/twitch');
+
+let $hostButton;
+let hosting = false;
+
+class HostButtonModule {
+    constructor() {
+        settings.add({
+            id: 'hostButton',
+            name: 'Host Button',
+            defaultValue: false,
+            description: 'Places a Host/Unhost button below the video player'
+        });
+        settings.on('changed.hostButton', value => value === true ? this.load() : this.unload());
+        watcher.on('load.chat', () => this.load());
+    }
+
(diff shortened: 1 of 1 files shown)


src/modules/host_button/index.js
```diff
@@ -0,0 +1,67 @@
...
+        $hostButton.addClass('button action button--hollow mg-l-1');
+        $hostButton.appendTo('.channel-info-bar__action-container');
+        $hostButton.click(() => this.toggleHost());
+    }
+
+    toggleHost() {
+        const command = hosting ? 'unhost' : 'host';
+        try {
+            const channelName = twitch.getCurrentChannel().name;
+            twitch.sendChatMessage(`/${command === 'host' ? `${command} ${channelName}` : command}`);
```
night: this only works in your own channel.. not the current channel

src/modules/host_button/index.js
```diff
@@ -0,0 +1,67 @@
...
+        if (currentUser.id === currentChannel.id) return;
+        this.embedHostButton();
+    }
+
+    embedHostButton() {
+        if ($('#bttv-host-button').length) return;
+        $hostButton = $('<button><span>Host</span></button>');
+        $hostButton.attr('id', 'bttv-host-button');
+        $hostButton.addClass('button action button--hollow mg-l-1');
+        $hostButton.appendTo('.channel-info-bar__action-container');
```
night: you need to append to the flex box container ..

PR #2800, opened 2017-11-25
[new-twitch] Add optional reason box for custom timeouts

lolPants:
Add a setting that enables a prompt window to enter a reason for the custom timeout menu.
Useful for Twitch channels that require you to put a reason down on bans and timeouts.

### DIFF
--- a/src/modules/chat_custom_timeouts/index.js
+++ b/src/modules/chat_custom_timeouts/index.js
@@ -2,6 +2,7 @@ const $ = require('jquery');
 const watcher = require('../../watcher');
 const keyCodes = require('../../utils/keycodes');
 const twitch = require('../../utils/twitch');
+const settings = require('../../settings');
 
 const CHAT_ROOM_SELECTOR = '.chat__pane';
 const CHAT_LINE_SELECTOR = '.chat-line__message';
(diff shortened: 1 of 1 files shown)


src/modules/chat_custom_timeouts/index.js
```diff
@@ -29,15 +30,30 @@ const ACTION_TYPES = {
...
+}
+
 function handleTimeoutClick(e) {
     const $customTimeout = $(`#${CUSTOM_TIMEOUT_ID}`);
     if (!$customTimeout.length || e.which === keyCodes.DOMVKCancel || e.shiftKey) return;
 
     if ($customTimeout.is(':hover')) {
         if (action.type === ACTION_TYPES.BAN) {
-            twitch.sendChatMessage(`/ban ${user}`);
+            const reason = setReason(action.type);
```
night: reason can be set outside these ifs since it's duplicate code.

src/modules/chat_custom_timeouts/index.js
```diff
@@ -29,15 +30,30 @@ const ACTION_TYPES = {
 let action;
 let user;
 
+function setReason(type) {
+    if (settings.get('customTimeoutReasons')) {
+        const reason = prompt(`Enter ${type === ACTION_TYPES.BAN ? 'ban' : 'timeout'} reason: (leave blank for none)`);
```
night: i would inject `${type}` directly .. and swap the timeout action type to `timeout` instead of `time`

src/modules/chat_custom_timeouts/index.js
```diff
@@ -29,15 +30,30 @@ const ACTION_TYPES = {
...
+        if (
```
night: why not

```
return reason ? ` ${reason}` : '';
```

src/modules/chat_custom_timeouts/index.js
```diff
@@ -29,15 +30,30 @@ const ACTION_TYPES = {
 let action;
 let user;
 
+function setReason(type) {
+    if (settings.get('customTimeoutReasons')) {
```
night: you can avoid nesting by 

```
if (!settings.get('customTimeoutReasons')) return '';
```

src/modules/chat_custom_timeouts/index.js
```diff
@@ -129,6 +137,13 @@ function handleClick(e) {
 class ChatCustomTimeoutsModule {
     constructor() {
         watcher.on('load.chat', () => this.loadClickHandler());
+
+        settings.add({
```
night: my last question is should this be a setting. the settings menu is getting long, i'd prefer less options

lolPants: It was my thought that some people wouldn't want an intrusive popup every time and just wanted it as it was, but anyone who does want to add reasons can turn it on.

night:
If you press shift while selecting it will pop a reason box.. subtle features that don't require options tend to be better

PR #2809, opened 2017-11-28
[new-twitch] Fix chat with recent UI update

DayBr3ak:
Updated css selector for the chat component


Aaron128l:
There are more classes that needed to be changed, including the emote menu. This is handled in #2810

DayBr3ak:
right

PR #2812, opened 2017-11-29
[new-twitch] [Feature] Show Deleted Chat Messages

DayBr3ak:
Show deleted chat message like the master branch did.


src/modules/chat_deleted_messages/index.js
```diff
@@ -0,0 +1,96 @@
...
+            childList: true,
+            subtree: true
+        });
+    }
+
+    patchClearChat() {
+        const chatBuffer = twitch.getChatController().chatBuffer;
+        const consumeEvent = chatBuffer.consumeChatEvent;
+        chatBuffer.consumeChatEvent = function(event) {
+            if (event.type === 17) {
```
night: event types are defined as a constant in the twitch util.. should use that instead

src/modules/chat_deleted_messages/index.js
```diff
@@ -0,0 +1,96 @@
...
+                // prevent chat clear
```
night: useless comment

src/modules/chat_deleted_messages/index.js
```diff
@@ -0,0 +1,96 @@
...
+
+const CHAT_CONTAINER = '.chat-room__container';
+const CHAT_LINE_SELECTOR = '.chat-line__message';
+const CHAT_LINE_DELETED_CLASS = 'chat-line__message--deleted';
+
+function chatOberserverHandler(mutations) {
+    const showDeleted = settings.get('showDeletedMessages');
+    const hideDeleted = settings.get('hideDeletedMessages');
+    if (!showDeleted && !hideDeleted) {
+        // return early
```
night: useless comment

src/modules/chat_deleted_messages/index.js
```diff
@@ -0,0 +1,96 @@
...
+
+    if (hideDeleted) {
+        elementsToRemove.forEach(el => el[0].remove());
+    }
+
+    if (showDeleted && !hideDeleted) {
+        if (elementsToRemove.length > 0) {
+            elementsToReattach.forEach(([target, node]) => {
+                target.appendChild(node);
+                target.style.opacity = 0.5;
```
night: styles should be css

src/modules/chat_deleted_messages/index.js
```diff
@@ -0,0 +1,96 @@
...
+        this.chatObserver.observe(document.querySelector(CHAT_CONTAINER), {
+            childList: true,
+            subtree: true
+        });
+    }
+
+    patchClearChat() {
+        const chatBuffer = twitch.getChatController().chatBuffer;
+        const consumeEvent = chatBuffer.consumeChatEvent;
+        chatBuffer.consumeChatEvent = function(event) {
```
night: you should prevent this from being double monkey patched

src/modules/chat_deleted_messages/index.js
```diff
@@ -0,0 +1,96 @@
...
+        // return early
+        return;
+    }
+    const elementsToReattach = [];
+    const elementsToRemove = [];
+
+    mutations.forEach(mutation => {
+        if (mutation.type === 'childList' && mutation.target.matches(CHAT_LINE_SELECTOR)) {
+            mutation.addedNodes.forEach(node => {
+                if (!node.__is_prevented && node.classList && node.classList.contains(CHAT_LINE_DELETED_CLASS)) {
```
night: i've tried something similar to this in the past.. it wasn't performant enough and caused the chat to flicker.. we might want to monkeypatch and handle deletes instead

src/modules/chat_deleted_messages/index.js
DayBr3ak: Those constants seems to be out of date as 17 is chat clear right now, and it's 16 in the twitch util.

src/modules/chat_deleted_messages/index.js
DayBr3ak: I've used the mutation observer on 30k+ channels daily with tampermonkey, and so far I've had no problems with chrome 61 and 62 on my potato computer. 

Was this flickering consistent?

DayBr3ak:
About the TMI events in twitch utils, they seem to be out of date. Chat clear is 17 now. Hence why I used a self defined constant. I think fixing this constant list should be in a separate PR.

DayBr3ak:
I've change the feature to respond to tmi events instead. I've also updated the tmi constants.
A review would be welcome when you have the time @night.

DayBr3ak:
Rebased.

src/watcher.js
```diff
@@ -266,6 +267,40 @@ class Watcher extends SafeEventEmitter {
         this.on('load.chat', updateChannel);
         this.on('load.vod', updateChannel);
     }
+
+    tmiEventObserver() {
```
night: `chatBufferObserver`

src/watcher.js
```diff
@@ -266,6 +267,40 @@ class Watcher extends SafeEventEmitter {
...
+        const PATCHED_SENTINEL = () => {};
```
night: constants should be pulled out .. you can also just use a Symbol

src/watcher.js
```diff
@@ -266,6 +267,40 @@ class Watcher extends SafeEventEmitter {
...
+        let twitchConsumeEvent;
+        const watcher = this;
+        function bttvConsumeEvent(event) {
```
night: unsure why this is named function and the function below is an arrow function

src/watcher.js
```diff
@@ -266,6 +267,40 @@ class Watcher extends SafeEventEmitter {
         this.on('load.chat', updateChannel);
         this.on('load.vod', updateChannel);
     }
+
+    tmiEventObserver() {
+        const PATCHED_SENTINEL = () => {};
+        let twitchConsumeEvent;
+        const watcher = this;
```
night: would prefer a separate function for emitting an event than storing a reference to this.

src/watcher.js
```diff
@@ -266,6 +267,40 @@ class Watcher extends SafeEventEmitter {
...
+        function bttvConsumeEvent(event) {
+            let prevent = false;
+            event.preventDefault = () => {
```
night: you shouldn't mutate Twitch's record here, you should instead copy it for our internal use.

also more performant to not keep creating a new function on each event..

since the logic of this observer is more complicated than the rest and this file is getting large, perhaps it's time to break out into a new `watchers` folder.

src/watcher.js
```diff
@@ -266,6 +267,40 @@ class Watcher extends SafeEventEmitter {
...
+                prevent = true;
+            };
+
+            watcher.emit('tmi.event', event);
```
night: `chat.buffer.event` seems better

src/watchers/chat-buffer-watcher.js
```diff
@@ -0,0 +1,38 @@
+const twitch = require('../utils/twitch');
+
+const PATCHED_SYMBOL = Symbol('patched');
+
+function chatBufferWatcher(watcher) {
```
DayBr3ak: @night Is this what you meant by a watchers folder?

DayBr3ak:
This also fixes the chat state since the ROOM_STATE event was wrong.

PR #2827, opened 2017-12-02
[new-twitch] [Fix] Bttv Channel Emotes syncing when changing channel.

DayBr3ak:
Currently, changing channel doesn't always update the bttv channel emotes. A refresh is usually needed. This PR fixes this.

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -255,13 +255,18 @@ class Watcher extends SafeEventEmitter {
             const currentChannel = twitch.getCurrentChannel();
             if (!currentChannel) return;
 
-            if (currentChannel.id === channel.id) return;
+            if (currentChannel.name === channel.name) return;
             channel = currentChannel;
             api
                 .get(`channels/${channel.name}`)
-                .then(d => this.emit('channel.updated', d));
+                .catch(error => ({
+                    bots: [],
+                    emotes: [],
+                    status: error.status || 0
+                }))
+                .then(data => this.emit('channel.updated', data));
+            return true;
         };
-
 
         this.on('load.chat', updateChannel);
         this.on('load.vod', updateChannel);
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -255,7 +255,7 @@ class Watcher extends SafeEventEmitter {
             const currentChannel = twitch.getCurrentChannel();
             if (!currentChannel) return;
 
-            if (currentChannel.name === channel.name) return;
+            if (currentChannel.id === channel.id) return;
             channel = currentChannel;
             api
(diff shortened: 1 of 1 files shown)


src/watcher.js
```diff
@@ -23,7 +23,12 @@ const loadPredicates = {
         const lastReference = currentChatReference;
         const currentChat = twitch.getCurrentChat();
 
-        if (currentChat && currentChat === lastReference) return false;
+        if (currentChat && currentChat === lastReference) {
+            return false;
+        }
+        if (lastReference && currentChat.props.channelID.toString() === lastReference.props.channelID.toString()) {
```
night: i actually want it to reload from our api in the case when a user navs to a new page from the same user.. emotes and bots can get updated

src/watcher.js
```diff
@@ -259,7 +264,16 @@ class Watcher extends SafeEventEmitter {
             channel = currentChannel;
             api
                 .get(`channels/${channel.name}`)
-                .then(d => this.emit('channel.updated', d));
+                .then(d => {
```
night: would prefer these without brackets

src/watcher.js
```diff
@@ -259,7 +264,16 @@ class Watcher extends SafeEventEmitter {
...
+                    this.emit('channel.updated', d);
+                })
+                .catch(error => {
+                    this.emit('channel.updated', {
+                        bots: [],
+                        emotes: [],
+                        status: error.status ? error.status : 404
```
night: status should be 0 if unknown.. you can also `error.status || 0`

src/watcher.js
```diff
@@ -255,14 +255,19 @@ class Watcher extends SafeEventEmitter {
             const currentChannel = twitch.getCurrentChannel();
             if (!currentChannel) return;
 
-            if (currentChannel.id === channel.id) return;
+            if (currentChannel.name === channel.name) return;
```
DayBr3ak: The name is updated with the `load.chat` event. It's not always the case for the id.

DayBr3ak:
If this is not suitable, maybe: Actually wait for the channel id to be updated (like the watcher `waitFor` does). Or: subscribe to the graphql client and watch channel updates requests?

src/watcher.js
night: im not sure why this was changed or why it's needed

DayBr3ak:
@night this issue is still present, the channel ID is not updated when you switch channels. Bttv emotes are then not updated.

night:
Unfortunately I cannot reproduce that issue. If you have reproduction steps make an issue/PR with steps so I can test.

PR #2830, opened 2017-12-02
[new-twitch] [Feature] Auto dismiss the Miniplayer

DayBr3ak:
This PR adds the possibility to auto dismiss the miniplayer on leaving the channel.

### DIFF
--- a/src/modules/miniplayer-disable/index.js
+++ b/src/modules/miniplayer-disable/index.js
@@ -3,8 +3,16 @@ const debug = require('../../utils/debug');
 const settings = require('../../settings');
 const watcher = require('../../watcher');
 
-const MINIPLAYER_CONTAINER = '[data-a-target=root-scroller]';
 const MINIPLAYER_DISMISS = '[data-test-selector=persistent-player-mini-dismiss]';
+
+function dismiss(shouldDismiss = true) {
+    if (!shouldDismiss) return;
+    if ($(MINIPLAYER_DISMISS)[0]) {
+        if (!settings.get('disableMiniplayer')) return;
+        $(MINIPLAYER_DISMISS).click();
+        debug.log('[Miniplayer] Auto dismissed');
+    }
+}
 
 class MiniplayerDisableModule {
     constructor() {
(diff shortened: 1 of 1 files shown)


src/modules/miniplayer-disable/index.js
```diff
@@ -0,0 +1,31 @@
+const $ = require('jquery');
+const debug = require('../../utils/debug');
+const settings = require('../../settings');
+const watcher = require('../../watcher');
+
+const MINIPLAYER_DISMISS = '[data-test-selector=persistent-player-mini-dismiss]';
```
night: this selector seems nonspecific, matching every element with that attribute

src/modules/miniplayer-disable/index.js
```diff
@@ -0,0 +1,31 @@
...
+class MiniplayerDisableModule {
+    constructor() {
+        settings.add({
+            id: 'disableMiniplayer',
+            name: 'Disable Miniplayer',
+            defaultValue: false,
+            description: 'Disable the Miniplayer when leaving the channel'
+        });
+
+        watcher.on('load', () => setTimeout(dismiss, 25));
```
night: if you need a set timeout then you have a race condition. this is not how you solve a race condition unfortunately

src/modules/miniplayer-disable/index.js
```diff
@@ -0,0 +1,31 @@
...
+const debug = require('../../utils/debug');
+const settings = require('../../settings');
+const watcher = require('../../watcher');
+
+const MINIPLAYER_DISMISS = '[data-test-selector=persistent-player-mini-dismiss]';
+
+function dismiss(shouldDismiss = true) {
+    if (!shouldDismiss) return;
+    if ($(MINIPLAYER_DISMISS)[0]) {
+        if (!settings.get('disableMiniplayer')) return;
```
night: im more surprised you're not just hiding this with css.. `display: none` should take care of the video element iirc

night:
i take it back, twitch already has this natively. if it's bugged it's something they need to fix

PR #2832, opened 2017-12-02
[new-twitch] [Feature] Add adaptive chat feature

x000000:
On vertical monitors it might be convenient to have chat below the player. Just added this feature, taht's it.


night:
Thanks for your PR. Unfortunately we won't do heavy modifications to the page like this anymore, sorry. If you can find a way to do this with CSS alone we can consider it, but DOM changes being required is a little too sketchy for us. We barely have enough time to maintain things as-is, and heavy DOM changes break frequently.

PR #2833, opened 2017-12-02
[new-twitch] [Feat] Anon chat & Chat Commands

DayBr3ak:
Refactored the old versions features. Everything seems to be working. I Enjoy self sub mode very much :) .

### DIFF
--- a/src/modules/chat_commands/index.js
+++ b/src/modules/chat_commands/index.js
@@ -2,13 +2,11 @@ const $ = require('jquery');
 const moment = require('moment');
 const twitch = require('../../utils/twitch');
 const twitchAPI = require('../../utils/twitch-api');
-const tmiAPI = require('../../utils/tmi-api');
 const chat = require('../chat');
 const anonChat = require('../anon_chat');
 
 const HELP_TEXT = `BetterTTV Chat Commands:
 /b — Shortcut for /ban
-/chatters — Retrieves the number of chatters in the chat
 /followed — Tells you for how long you have been following a channel
 /follows — Retrieves the number of followers for the channel
 /join & /part — Temporarily join/part chat (anon chat)
@@ -16,7 +14,6 @@ const HELP_TEXT = `BetterTTV Chat Commands:
 /localasciioff — Turns off local ascii-only mode
 /localmod — Turns on local mod-only mode (only your chat is mod-only mode)
 /localmodoff — Turns off local mod-only mode
-/localunpin — Removes pinned cheer locally
 /localsub — Turns on local sub-only mode (only your chat is sub-only mode)
 /localsuboff — Turns off local sub-only mode
 /massunban (or /unban all or /u all) — Unbans all users in the channel (channel owner only)
@@ -171,11 +168,6 @@ function handleCommands(message) {
             command === 'join' ? anonChat.join() : anonChat.part();
             break;
 
-        // info
-        case 'chatters':
-            tmiAPI.getChatterCount(channel.name)
-                .then(chatterCount => twitch.sendChatAdminMessage(`Current Chatters: ${chatterCount.toLocaleString()}`));
-            break;
         case 'followed':
             const currentUser = twitch.getCurrentUser();
             if (!currentUser) break;
@@ -212,11 +204,6 @@ function handleCommands(message) {
                     twitch.sendChatAdminMessage(`Current Uptime: ${secondsToLength(secondsSince)}`);
                 })
                 .catch(() => twitch.sendChatAdminMessage('Could not fetch stream.'));
-            break;
-
-        // misc
-        case 'localunpin':
-            chat.dismissPinnedCheer();
             break;
 
         case 'help':
--- a/src/modules/chat_commands/index.js
+++ b/src/modules/chat_commands/index.js
@@ -5,26 +5,29 @@ const twitchAPI = require('../../utils/twitch-api');
 const chat = require('../chat');
 const anonChat = require('../anon_chat');
 
-const HELP_TEXT = `BetterTTV Chat Commands:
-/b — Shortcut for /ban
-/followed — Tells you for how long you have been following a channel
-/follows — Retrieves the number of followers for the channel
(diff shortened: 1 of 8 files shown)


src/modules/chat_commands/index.js
```diff
@@ -0,0 +1,247 @@
...
+
+                    const secondsSince = Math.round((Date.now() - startedTime.getTime()) / 1000);
+                    twitch.sendChatAdminMessage(`Current Uptime: ${secondsToLength(secondsSince)}`);
+                })
+                .catch(() => twitch.sendChatAdminMessage('Could not fetch stream.'));
+            break;
+
+        // misc
+        case 'localunpin':
+            chat.dismissPinnedCheer();
```
night: we should just remove this command since it's duplicate functionality of the dismiss button on them directly

src/modules/send_message/index.js
```diff
@@ -0,0 +1,74 @@
...
+        this.message = '';
+    }
+}
+
+let twitchSendMessage;
+const methodList = [
+    // msgObj => tabCompletion.onSendMessage(msgObj),
+    msgObj => chatCommands.onSendMessage(msgObj),
+    msgObj => anonChat.onSendMessage(msgObj)
+];
```
night: emojis should be readded too.. it should be pretty much drop-in since it's mutating outgoing message

src/utils/tmi-api.js
```diff
@@ -0,0 +1,47 @@
+
+const ENDPOINT = 'https://tmi.twitch.tv/';
+
+function get(path, options) {
+    return fetch(`${ENDPOINT}${path}`, options);
```
night: you need to make sure fetch is being polyfilled, otherwise just rely on jquery's ajax for now until all the api utils get upgraded

src/utils/tmi-api.js
```diff
@@ -0,0 +1,47 @@
+
```
night: whitespace

src/utils/tmi-api.js
```diff
@@ -0,0 +1,47 @@
...
+const ENDPOINT = 'https://tmi.twitch.tv/';
+
+function get(path, options) {
+    return fetch(`${ENDPOINT}${path}`, options);
+}
+
+function toJson(response) {
```
night: why is this a separate function?

src/utils/tmi-api.js
```diff
@@ -0,0 +1,47 @@
...
+function getHostedChannel(userId) {
+    return get(`/hosts?include_logins=1&host=${userId}`)
+        .then(toJson)
+        .then(data => data && data.hosts && data.hosts[0]);
+}
+
+function getChannelsHosting(userId) {
+    return get(`/hosts?include_logins=1&target=${userId}`)
+        .then(toJson)
+        .then(data => data ? data.hosts : data);
```
night: not sure why you would return data if data is falsey

src/utils/tmi-api.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function getChatterCount(channelName) {
+    if (!channelName) return Promise.resolve(0);
+    return get(`/group/user/${channelName}`)
+        .then(toJson)
+        .then(data => data.chatter_count);
+}
+
+function getChatters(channelName) {
```
night: i dont think we need this endpoint

src/modules/send_message/index.js
```diff
@@ -0,0 +1,74 @@
+const twitch = require('../../utils/twitch');
+const watcher = require('../../watcher');
+const debug = require('../../utils/debug');
+
+const chatCommands = require('../chat_commands');
+const anonChat = require('../anon_chat');
+const socketClient = require('../../socket-client');
+
+const PATCHED_SENTINEL = () => {};
```
night: should swap to a Symbol

src/modules/send_message/index.js
```diff
@@ -0,0 +1,74 @@
...
+class SendMessagePatcher {
+    constructor() {
+        watcher.on('load.chat', () => this.patch());
+    }
+
+    patch() {
+        const client = twitch.getTmiClient();
+        if (!client) return;
+
+        const newTwitchSendMessage = client.sendCommand;
```
night: is `sendCommand` actually messages too? it's odd they'd name it that if it weren't just handling chat room commands.

src/utils/twitch.js
```diff
@@ -200,6 +200,14 @@ module.exports = {
         return controller;
     },
 
+    getTmiClient() {
```
night: getChatServiceClient

src/modules/send_message/index.js
```diff
@@ -0,0 +1,74 @@
...
+const methodList = [
+    // msgObj => tabCompletion.onSendMessage(msgObj),
+    msgObj => chatCommands.onSendMessage(msgObj),
+    msgObj => anonChat.onSendMessage(msgObj)
+];
+
+function bttvSendMessage(channelName, messageToSend) {
+    const channel = twitch.getCurrentChannel();
+    if (channel) {
+        socketClient.broadcastMe(channel.name);
```
night: this will double broadcast on send.. maybe remove the new code i added recently that emits on render (having it before send is actually better)

src/modules/send_message/index.js
```diff
@@ -0,0 +1,74 @@
...
+        try {
+            method(sendState);
+        } catch (e) {
+            debug.log(e);
+        }
+    }
+
+    if (sendState.defaultPrevented) return;
+    messageToSend = sendState.message;
+    twitchSendMessage.call(this, channelName, messageToSend);
```
night: twitch's function can change. to be safe you should always just take in args and pass through args.. also would be smart to check that messageToSend is a string when pulling it out of args.

src/modules/send_message/index.js
DayBr3ak: to answer this, yes, `sendCommand` uses as parameters `channelLogin` and `message` (raw user input).

PR #2834, opened 2017-12-03
[new-twitch] [Feat] Auto hide top cheer and dismiss button

DayBr3ak:
Top cheers can be annoying sometimes, here's a solution for that.

### DIFF
--- a/src/modules/hide_topcheer/index.js
+++ b/src/modules/hide_topcheer/index.js
@@ -1,0 +1,76 @@
+const $ = require('jquery');
+const watcher = require('../../watcher');
+const settings = require('../../settings');
+
+const CARET_BUTTON_SELECTOR = '.pinned-cheer__expand-caret';
+const TOPCHEER_PARENT_SELECTOR = '.chat-room__pane';
+
+const closeButtonTemplate = `
+    <button aria-label="Close" class="tw-button-icon bttv-topcheer-close">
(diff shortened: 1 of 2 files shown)


src/modules/hide_topcheer/index.js
```diff
@@ -0,0 +1,76 @@
+const $ = require('jquery');
+const watcher = require('../../watcher');
+const settings = require('../../settings');
+
+const CARET_BUTTON_SELECTOR = '.pinned-cheer__expand-caret';
+const TOPCHEER_PARENT_SELECTOR = '.chat-room__pane';
+
+const closeButtonTemplate = `
```
night: we already have an open PR readding this

src/modules/hide_topcheer/index.js
```diff
@@ -0,0 +1,76 @@
...
+    </button>
+`;
+
+class HideTopcheerModule {
+    constructor() {
+        settings.add({
+            id: 'hideTopcheer',
+            name: 'Hide Top Cheer',
+            defaultValue: false,
+            description: 'Unclutter your chat, hide the top cheer!'
```
night: top cheers are already hidden by the hide bits option. this isn't a separate option because i dont want to add another setting toggle for something like this. in the future when we redo settings we might break out this encompassing setting into multiple options

PR #2836, opened 2017-12-03
[Feature] Mirror emotes

DayBr3ak:
This PR mirrors emotes when they are preceded by `Flip`. For example: ` Flip PogChamp` or `Flip RareParrot`


DayBr3ak:
Reopening since #2837 got declined.

night:
I don't know why all this code is needed to apply a css transform class on an emote. .. why not just check if emote is proceeded in like a bitwise not (`~`) and call it a day?

DayBr3ak:
I thought about that too. But that would break emotes for non bttv users, is that ok?

PR #2837, opened 2017-12-04
[ Feature ] Mirror emotes & modules/chat/index.js rewrite

DayBr3ak:
This PR hooks up into the the tmi data from the chat service and replace the any bttv emotes it finds into twitch emotes format. Should be more robust, as there is less DOM manipulation.

Added emote mirroring for fun, but let me know if you want me to remove it.

### DIFF
--- a/src/modules/chat/index.js
+++ b/src/modules/chat/index.js
@@ -1,10 +1,10 @@
 const $ = require('jquery');
 const watcher = require('../../watcher');
 const colors = require('../../utils/colors');
+const debug = require('../../utils/debug');
 const twitch = require('../../utils/twitch');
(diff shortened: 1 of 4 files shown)


night:
This feels more unperformant than the prior solution, far more confusing to reason about, and more prone to breaking when Twitch publishes changes. Monkey patching for us is generally only used nondestructively (i.e. to get events out of their code) since historically bad monkey patching with our extension can break the entire site when Twitch updates. Since this is directly mutating records within their stores I'm not comfortable with these changes. For these reasons I'm gonna have to decline this PR, sorry.

DayBr3ak:
i understand, thanks

PR #2856, opened 2017-12-11
[Feature] Tabcompletion & Chat history

DayBr3ak:
This PR adds tab completion and chat history. It's a refactor of the old code.

### DIFF
--- a/src/modules/chat_tab_completion/index.js
+++ b/src/modules/chat_tab_completion/index.js
@@ -7,29 +7,25 @@ const emotes = require('../emotes');
 
 const CHAT_INPUT_SELECTOR = '.chat-input textarea';
 const AUTOCOMPLETE_SUGGESTIONS_SELECTOR = 'div[data-a-target="autocomplete-balloon"]';
+const INPUT_EVENT = new Event('input', {bubbles: true});
 
-function isSuggestionsShowing() {
-    return $(AUTOCOMPLETE_SUGGESTIONS_SELECTOR).length > 0;
-}
-
-const INPUT_EVENT = new Event('input', { bubbles: true });
 function setTextareaValue($inputField, msg) {
     $inputField.val(msg)[0].dispatchEvent(INPUT_EVENT);
 }
 
-function duplicates() {
-    const set = new Set();
-    return emoteName => {
-        if (set.has(emoteName)) {
-            return false;
-        }
-        set.add(emoteName);
-        return true;
-    };
+function normalizedStartsWith(word, prefix) {
+    return word.toLowerCase().startsWith(prefix);
 }
 
 class ChatTabcompletionModule {
     constructor() {
+        settings.add({
+            id: 'tabCompletionTooltip',
+            name: 'Completion Tooltip',
+            defaultValue: true,
+            description: 'Shows a tooltip with suggested names when using @ completion'
+        });
+
         settings.add({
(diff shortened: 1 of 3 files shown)


src/modules/chat_history/index.js
```diff
@@ -0,0 +1,74 @@
...
+const keyCodes = require('../../utils/keycodes');
+
+const CHAT_INPUT_SELECTOR = '.chat-input textarea';
+const AUTOCOMPLETE_SUGGESTIONS_SELECTOR = '[data-a-target=autocomplete-balloon]';
+
+function isSuggestionsShowing() {
+    return $(AUTOCOMPLETE_SUGGESTIONS_SELECTOR).length > 0;
+}
+
+function setTextareaValue($inputField, msg) {
```
DayBr3ak: If you're wondering why I use this, it's because without a 'input' event, the textarea value doesn't stick (like if someone post a message in the chat). This is what's happening currently in the `DoubleClickMentionModule`.

src/modules/chat_history/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+const CHAT_INPUT_SELECTOR = '.chat-input textarea';
+const AUTOCOMPLETE_SUGGESTIONS_SELECTOR = '[data-a-target=autocomplete-balloon]';
+
+function isSuggestionsShowing() {
+    return $(AUTOCOMPLETE_SUGGESTIONS_SELECTOR).length > 0;
+}
+
+const INPUT_EVENT = new Event('input', { bubbles: true });
+function setTextareaValue($inputField, msg) {
```
DayBr3ak: If you're wondering why I use this, it's because without a 'input' event, the textarea value doesn't stay (like if someone post a message in the chat). This is what's happening currently in the DoubleClickMentionModule.

src/modules/chat_tabcompletion/index.js
```diff
@@ -0,0 +1,146 @@
...
+            emoteList.sort();
+        }
+
+        if (includeUsers) {
+            userList = this.getChatMembers().filter(word => word.toLowerCase().indexOf(prefix.toLowerCase()) === 0);
+            userList.sort();
+        }
+
+        if (settings.get('tabCompletionEmotePriority') === true) {
+            return [ ...emoteList, ...userList];
```
night: extra space

src/modules/chat_tabcompletion/index.js
```diff
@@ -0,0 +1,146 @@
...
+        } else {
+            return [...userList, ...emoteList];
+        }
+    }
+
+    onFocus() {
+        this.tabTries = -1;
+    }
+
+    onKeydown(e, includeUsers = true) {
```
night: rather than having two things bound to onKeydown of the textarea and duplicating a lot of code, you should have the chat history in here too.. they are related features, i dont get why you decoupled them when they were coupled for a reason

src/modules/chat_tabcompletion/index.js
```diff
@@ -0,0 +1,146 @@
+const $ = require('jquery');
```
night: this folder name is not correct.

DayBr3ak:
I've done the changes and squashed the branch.

DayBr3ak:
With twitch using a suggestion box when typing `:emote` couldn't we change emojis to something else than `:` ?

PR #2859, opened 2017-12-11
[Feature] Moderator cards

DayBr3ak:
Extend twitch moderator cards when you click an username.


DayBr3ak:
Keybinds are added. Need some testing, I only tested that with my 3 twitch accounts.

DayBr3ak:
rebased and squashed the branch.

Aaron128l:
A few things after my testing.

Things to fix:

- An opaque black background needs to be added to make names readable. [Like this one](https://i.aaron128l.com/xWiS6sG.png)
- Users can also access user cards from the viewerlist which this doesn't add functionality for. (I do understand the difficulty of determining the userstate)
- In my opinion keybinds should be a setting instead enabled by default.

Things to add: 
- There needs to be a way to edit user nicknames. The back-end to render and store them is already there; users just need a way to add and edit them.
- It would be great to have recent chat messages return like in the legacy version.
- Display user information such as follows and views and created date. Maybe even a verified badge if the user is a partner/gmod/admin/staff,etc. 
- It would also be great to make the cards draggable in some way shape or form.

DayBr3ak:
I'm not sure I understand about the opaque background. Did you try with the CSS fix?

Settings for keybinds? Sure.

From the viewerlist, yes it would be great. But userstate is determined through chat badges. And I spent a lot of time trying to find a twitch function to change that, but I couldn't. 

Quick&dirty fix would be to give modcards to every users that is a mod/owner.

Nicknames are doable, need to try.

The rest could be added through further PR, as this is already doing what is intended to do.

DayBr3ak:
Need feedback on my last commit. Is this kind of logic OK? It works pretty well as far as I tested.

Aaron128l:
There is an undocumented TMI endpoint for channel moderators. Not entirely sure if we should use it for this, but it would be better than watching the chat buffer. If all else fails I think it's fine to just add the moderator functions to all users in the viewerlist.

The background CSS can be:
```css
.viewer-card__overlay {
    background: rgba(20, 20, 20, 0.7);
}
```

DayBr3ak:
thanks for the feedback. What about making @mentions clickable too?

DayBr3ak:
current state: 
Image of Yaktocat

DayBr3ak:
current state 

img

Adviceguy:
Hey there, as a mod myself i can't wait too get this back :) I would love too see 8h / permban added aswell  like there was in the old one ( https://gyazo.com/61f620c0a202bf1629511d72790d6826 ) other than that looks great! pce

DayBr3ak:
Permban is already done by twitch by default. 8hrs? could try to squeeze it in.

Adviceguy:
ye that's true. 10min is also there by default , but it's just nice too have it in the card overall :)

DayBr3ak:
new state: 

img

src/modules/chat_moderator_cards/index.js
```diff
@@ -0,0 +1,426 @@
...
+const ACTIONS_MAP = {
+    BAN: '/ban',
+    UNBAN: '/unban',
+    MOD: '/mod',
+    UNMOD: '/unmod',
+    TIMEOUT: '/timeout',
+    PERMIT: '!permit'
+};
+
+const svgCheck = '<figure class="tw-svg"><svg class="tw-svg__asset tw-svg__asset--followcheck tw-svg__asset--inherit" width="16px" height="16px" version="1.1" viewBox="0 0 16 16" x="0px" y="0px"><path clip-rule="evenodd" d="M6.5,12.75L2,8.25l2-2l2.5,2.5l5.5-5.5l2,2L6.5,12.75z" fill-rule="evenodd"></path></svg></figure>';
```
night: business logic of the moderator card itself (all of its rendering/templating) should be in a separate file.

additionally, you should mutate their moderator card html rather than rebuilding it yourself

src/utils/drag-dom-element.js
```diff
@@ -0,0 +1,45 @@
+function dragDomElement(el, selector) {
+    if (!el) return;
```
night: this feels like something we should use jquery.draggable for instead of reinventing the wheel

src/modules/chat_moderator_cards/style.css
```diff
@@ -0,0 +1,49 @@
...
+        -webkit-box-shadow: inset 0 0 6px rgba(0,0,0,.3);
+        background-color: #777;
+    }
+
+    .chat-line__mod-icons {
+        display: none!important;
+    }
+}
+
+[data-a-target=chat-message-mention] {
```
night: this is a bad selector

night:
i can review this more when it's more cleaned up. your index.js is too long, and i'd prefer not to comment on something that really needs refactoring entirely

src/modules/chat_moderator_cards/index.js
DayBr3ak: @night what do you mean by mutating? I'm adding my templates to the "viewer_card" component, not rebuilding it. Unless you mean the part with the moderator buttons?

src/modules/chat_moderator_cards/style.css
DayBr3ak: @night You mean the double quotes missing? Otherwise this selector is used in "doubleclick_mention" module, how is it bad?

src/utils/drag-dom-element.js
DayBr3ak: jquery-ui is not compatible with browserify. see https://jqueryui.com/upgrade-guide/1.12/#official-package-on-npm

DayBr3ak: I'm trying https://www.npmjs.com/package/draggable

DayBr3ak:
it's cleaned up :)

night:
0cb85139c56f645e8ec1e08267c9076de65e7261

Issue #2783, opened 2017-11-19
[Feature Request] auto-default to show all live followed streams on the /directory/following/live page


RTARD:
I've noticed since Twitch changed recently that when I go to my /directory/following/live directory, if there are 50 or more current streamers that you follow live, it now shows the first 49, followed by a box with a "View all" link I have to click to refresh the page and show any more beyond that.

[https://i.imgur.com/F6T5kDd.png](https://i.imgur.com/F6T5kDd.png)

It is getting a bit annoying to me because, of course I want to view all of the streams that are live that I am now following, and I am typically even more interested in seeing who is live down at the bottom of the list... one of my buddies might be live with 1 or 3 or 5 viewers. This is one of those "features" that is just like, what were they even thinking? 

I hope BTTV can make it so the directory pages show everyone you are following by default. I'd think it'd even be safe to just auto-disable it for every single BTTV user without even providing the option, if you didn't want to throw in another toggle switch.

Obviously when I click the /directory/following/live link, I want to see ALL the people I am following that are now live, not just the 49 with the most viewers at the moment. This is literally a feature that, by default, just makes it very hard to notice when your buds are streaming, for no reason.  ty

thinkpad4:
THIS, YES please.

thinkpad4:
This setting Auto Expand Followed Channels List — Automatically clicks the "Load More" option for you doesn't seems to be working.  View all Live Channels is still at the bottom.  Latest Chrome, win7 64 no other extensions.

Bren002:
There is already a setting but I guess it's still broken

night:
I don't think this is a feature most people would use (most people have < 50 followed streams), so this is a wontfix.

PR #2860, opened 2017-12-11
[Feature / fix] Auto expand followed live channels

DayBr3ak:
This PR change a setting (which is actually broken) to permit the user to auto expand his live followed channels #2783


night:
these are two different features

DayBr3ak:
Ok my bad, I'll change it to a new feature then.

DayBr3ak:
Actually having those 2 features is confusing in the settings list. I'll close this.

PR #2864, opened 2017-12-12
[Feature] Chat scrollback

DayBr3ak:
This PR changes the maxsize of the chatController buffer. Users can set their own size. Default is 200 (twitch default).

### DIFF
--- a/src/modules/chat_scrollback_settings/index.js
+++ b/src/modules/chat_scrollback_settings/index.js
@@ -1,45 +1,65 @@
 const watcher = require('../../watcher');
 const storage = require('../../storage');
 const twitch = require('../../utils/twitch');
+const debug = require('../../utils/debug');
 
-const DEFAULT_SIZE = 200;
 const PROMPT_BODY = `Change the number of messages the chat is able to scrollback to. Increasing this value can slow down your browser...
-Leaving this value empty restore default settings.
+
+Leave the field blank to use the default.
 `;
 
-function changeSizeSettings(promptBody, storageID) {
-    const size = prompt(promptBody, storage.get(storageID) || DEFAULT_SIZE);
-    if (size !== null) {
-        storage.set(storageID, size.trim());
+const STORAGE_ID = 'scrollbackAmount';
+let initialSize;
+
+function setChatBufferSize(value) {
+    try {
+        twitch.getChatController().chatBuffer.maxSize = value;
+    } catch (_) {}
+}
+
+function getChatBufferSize() {
+    try {
+        return twitch.getChatController().chatBuffer.maxSize;
+    } catch (_) {}
+}
+
+function checkError() {
+    if (!initialSize) {
+        debug.error('"chatBuffer.maxSize" is undefined, did twitch change the path to the chatBuffer?');
+        return true;
+    }
+    return false;
+}
+
+function promptSizeSettings() {
+    if (checkError()) return;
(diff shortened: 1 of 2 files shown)


src/modules/chat_scrollback_settings/index.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function updateSizeSettings() {
+    const sizeString = storage.get('chatScrollbackSize') || '';
+    let size;
+    if (sizeString.trim().length === 0) {
+        size = DEFAULT_SIZE;
+    } else {
+        try {
+            size = parseInt(sizeString, 10);
```
night: i dont think parseInt can throw

src/modules/chat_scrollback_settings/index.js
```diff
@@ -0,0 +1,47 @@
+const watcher = require('../../watcher');
+const storage = require('../../storage');
+const twitch = require('../../utils/twitch');
+
+const DEFAULT_SIZE = 200;
+const PROMPT_BODY = `Change the number of messages the chat is able to scrollback to. Increasing this value can slow down your browser...
+Leaving this value empty restore default settings.
```
night: this sentence is not proper grammar

src/modules/chat_scrollback_settings/index.js
```diff
@@ -0,0 +1,47 @@
+const watcher = require('../../watcher');
+const storage = require('../../storage');
+const twitch = require('../../utils/twitch');
+
+const DEFAULT_SIZE = 200;
```
night: we should pull this value from twitch, not make our own

src/modules/chat_scrollback_settings/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+function changeSizeSettings(promptBody, storageID) {
+    const size = prompt(promptBody, storage.get(storageID) || DEFAULT_SIZE);
+    if (size !== null) {
+        storage.set(storageID, size.trim());
+    }
+}
+
+function updateSizeSettings() {
+    const sizeString = storage.get('chatScrollbackSize') || '';
```
night: we have a legacy key for this we should use: `scrollbackAmount`

src/modules/chat_scrollback_settings/index.js
```diff
@@ -0,0 +1,47 @@
...
+    let size;
+    if (sizeString.trim().length === 0) {
+        size = DEFAULT_SIZE;
+    } else {
+        try {
```
night: i dont get why the business logic of parsing is here. it should be coerced after the prompt to change, then stored as a number

src/modules/chat_scrollback_settings/index.js
DayBr3ak: True... my bad

src/modules/chat_scrollback_settings/index.js
DayBr3ak: I've changed it in my new commit. If it's still not good, help me please, english isn't my first language :)

night:
So.. given your other pull request.. doesn't this get overwritten when you scroll?

DayBr3ak:
Yes, but scrolling is temporary usually. It reverts back to the settings when the user resumes autoscrolling.

night:
closing due to inactivity

PR #2883, opened 2017-12-21
[Feature] Improve chat freeze

DayBr3ak:
Currently, if the chatBuffer is full, messages will drop even if the chat is "frozen". 

This PR delays messages for ever, as long as the chat is frozen. Messages are queued back when autoscrolling resumes.

### DIFF
--- a/src/modules/chat_freeze/index.js
+++ b/src/modules/chat_freeze/index.js
@@ -1,4 +1,5 @@
 const $ = require('jquery');
+const watcher = require('../../watcher');
 const twitch = require('../../utils/twitch');
 const keycodes = require('../../utils/keycodes');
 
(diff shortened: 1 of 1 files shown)


DayBr3ak:
doesn't work on subs / gift

DayBr3ak:
The new version extend the buffer size.

src/modules/chat_freeze/index.js
```diff
@@ -22,6 +56,7 @@ function shouldFreeze(e) {
 
 class ChatFreezeModule {
     constructor() {
+        let keysPressed = 0;
```
night: why is this moved

src/modules/chat_freeze/index.js
```diff
@@ -1,12 +1,46 @@
...
+}
+
+function patchedSetState({isAutoScrolling}) {
+    if (isAutoScrolling !== undefined) {
+        setChatBufferSize(isAutoScrolling ? bufferSize : Infinity);
+    }
+    return twitchSetState.apply(this, arguments);
+}
+
+function patchScroller() {
```
night: we should be able to do this without monkeypatching, so i'd rather do it without monkeypatching.

DayBr3ak: If you don't monkey patch, and "click" on the `More messages below`, you're stuck with you're infite value, running into memory issues.

DayBr3ak: I guess I could listen to the click on that element.

DayBr3ak: No sometimes you can resume scrolling without clicking on the element. Monkeypatching seems to be the 'safe' way.
Should I drop this PR? @night

night: i dont think you need to drop it.. just needs to be fixed to be safer.. you can detect when scrolling is happening with js events .. not sure why we need to patch to know if user is scrolling or not

night:
closing due to inactivity

PR #2932, opened 2018-01-21
Add "Hide New Viewer Greeting" module

### LINKED ISSUE #2931: Hide "User is new here" feature
Few days ago, Twtich started showing this kind of messages for new people on the channel.
image
When channel is big - it starts to be annoying as hell. 

Could be great, if you could implement option to disable this messages.

Senpos:
Closes #2931 

Created with `hide_friends` module example.

### DIFF
--- a/src/modules/hide_new_viewer_greeting/style.css
+++ b/src/modules/hide_new_viewer_greeting/style.css
@@ -1,6 +1,5 @@
 .bttv-hide-new-viewer-greeting {
-    .chat-line__ritual {
-      display: none;
-    }
+  .chat-line__ritual {
+    display: none;
   }
-  
+}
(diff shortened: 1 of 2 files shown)


Senpos:
Hmm, it seems like Twitch disabled these "Say hello" messages. I am not sure is this permanently or just for some time, but i think we should hold PR for some time, at least for now.

Senpos:
Dang! Twitch brought this feature back!
image

So, patch should be viable now.

src/modules/hide_new_viewer_greeting/style.css
```diff
@@ -0,0 +1,6 @@
+.bttv-hide-new-viewer-greeting {
+    .chat-line__ritual {
```
night: you're mixing 2 space /w 4 space

PR #2933, opened 2018-01-22
Fix "hide-followed-offline" module

### LINKED ISSUE #2853: Offline followed channels no longer hidden
Looks like twitch changed the href on offline followed channels from linking to `/videos/all` to `/videos/[latest vod number]`

### LINKED ISSUE #2916: Hide Offline Followed Channels
Above setting doesn't work, I still see all my offline channels in the left sidebar.

Senpos:
Closes #2916

Some explanations:
- `.side-nav-card[href*="/videos/all"]` **removed**, since i couldn't find a single `.side-nav-card` with `href`
- in `.side-nav-card__link[href*="/videos/all"]`,  `/videos/all` was **replaced** with `/videos/`, since now even if user is offline, he could have `X new video(s)` with `href` containing specific video ID (i.e. `/videos/999999999`), so not only `/videos/all`

Notice, that this functionality **does not** handle `Reruns`
image

I am not even sure if it has to, since technically - stream is ON, but not the live.

### DIFF
--- a/src/modules/hide_sidebar_elements/style.css
+++ b/src/modules/hide_sidebar_elements/style.css
@@ -17,7 +17,7 @@
 }
 
 .bttv-hide-followed-offline {
-  .side-nav-card[href*="/videos/all"], .side-nav-card__link[href*="/videos/all"] {
+  .side-nav-card__link[href*="/videos/"] {
     display: none !important;
   }
 }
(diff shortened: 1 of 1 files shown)


night:
why dont we swap the selector to startswith and call it a day 

```
[href^="/videos/"]
```

Senpos:
@night 
Fixed. `^=` works fine too.

PR #2952, opened 2018-02-03
Added module for highlight users you follow and streamer in the chat

im1x:


src/modules/highlight_follow_streamer_chat/index.js
```diff
@@ -0,0 +1,77 @@
...
+const twitchAPI = require('../../utils/twitch-api');
+
+const STYLE_HIGHLIGHT_FOLLOW = 'bttv-highlight-follow-nick';
+const STYLE_HIGHLIGHT_STREAMER = 'bttv-highlight-streamer-nick';
+
+const followsArray = [];
+let currentChannelName;
+let secondRequestAllowed = true;
+
+function putFollowsInArray(direction, limit) {
```
night: i would not feel comfortable with the amount of requests this generates.. every page load could cause many records being loaded from twitch, and this doesn't even cache that data

src/modules/highlight_follow_streamer_chat/index.js
```diff
@@ -0,0 +1,77 @@
...
+        settings.on('changed.highlightFollowAndStreamerChat', () => this.load());
+        watcher.on('chat.message', ($message, messageObj) => this.onMessage($message, messageObj));
+        watcher.on('vod.message', $message => this.onVodMessage($message));
+        watcher.on('load.chat', () => this.load());
+        watcher.on('load.vod', () => this.load());
+    }
+
+    load() {
+        if (settings.get('highlightFollowAndStreamerChat') === false) return;
+        currentChannelName = twitch.getCurrentChannel().name;
```
night: we should make this feature just followed channels, not streamer too .. we can probably assume that if you care about seeing the streamer in chat you'd follow their channel

im1x: @night Then we can't highlight streamer's nickname with certain CSS. But if you want, I'll do it.

src/modules/highlight_follow_streamer_chat/index.js
im1x: @night I can add a cache. But will this be enough? There will still be many requests.
You need to decide. If the cache helps with this problem, then I'll do it. But if it doesn’t make sense, then I can just close the pull request.

night:
gonna close this since I don't really want that many requests being generated. if twitch ever exposes this data in chat it would be easily doable, but otherwise not the greatest feature for now.

PR #2957, opened 2018-02-07
Fix Hide Offline Followed Channels and Show All Channels

SexualRhinoceros:
Also I made Show All Channels better so it'll unshow all channels when its disabled.

### DIFF
--- a/src/modules/hide_sidebar_elements/index.js
+++ b/src/modules/hide_sidebar_elements/index.js
@@ -55,12 +55,11 @@ class HideSidebarElementsModule {
     toggleAutoExpandChannels() {
         if (settings.get('autoExpandChannels') === true) {
             while ($(".side-nav-show-more-toggle__button:contains('Show more')").length) {
-                $(".side-nav-show-more-toggle__button:contains('Show more')").trigger('click');   
+                $(".side-nav-show-more-toggle__button:contains('Show more')").trigger('click');
             }
-        }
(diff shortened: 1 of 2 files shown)


src/modules/hide_sidebar_elements/index.js
```diff
@@ -54,7 +54,13 @@ class HideSidebarElementsModule {
 
     toggleAutoExpandChannels() {
         if (settings.get('autoExpandChannels') === true) {
-            $('.side-nav-load-more__button').trigger('click');
+            while ($(".side-nav-show-more-toggle__button:contains('Show more')").length) {
```
night: would prefer a `for` loop.. imagine a case where triggering click does not work

SexualRhinoceros: The while loop would cover more cases than a for loop in this situation as it’s unsure how many iterations are needed without expressly querying the number of followed channels from twitch’s api. If only 20 channels, the button only needs to be clicked once but if 200, many more than that. Looping while the button actually exists (what the new whole loop achieves) would solve this issue and provide an easy reverseable solution for those who’ll want to disable and quickly hide everything without having to refresh the page, making this overall (in my opinion) the least janky solution for a janky problem.

night: I get the intention, but you're making a false assumption that triggering a click will always remove the record.. if twitch breaks that functionality you've essentially just made a `while(true)`

PR #2977, opened 2018-02-21
Legacy

jernulf:
option to change to legacy chat without having it as a popout!


night:
twitch is removing this soon

PR #3027, opened 2018-03-09
Replace player extension class

### LINKED ISSUE #3026: Hide Twitch Extensions does not hide streamlabs extension
as title says, the extension's overlay is still there and as far as I know I can't get rid of it.

allanbrault:
fixes #3026, it seems the extension overlay class changed ?


xaxxon:
confirmed that $(".extension-container").remove() gets rid of extension overlay when setting on betterttv doesn't.

Issue #1219, opened 2016-05-29
Search BetterTTV settings


sguergachi:
The settings list grows, and that's awesome. But at some point I think it would be great if we could search for a setting we are looking for instead of reading each one at a time.

EhsanKia:
There's a new settings menu in the works. This is definitely a feature I'd love to see in it. Something similar to how RES on reddit works.

DennisMellegaard:
I've noticed the the branch new-settings hasn't had any commits since  Mar 15. Do you mind if i give it a go. Thinking i will follow up on the react code that is there and continue to build upon that.

Just want to make sure that the new settings page idea isn't dropped or canceled.

AlcaDesign:
Mock up using [Fuse](http://fusejs.io): http://codepen.io/Alca/pen/aJOzQM

night:
we added this a while ago

PR #3049, opened 2018-03-30
Adds search for settings

tankersleyk:
Adds a way to search the BTTV settings - based on the mockup by @AlcaDesign in #1219


night:
thanks for the pr. the implementation is ok, but the design is not. i would drop the h2 line and replace it with the full width search and add a search icon.

PR #3055, opened 2018-04-03
Update chat container selector... again. fixes #3051

### LINKED ISSUE #3051: Chat-relevant selectors are broken (emotes don't work, can't adjust chat font size, etc.)
Since a few hours ago, I can't adjust the font size of Twitch chat with BTTV in Chrome.

The relevant line of code is [here](https://github.com/night/BetterTTV/blob/250e7535b1ff04479973c936766dca2ac33e1256/src/modules/chat_font_settings/index.js#L29). I assume that the changes made [here](https://github.com/night/BetterTTV/commit/250e7535b1ff04479973c936766dca2ac33e1256) will have to be done again, too.

It seems like `div[data-test-selector="chat-room-component-layout"]` doesn't exist anymore, so the CSS rules are not applied. The `div` was changed to a `section` tag.

Full tag:

```html
<section role="complementary" aria-labelledby="chat-room-header-label" data-test-selector="chat-room-component-layout" data-a-target="chat-theme-dark" class="tw-full-width tw-full-height tw-overflow-hidden tw-flex tw-flex-column tw-flex-grow-1 tw-flex-shrink-1">
```

Nareese:


garyking:
Does this work in Twitch chat "Legacy" popout as well? Because I saw some people say that the popout fixed BTTV for them, so this PR might break that.

Nareese:
The Legacy popout chat uses a completely different version of BTTV (v7.0.30) that is specifically designed to work with the old version of the twitch infrastructure that the legacy site uses (Ember framework vs the current React).  Neither the old Ember site, nor the old version of BTTV, are being updated any longer, and this fix will not affect that version.  The legacy chat is likely to disappear completely very soon.

garyking:
Okay thanks for the explanation.

garyking:
Do updates get pushed to BTTV without requiring an update to the Chrome extension? The extension hasn't been updated since November, yet it seems like updates are still being pushed to all users.

Nareese:
Yes.  The extension is literally just a loader that pulls the latest code from the network, so there is rarely ever a need to update the extension itself.  It's < 10 lines of code.  There's very little that can go wrong with it or that changes with any frequency.   It loads whatever the most up to date version is every time you load a twitch page (there is cache busting on the loader).

PR #3069, opened 2018-04-15
update mod icons selector - the second - fixes #3065 (#3067)

### LINKED ISSUE #3065: Mod icons are not properly hidden
Reported on Discord: As a moderator mod icons are shown next to the broadcaster, other mods and yourself.
For example: https://imgur.com/a/vSMbZ

Twitch probly changed the name of the element: https://github.com/night/BetterTTV/blob/e1e9284a71d637c11e2f2f4ab59b3b0a2bb1533b/src/modules/chat/index.js#L227

Smatify:
Yesterday @Nareese commited a fix for #3065 (#3067), but it seems that the selector is `.mod-icon` instead of `.mod_icon`.

Also this will remove the mentioned empty div

### DIFF
--- a/src/modules/chat/index.js
+++ b/src/modules/chat/index.js
@@ -224,7 +224,7 @@ class ChatModule {
             $element.hide();
         }
 
-        const $modIcons = $element.find('.mod_icon');
+        const $modIcons = $element.find('.mod-icon').parent();
         if ($modIcons.length) {
             const userIsOwner = twitch.getUserIsOwnerFromTagsBadges(user.badges);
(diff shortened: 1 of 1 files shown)


src/modules/chat/index.js
```diff
@@ -224,7 +224,7 @@ class ChatModule {
             $element.hide();
         }
 
-        const $modIcons = $element.find('.mod_icon');
+        const $modIcons = $element.find('.mod-icon').parent();
```
night: this is dangerous, so i cannot commit this. if parent is ever the message itself we could accidentally remove the entire message.

PR #3072, opened 2018-04-16
Hide mouse fullscreen

### LINKED ISSUE #3008: Feature: Make the cursor disappear in full screen mode
I don't know why twitch is not making this cursor disappear so if you guys could make this feature in Better TTV would be great to see it !!!

mroseman95:
Sets up a timer that will hide the mouse while in fullscreen mode, and resets that timer after every mouse move.

closes #3008

### DIFF
--- a/src/modules/video_player/index.js
+++ b/src/modules/video_player/index.js
@@ -58,10 +58,30 @@ function handlePlayerClick() {
     }, 250);
 }
 
+// function handleFullScreen() {
+//     let timer;
+//
+//     timer = setInterval(() => {
+//         $('.video-player--fullscreen')
+//             .toggleClass('bttv-hide-cursor-fullscreen', true);
+//     }, 3000);
+//
+//     $('body').mousemove(() => {
+//         $('.video-player--fullscreen')
+//             .toggleClass('bttv-hide-cursor-fullscreen', false);
+//         clearInterval(timer);
+//         timer = setInterval(() => {
+//             $('.video-player--fullscreen')
+//                 .toggleClass('bttv-hide-cursor-fullscreen', true);
+//         }, 3000);
+//     });
+// }
+
 class VideoPlayerModule {
     constructor() {
         this.keybinds();
         watcher.on('load.player', () => this.clickToPause());
+        watcher.on('load.player', () => this.hideMouseFullscreen());
         settings.add({
             id: 'hidePlayerExtensions',
             name: 'Hide Twitch Extensions',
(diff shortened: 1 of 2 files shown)


mroseman95:
The code I wrote will only hide the mouse cursor when the player is fullscreen, but could easily be changed to hide the mouse whenever it is inactive over the player. Say in theater mode.

I think that option is useful, but decided to just stick to fullscreen, unless others agree that it would make sense to always hide an inactive mouse cursor over the player.

src/modules/video_player/index.js
```diff
@@ -94,6 +104,36 @@ class VideoPlayerModule {
             $(VIDEO_PLAYER_SELECTOR).on('click', '.player-overlay.player-fullscreen-overlay', handlePlayerClick);
         }
     }
+
+    hideMouseFullscreen() {
+        const showMouse = () => {
```
night: these functions don't need to be created here. we should move them outside of this class

src/modules/video_player/index.js
```diff
@@ -94,6 +104,36 @@ class VideoPlayerModule {
...
+        showMouse();
+
+        if (settings.get('hideMouseFullscreen') === true) {
+            hideMouseTimer = setInterval(() => {
+                hideMouse();
+            }, 3000);
+
+            $('body').on('mousemove', '.video-player--fullscreen', () => {
+                showMouse();
+                clearTimeout(hideMouseTimer);
```
night: we should not create a timer that runs to hide the mouse, that is just wrong. you can easily use a debounce function that gets called on mouse move that hides the mouse after a timeout.

src/modules/video_player/index.js
```diff
@@ -94,6 +104,36 @@ class VideoPlayerModule {
...
+            $('.video-player--fullscreen')
+                .toggleClass('bttv-hide-cursor-fullscreen', false);
+        };
+
+        const hideMouse = () => {
+            $('.video-player--fullscreen')
+                .toggleClass('bttv-hide-cursor-fullscreen', true);
+        };
+
+        $('body').off('mousemove', '.video-player--fullscreen');
```
night: create a proper function that adds and removes the listener on page load and setting change.. you're creating a listener on body, which does not need to be recreated on every player load..

src/modules/video_player/index.js
mroseman95: I did try that, but an issue arises if a user goes fullscreen without moving their mouse. If they use a hotkey for instance. Then they would have to move their mouse initially for the timer to start.

One solution to this I was thinking of, is adding an event in watcher.js that fires when the screen goes fullscreen, then in the video_player module starting the timer off that.

night: i think requiring mouse movement is an acceptable constraint to hiding the mouse. it's not worth adding additional plumbing for an edge case benefitting only a few people.

mroseman95: Alright, then that should simplify the code a bit.

night:
solved in a32407f1f2d9cb6d4bb0ab5b4e58321a8bcc5727

PR #3079, opened 2018-04-24
Feat mute invisible tabs

### LINKED ISSUE #3059: [Feature Request] Keep all non-focused tabs/streams muted
I should start by saying that I don't know whether this is technically possible. Anyway...

Even though I always have a bunch of streams open, I've found myself only using the audio from one stream at a time, 99% of the time.

It would be great if you could activate a mode in which all streams but the currently focused one are muted. When switching to a new tab with this setting on, the old one would be muted and the new (now focused) one would be unmuted. That way, you could forget about sound management (i.e. spamming the `m` hotkey) all together when frequently switching between stream tabs.

mroseman95:
This code adds a setting to mute any streams that aren't in a visible tab. 

A tab is visible if it is the currently selected tab of a window, and the window is not minimized. The documentation for `visibilitychanged` event is [here](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)

closes #3059

### DIFF
--- a/src/modules/video_player/index.js
+++ b/src/modules/video_player/index.js
@@ -6,6 +6,18 @@ const twitch = require('../../utils/twitch');
 
 const VIDEO_PLAYER_SELECTOR = '.video-player .player';
 
+let hidden, visibilityChange;
+if (typeof document.hidden !== 'undefined') { // Opera 12.10 and Firefox 18 and later support
+    hidden = 'hidden';
+    visibilityChange = 'visibilitychange';
+} else if (typeof document.msHidden !== 'undefined') {
+    hidden = 'msHidden';
+    visibilityChange = 'msvisibilitychange';
+} else if (typeof document.webkitHidden !== 'undefined') {
+    hidden = 'webkitHidden';
+    visibilityChange = 'webkitvisibilitychange';
+}
+
 function stepPlaybackSpeed(faster) {
     const currentPlayer = twitch.getCurrentPlayer();
     if (!currentPlayer || !currentPlayer.props.vodID) return;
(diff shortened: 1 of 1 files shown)


src/modules/video_player/index.js
```diff
@@ -6,6 +6,18 @@ const twitch = require('../../utils/twitch');
 
 const VIDEO_PLAYER_SELECTOR = '.video-player .player';
 
+let hidden, visibilityChange;
```
night: none of this is necessary. `document.hidden` and `visibilitychange` are sufficient. we don't care about old browsers.

src/modules/video_player/index.js
```diff
@@ -94,6 +122,14 @@ class VideoPlayerModule {
             $(VIDEO_PLAYER_SELECTOR).on('click', '.player-overlay.player-fullscreen-overlay', handlePlayerClick);
         }
     }
+
+    muteInvisibleTabs() {
+        $(document).off(visibilityChange, handleVisibilityChange);
```
night: this would be a lot more trivial to just use jQuery's blur and focus events..

```
$(window).on('blur', () => console.log('the page is not active'));
$(window).on('focus', () => console.log('the page is active'));
```

src/modules/video_player/index.js
```diff
@@ -58,6 +70,14 @@ function handlePlayerClick() {
     }, 250);
 }
 
+function handleVisibilityChange() {
+    if (document[hidden]) {
+        $('.player-button--volume:has(.mute-button)').click();
```
night: this feature should not override a user-selected setting. if a user manually mutes and alt-tabs, and alt-tabs back.. this would unmute the player.. which is bad.

night:

PR #3086, opened 2018-04-29
hide moderator card when a username is doubleclicked

mroseman95:
When you double click a username in chat in order to mention them, the moderator card still pops up.

I added a line to close that moderator card only on double click.

Signed-off-by: Matthew Roseman <mroseman95@gmail.com>


src/modules/doubleclick_mention/index.js
```diff
@@ -23,6 +23,9 @@ class DoubleClickMentionModule {
     load() {
         $(CHAT_ROOM_SELECTOR).off('dblclick.mention').on('dblclick.mention', USERNAME_SELECTORS, e => {
             if (e.shiftKey || e.ctrlKey) return;
+
+            $('.viewer-card__hide').find('button').click();
```
night: i believe this is a race since there's no guarantee the viewer card hide button will be loaded by the time this callback occurs.

Issue #3205, opened 2018-08-17
Highlights aren't working anymore


Rbcca:
Out of the blue my highlights weren't working anymore since yesterday. I checked my highlights and blacklist settings, deleted my cache and cookies, reinstalled BTTV countless times, disabled all other extensions on Firefox and even switched to Chrome and it isn't working.

Nareese:
duplicate of #3204

Issue #3203, opened 2018-08-17
Highlighted message with mention bugged


CauzYolo:
Hi all,
I encounter a strange bug with the mentions. Since today the mentionned message is not highlited on many cases but on some case it is. I've searched on google if someone had this bug but it wasn't succesfull.
There is some screenshoots which i hope may help u.

![bug_bttv_1](https://user-images.githubusercontent.com/10593700/44269723-b5b05a00-a235-11e8-9478-eb5e13623758.png)

![bug_bttv_2](https://user-images.githubusercontent.com/10593700/44269724-b5b05a00-a235-11e8-89f4-1e6c52b7ad9b.png)

And there my BTTV Settings:

`{"bttv_gwEmotes":true,"bttv_tabCompletionTooltip":true,"bttv_darkenedMode":true,"bttv_disableVodRecommendationAutoplay":true,"bttv_importedV6":true,"bttv_highlightKeywords":"cauzyolo_","bttv_splitChat":false,"bttv_pinnedHighlights":true,"bttv_showDeletedMessages":true,"bttv_disableLocalizedNames":false,"bttv_ffzEmotes":true,"bttv_timeoutHighlights":false,"bttv_bttvGIFEmotes":true,"bttv_clickTwitchEmotes":true,"bttv_blacklistKeywords":"","bttv_hideFeaturedChannels":false,"bttv_leftSideChat":false,"bttv_chatImagePreview":true,"bttv_disableUsernameColors":false,"bttv_bttvEmotes":true,"bttv_disableFPVideo":true}`

I hope this issue is not a duplicate.

PR #3207, opened 2018-08-18
fixes #3204 (#3203, #3205)

### LINKED ISSUE #3204: Pin Highlighted Messages bugged/issue
I was receiving Pinned highlighted messages just fine until recently it has stopped working. I've tried updating and clearing cache, turning the setting on and off makes no difference. Any reason for this?

Digikid13:
fixes #3204 (#3203, #3205)

This is a fix for message highlighting, seems the message types have been updated. I tested Mentions, Links, Emotes, and self mentions and all seem to work again. I'm not sure what `CurrentUserHighlight` is, I was thinking it was self mentions, but those are type `3`.

I'm assuming these changes were created because something was added above them. Not sure if `CurrentUserHighlight` is still 1 or if that was also pushed down.

### DIFF
--- a/src/modules/chat_highlight_blacklist_keywords/index.js
+++ b/src/modules/chat_highlight_blacklist_keywords/index.js
@@ -147,11 +147,11 @@ function messageTextFromAST(ast) {
                 return node.content.trim();
             case 1: // CurrentUserHighlight
                 return node.content;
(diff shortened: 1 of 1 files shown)


Digikid13:
self_highlight

PR #3321, opened 2018-12-07
Disable automatic theatre mode when not online/hosting

Technoblazed:
This PR adds a new setting to allow the user to disable automatic theatre mode when a channel is offline/not hosting a user.

Each combination has been tested with the two theatre mode settings.

### DIFF
--- a/src/modules/auto_theater_mode/index.js
+++ b/src/modules/auto_theater_mode/index.js
@@ -10,6 +10,12 @@ class AutoTheaterModeModule {
             defaultValue: false,
             description: 'Automatically enables theatre mode'
         });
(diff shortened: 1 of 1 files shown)


src/modules/auto_theater_mode/index.js
```diff
@@ -19,6 +25,9 @@ class AutoTheaterModeModule {
         const player = twitch.getCurrentPlayer();
         if (!player) return;
 
+        const isLive = !!player.player.ended;
```
night: are you sure this value is accurate? iirc you need to check the player state

night:
closing due to lack of progress

PR #3322, opened 2018-12-08
Added feature to filter followed live channels by game.

xTerminatioNx:
If following a lot of channels that are streaming a variety of games the list of followed channels that are live gets quite cluttered. That can make it hard to discern which channels are streaming a specific game at that time. This remedies that by allowing filtering by game.

This feature has currently only been implemented on the live "Channels" tab for simplicity.

### DIFF
--- a/src/modules/directory_filter_live/filter.css
+++ b/src/modules/directory_filter_live/filter.css
@@ -1,0 +1,12 @@
+#bttvLiveFilterPanel {
+  float: right;
+}
+
+#bttvLiveFilterPanel label {
+  font-size: 1.6rem;
+  padding-right: 10px;
+}
+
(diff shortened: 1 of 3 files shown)


night:
thanks for submitting a PR, but I don't want to accept this PR. arbitrarily filtering like this generates weird results. unless the Twitch api can support filtering I'd rather not

xTerminatioNx:
> thanks for submitting a PR, but I don't want to accept this PR. arbitrarily filtering like this generates weird results. unless the Twitch api can support filtering I'd rather not

@night Can you explain what you mean by arbitrary and why you think its generating weird results? Seems to work predictably and accomplish the goal. I don't understand what the issue is. Happy to rectify if you explain.

PR #3346, opened 2019-01-06
[Feature] Add highlighted messages sound notification

### LINKED ISSUE #3338: [Feature Request] Audio notification for chat mention
Add an option to get notified by sound (ding, beep etc.) when mentioned on chat (tagged). Makes it easier to see tags as you don't have to stare at chat window.

whitecrownclown:
- Added sound notification for chat mention
- Added setting for this (default false)

Fixes #3338


whitecrownclown:
@night Hello, would you kindly have a look? I've also added the mp3 file (audio notification, can be changed of course) and i assume it needs to be manually uploaded to the cdn (not sure about this). Thanks!

whitecrownclown:
@night Yes, it works, i've tested it on both firefox and chrome. I've also moved this into the highlight module.

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -184,6 +186,55 @@ class ChatHighlightBlacklistKeywordsModule {
...
+
+        settings.add({
+            id: 'soundHighlight',
+            name: 'Play Sound On Highlight',
+            defaultValue: false,
+            description: 'On highlight you hear a sound'
+        });
+
+        this.username = twitch.getChatServiceClient().configuration.username.toLowerCase();
+        this.soundHighlight = false;
```
night: why are we storing this as a property of the class? why not just check the settings?

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -184,6 +186,55 @@ class ChatHighlightBlacklistKeywordsModule {
...
+        this.setSoundHighlight(settings.get('soundHighlight'));
+        settings.on('changed.soundHighlight', this.setSoundHighlight);
+    }
+
+    setSoundHighlight(value) {
+        this.soundHighlight = value;
+    }
+
+    handleSoundHighlight(messageParts = []) {
+        const isMentioned = messageParts.some(({content}) => {
```
night: why don't we just play the sound from at line 271 and skip all this "mentioned" logic?

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -184,6 +186,55 @@ class ChatHighlightBlacklistKeywordsModule {
...
+        if (!isMentioned) {
+            return;
+        }
+
+        let node = $(`#${CHAT_MENTION_NOTIFICATION_ID}`);
+
+        if (node.length) {
+            node[0].play();
+        } else {
+            node = document.createElement('audio');
```
night: why do we need to make an audio element? can't we just load it in js and play it in js?

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -184,6 +186,55 @@ class ChatHighlightBlacklistKeywordsModule {
...
+            node.hidden = 'hidden';
+            node.id = CHAT_MENTION_NOTIFICATION_ID;
+            node.autoplay = true;
+            node.src = cdn.url('assets/notification.mp3');
```
night: we have a `tink` sound someone in git history that we should use instead

src/modules/chat_highlight_blacklist_keywords/index.js
whitecrownclown: Updated.

src/modules/chat_highlight_blacklist_keywords/index.js
whitecrownclown: So scrap the mentioned logic and add this for highlights instead?

whitecrownclown:
@night 	A summary of all the changes:

- Changed notification sound to `tink` (got it by searching through git history);
- Removed `mentioned` logic in favor of playing the notification for `highlighted` messages.

PR #3348, opened 2019-01-06
[Feature] Mute inactive tabs

### LINKED ISSUE #3059: [Feature Request] Keep all non-focused tabs/streams muted

whitecrownclown:
- Add setting to toggle mute inactive tabs

Fixes #3059


src/modules/video_player/index.js
```diff
@@ -19,6 +19,30 @@ function stepPlaybackSpeed(faster) {
     currentPlayer.player.setPlaybackRate(rates[idx]);
 }
 
+const handleVisibilityChange = (function() {
```
night: this closure shouldn't really work as written.. there's no guarantee that the player is loaded when this executes

src/modules/video_player/index.js
```diff
@@ -19,6 +19,30 @@ function stepPlaybackSpeed(faster) {
...
+    let userMuted = $('.player-button--volume .mute-button').length;
+
+    const clickHandler = () => {
+        userMuted = $('.player-button--volume .mute-button').length;
+    };
+
+    $('.player-button--volume').on('click', clickHandler);
```
night: this doesn't seem to consider the player mutating mute, so it could get inconsistent with player state .. we should rely instead on the player's mute state instead of maintaining our own

src/modules/video_player/index.js
```diff
@@ -19,6 +19,30 @@ function stepPlaybackSpeed(faster) {
...
+
+    return () => {
+        const button = $('.player-button--volume');
+
+        button.off('click', clickHandler);
+
+        if (document.hidden) {
+            $('.player-button--volume .mute-button').trigger('click');
+        } else if (!document.hidden && !userMuted) {
+            $('.player-button--volume .unmute-button').trigger('click');
```
night: should we be using the player api directly instead of triggering clicks?

whitecrownclown: @night I've updated the PR, i kept the page visibility api. Things to note:

* the document that is shown triggers visibility before the one that is hidden (that's why there's a settimeout in there, better ideea is more than welcome)
* could've used onBlur/onFocus, but the drawback is that you cannot watch a stream with sound on and do something on another monitor at the same time since you'll trigger blur
* if user mutes the player it will take it into account when switching tabs, `onbeforeunload` handles a case where `visibilitychange` triggers on refresh and breaks it

whitecrownclown:
@night I've updated the PR, i kept the page visibility api. Things to note:

* the document that is shown triggers visibility before the one that is hidden (that's why there's a settimeout in there, better ideea is more than welcome)
* could've used onBlur/onFocus, but the drawback is that you cannot watch a stream with sound on and do something on another monitor at the same time since you'll trigger blur
* if user mutes the player it will take it into account when switching tabs, `onbeforeunload` handles a case where `visibilitychange` triggers on refresh and breaks it

src/modules/video_player/index.js
whitecrownclown: Done

src/modules/video_player/index.js
whitecrownclown: This was rewritten

src/modules/video_player/index.js
whitecrownclown: This was rewritten

src/modules/video_player/index.js
```diff
@@ -128,6 +148,44 @@ class VideoPlayerModule {
...
+        if (document.hidden) {
+            this.onWindowBlur();
+        } else {
+            this.onWindowFocus();
+        }
+    }
+
+    muteInvisibleTabs() {
+        $(document).off('visibilitychange', this.onVisibilityChange);
+        $(window).off('blur', this.onWindowBlur);
```
night: why do we support both visibility and blur? window could be blurred but still visible, right?

src/modules/video_player/index.js
```diff
@@ -80,6 +81,18 @@ class VideoPlayerModule {
...
             this.clickToPause();
             watchPlayerRecommendationVodsAutoplay();
+
+            const currentPlayer = twitch.getCurrentPlayer();
+            this.isMutedByUser =
+                currentPlayer &&
+                currentPlayer.player &&
+                currentPlayer.player.getMuted();
+
+            this.onVisibilityChange = this.onVisibilityChange.bind(this);
```
night: rather than bind to `this` you generally want to define the method as this instead:

```

onWindowBlur = () => ...
```

in reality you shouldn't really need this though.. you should define handle functions outside of this class

src/modules/video_player/index.js
```diff
@@ -80,6 +81,18 @@ class VideoPlayerModule {
         watcher.on('load.player', () => {
             this.clickToPause();
             watchPlayerRecommendationVodsAutoplay();
+
+            const currentPlayer = twitch.getCurrentPlayer();
+            this.isMutedByUser =
```
night: we probably want to store the value of muted on blur/invisible .. then restore to that value on focus/visible

src/modules/video_player/index.js
whitecrownclown: Yeah, I've updated that, no reason to be class methods.

src/modules/video_player/index.js
whitecrownclown: The original issue was with the logic of getting the player mute state, but this no longer applies. I've switched to `visibilitychange` only.

src/modules/video_player/index.js
```diff
@@ -78,12 +78,58 @@ function togglePlayerCursor(hide) {
...
+
+function onVisibilityChange() {
+    if (document.hidden) {
+        onWindowBlur();
+    } else {
+        onWindowFocus();
+    }
+}
+
+function setUnloading() {
```
whitecrownclown: This is needed because `visibilitychange` triggers when unloading/refreshing the page. It mutes the stream -> twitch updates localstorage and then the stream will always start muted.

Issue #3363, opened 2019-01-31
Click to Play/Pause Stream


erenolgac:
There already is an issue about this but this feature still doesn't work

Nareese:
duplicate of #3354

PR #3367, opened 2019-02-03
Fix for click to pause issue

itsazhuhere:
#3354
#3363

> * Changed/added to the CSS selector.
> * Fixed a subsequent issue with clicking the center play button when paused.


shinncchen:
I understand the issue was closed but "Click to Play/Pause Stream" works only when the streamer doesn't have any plugin over the player (Amazon Prime, Tobii Ghost or any other game plugin).

Chrome 76.0.3809.87 (Official Build) (64-bit) (Win10)
BetterTTV 7.2.1

Issue #3359, opened 2019-01-21
Split Chat doesnt work as intended if blacklisted messages are deleted


tman610:
e.g normally you get dark, light, dark. If light message is removed due to blacklisted phrase, you see dark, dark.

PR #3373, opened 2019-02-11
[ONHOLD] Refactor split chat

whitecrownclown:
Hello @night ,

This PR resolves : #3359

* instead of toggling a class on each `message_line`, we're toggling one on the body
* `:nth-child(2n)` will alternate the background, this also works with messages that get removed, although the background of the messages that come after will change to the other (don't think that's an issue)

Thanks!


whitecrownclown:
@night You're right. Twitch keeps about 150 messages so the background will change. I'd like to keep this PR open/onhold, though i'm not sure there's a css only way to do this.

night:
you can reopen this when there's progress

PR #3376, opened 2019-02-14
[Feature] Add settings for hiding reruns in directory/following

### LINKED ISSUE #3368: [Feature Request] Hide "Rerun" VoD's from /directory/following
Would it be possible to have an option to block "Rerun" VoD's on Twitch's Following directory? They're distracting and I always think there are more channels streaming than there really is.

whitecrownclown:
@night This PR adds a toggle for hiding/showing reruns in directory/following Fixes #3368


cmccall:
FYI I am working on something a bit similar. 

This seems to only work on the following page, not the live directory page. I have the guts for getting it working on that live page in my fork. Some other things I am doing is also filtering on the title of the stream (as most streamers don't use the actual rerun feature, they just put [RERUN] or something at the start of the stream title I've noticed), and also had begun work on moving Rerun streams into their own section rather than hiding them (though haven't completed that). If you'd like to use some of those ideas in this PR, this commit has all my work:

https://github.com/cmccall/BetterTTV/commit/6e0dd40d234758d37e64919d9a7da99d0b887a57

whitecrownclown:
@cmccall Hope you don't mind,i can pickup from your work and merge it with my PR for a fully functioning "disable reruns" functionality. Thank you for your work!

cmccall:
@whitecrownclown I don't mind at all, thanks for your work as well!

whitecrownclown:
As a summary of the changes:

- Option to toggle reruns into their own category on /directory/following or /directory/following/live
- Added Mutation observer for /directory/following (when toggling between Overview/Hosts/etc we need to watch for addedNodes on the channel wrapper and then move the reruns into their own category). The observer is on the container itself and we don't need to monitor subtree, only childList (shouldn't be too costly).

Feedback is welcome @night , @cmccall .

Also, thanks to @cmccall for his contribution on this!

night:
I think it's worth holding out for https://caniuse.com/#feat=css-has on complex features like this, especially since reruns aren't a major problem for most users

whitecrownclown:
It does seem a bit far away, but if not a lot of users request it, I think waiting is ok.

night:
@whitecrownclown i refactored watcher to use a much smarter real-time selector implementation, so it should be easy to now target rerun cards being added to the directory page. can you refactor this PR to use it.. perhaps we can finally get this in

night:
closing this due to inactivity, but feel free to re-open if there's been any progress.

PR #3378, opened 2019-02-17
[Feature] Add libhrc for compressing chat messagess

whitecrownclown:
@night This PR follows up on #1922, however it's not handling emotes right now. I can update the PR once i get emotes working with libhrc.


night:
this implementation will not work as intended. due to how react creates an AST and how we're parsing that AST this will run compression on many subtree elements, not the entire message

PR #3379, opened 2019-02-17
[Feature] Scrolling Over Video to Change Volume

### LINKED ISSUE #792: Scrolling Over Video to Change Volume
The ability to adjust the volume of the video player when scrolling with a mouse wheel over the player would be useful.

whitecrownclown:
Fixes #792


src/modules/video_player/index.js
```diff
@@ -103,10 +103,43 @@ class VideoPlayerModule {
             defaultValue: false,
             description: 'Disables autoplay of recommended videos on VoDs'
         });
+        settings.add({
```
night: remove this setting. we don't need more settings.

whitecrownclown: Removed. It's enabled by default now.

PR #3388, opened 2019-02-23
[Feature] Add delete message option for moderator contextmenu

### LINKED ISSUE #3340: [Feature Request] Delete Message
Twitch recently added a new Moderator feature "Delete Message" which only deletes one message.

I really dislike having the Mod Buttons to the left of every message in chat. 
So would it be possible to get this feature implemented in the right-click menu that BTTV adds for Mods?

whitecrownclown:
@night 

Fixes #3340 

This PR adds the possibility to delete a single message using the moderator context menu.


night:
you're gonna have to fix up that svg, it looks terrible compared to the original

whitecrownclown:
I've updated the svg.

Screenshot 2019-03-21 at 11 44 20

AlexWayfer:
It would be very useful!

PR #3394, opened 2019-02-27
[Feature] Add auto leave raid module

### LINKED ISSUE #2936: Suggestion: Feature to opt-out of auto joining a raid
Raids were changed so that the viewer is now required to click to opt-out of the raid.

This can be a problem, if for example, the stream is paused then the viewer will still get included in the raid and the new stream will start playing. Or maybe you have the channel up but you missed the raid notification and you simply don't want to have your browser window forcibly redirected.

It would be nice to have some features to control this behavior. Some ideas:

  * If the stream is paused, then don't join the raid
  * Global BTTV setting to control whether the user opts-in or opts-out of raids

Thanks!

whitecrownclown:
@night Fixes #2936


src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+        const autoLeaveRaidWatcher = new window.MutationObserver(mutations => {
+            if (!settings.get('autoLeaveRaid')) {
+                return;
+            }
+
+            mutations.forEach(mutation => {
+                for (const el of mutation.addedNodes) {
+                    if (el.getAttribute('data-test-selector') === 'raid-banner') {
+                        $(el).find('button').click();
```
night: probably should be more specific than clicking an arbitrary button, no?

whitecrownclown:
> is there a better way than mutation observer? can we instead catch this at a higher level inside of react (like we do for host mode)?

I think so, I'll have a look over it and update the PR.

whitecrownclown:
@night 	Couldn't do the same as `host mode` since the raid notifications are added from a WebSocket message (so the react context isn't available until the notification arrives). I've changed from "button click" (which wasn't reliable as it would also cancel the raid from the broadcaster) to calling `handleLeaveRaid` when the mutation triggers.

In the end, I couldn't really drop the mutation observer since I couldn't find any method to patch, as done in host mode.

redwyre:
Any update on this?

whitecrownclown:
@redwyre Waiting on a review atm

night:
@whitecrownclown similarly for this one.. this should be like 4 lines of observer code using the new system!

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+const CHAT_ROOM_NOTIFICATIONS_SELECTOR = '.chat-room__notifications';
+
+class AutoLeaveRaid {
+    constructor() {
+        settings.add({
+            id: 'autoLeaveRaid',
+            name: 'Automatically Leave Raids',
+            defaultValue: false,
+            description: 'Automatically leaves raids'
```
night, line 15: nit: lets make this description be a little different than the name?

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,53 @@
...
+        });
+        watcher.on('load.chat', () => {
+            this.autoLeaveRaidObserver();
```
night, line 18: nit: we don't need `{}` on this

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+    autoLeaveRaidObserver() {
+        const observe = (_watcher, element) => {
+            if (!element) return;
+            if (_watcher) _watcher.disconnect();
+            _watcher.observe(element, {childList: true, subtree: true});
+        };
+
+        const autoLeaveRaidWatcher = new window.MutationObserver(mutations => {
+            if (!settings.get('autoLeaveRaid')) {
```
night, line 30: nit: we don't need `{}` on this

night:
@whitecrownclown looks like this PR is in a bit of a mangled state.. you may need to merge master or rebase and exclude everything but your expected commits

whitecrownclown:
@night Yeah, I'm not sure what happened there, I did rebase, perhaps I messed it up somehow. I'll update the PR, there's still an issue I need to solve with the functionality.

night:
@whitecrownclown because safe emitter wraps the callback it is not possible for `off` to work in your code because the reference does not match. if you console log the output you will see the event handler does not get removed if you call off. the closure returned from `on` should work. if you can provide more information on the misbehavior you see i'm happy to take a look.

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,55 @@
...
+            debug.log("Couldn't find raidContext - autoLeaveRaid module");
+        }
+    }
+}
+
+class AutoLeaveRaid {
+    constructor() {
+        settings.add({
+            id: 'autoLeaveRaid',
+            name: 'Leave Raids Automatically',
```
night, line 31: this naming is inconsistent with the other automatic setting we're adding.. should this be `Auto-leave Raids`?

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,55 @@
...
+        } else {
+            debug.log("Couldn't find raidContext - autoLeaveRaid module");
+        }
+    }
+}
+
+class AutoLeaveRaid {
+    constructor() {
+        settings.add({
+            id: 'autoLeaveRaid',
```
night, line 30: `autoLeaveRaids`

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,55 @@
...
+            name: 'Leave Raids Automatically',
+            defaultValue: false,
+            description: 'Disable auto-joining a raid',
```
night, line 33: `Automatically removes you from future channel raids`

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,55 @@
...
+const COMMUNITY_HIGHLIGHT_SELECTOR = '.community-highlight-stack__card';
+
+let autoLeaveRaidListener;
+
+function handleLeaveRaid(node, isConnected) {
+    if (!isConnected) return;
+    const $node = $(node);
+    const raidBanner = $node.find('[data-test-selector="raid-banner"]');
+    if (raidBanner.length > 0) {
+        const raidContext = twitch.getChannelRaidContext(raidBanner.get(0));
```
night, line 17: do we actually need this raid context? can we detect this from the button state?

src/utils/twitch.js
```diff
@@ -398,6 +398,36 @@ module.exports = {
         return currentUser.id === currentChannel.id;
     },
 
+    getChannelHostingContext() {
```
night, line 401: rm

src/modules/auto_leave_raid/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+        if (raidContext && raidContext.hasJoined) {
+            raidBanner.find('button').click();
+        } else {
+            debug.log("Couldn't find raidContext - autoLeaveRaid module");
```
night, line 22: did you mean to leave this debug line in?

night (changes requested):

ExploreTheFloor:
What happened to this PR?

night:
closing due to inactivity, and superseded by #4683

Issue #3355, opened 2019-01-16
Hide Recommended Channels is not working


ian9k:
I don't use any other extensions. I tried clean chrome install. 

![bttv](https://user-images.githubusercontent.com/46759514/51271763-ab52e180-19c8-11e9-850c-5f5872e56294.JPG)

Sintamatix:
I'm having the same issue

whitecrownclown:
@ian9k @Sintamatix I cannot seem to reproduce this, do you still have this issue?

ian9k:
Yes.

whitecrownclown:
@ian9k Could you provide more details? Like the browser you're using, if you tried it without other extensions, etc? Thank you!

ian9k:
I use chrome, i did fresh install with no other extensions and the problem is still there. I also tried firefox same problem.

ufos:
Same here. I'm using newest Firefox 65.0.1 @ Windows 10 1809 and this feature doesn't work at all.

whitecrownclown:
@Reqruiz @ian9k Are you using the latest bttv version?

ian9k:
Yes 7.2.1.

Peuuuur-Noel:
Hi, got the same issue since a while. Using the latest version (7.2.52)
Looks like ".recommended-channels" is not on the element.

The element can be reach with this query:
`.tw-full-width.tw-flex-grow-1 > .tw-mg-b-2`

Possible fixe in "src/modules/hide_sidebar_elements/style.css":
```
.bttv-hide-featured-channels {
  .recommended-channels,
  .tw-full-width.tw-flex-grow-1 > .tw-mg-b-2 {
    display: none;
  }
}
```

![featured-channels](https://user-images.githubusercontent.com/22428121/53605610-50f89380-3bb8-11e9-9c6e-27876c374e0b.png)

whitecrownclown:
That's weird, it must mean they're AB testing something... It's still working for me.

![image](https://user-images.githubusercontent.com/8309417/53652706-2fe48100-3c52-11e9-80d1-b3415223665f.png)

Peuuuur-Noel:
Yep, maybe testing something.
As far as I can remember, I've got this issue before november 2018 but not really troubled by that.

night:
closing since I cannot reproduce this and we've not gotten recent reports. maybe it was an experiment that never shipped.

GreenTea13:
[...]

PR #3395, opened 2019-03-01
[Fix] Hide recommended channels

whitecrownclown:
@night Possible fix for #3355. Apparently, twitch is running some AB tests.


night:
this is unfortunately not a fix. selecting arbitrary table classes will only surface bugs.

Issue #3396, opened 2019-03-03
[Feature Request] Color back offline channels


Toni67:
Hi, i watch a lot of vods and with everything greyed out in the left column it's REALLY painfull to find the channel i want when they are offline.

There already a big red dot and viewers numbers next to the channel name to distinguish online from offline, so it would be really amazing if we had the possibility to color back all these streamer's avatars.

To add to the case, Twitch is not only about live, there a lot of content (maybe even more) on offline channels : vods, clips, uploads, highlights, infos/important links in channel home pages etc...

It would be even better if we had the possibility to rank them by alphbetical number.

Thank you

night:
Personally speaking I think without grey-ed out avatars it's unclear if a channel is live or not, so I agree with Twitch's decision to grey these out. Additionally, Twitch is primarily for live content and really only clips tend to be the only offline content people seem to care about. This is what use-case this sidebar is built for.

Toni67:
Dindn't see the reply before, i'm sorry.

If i can add:
Live channel would stay on top, offline on bottom with a line between both, would be easy to not confuse them, but maybe just let it off by defaut and offer the choice for peoples who would really need it?

I agree Twitch was and is still primaly made for live, but not only anymore as the quality of content never stop to increase.
And it depend on ppl but clips is not the only offline content people care about.
To take an exemple right now the top 1 live channel (lirik) has 54,254 viewers 
His last Vods from yesterday already get 30k view then it increase as the days pass 40k, 50k, 60,k 70k, his most viewed vod get 120.000 view

https://www.twitch.tv/lirik/videos?filter=archives&sort=time

I think Twitch dit it mostly because Vods is loss of earnings for them, no bits no sub, no donate..

There lot of different peoples on Twitch who use Twitch differently than you and me, but i know if i'm pulling out my hair everyday to scroll down / scroll back / "show more offline channel" / scroll down / "show more offline channel" / scroll down to fully deploit it and then notice i missed the channel i'm looking for... it's obliged that there should be other peoples who experience the same.

That why i beg you to rethink before throw out too fast this possible feature

Thank you

PR #3401, opened 2019-03-08
[Feature] Add option to have colored offline channels

itsazhuhere:
This should address the main request of #3396


night:
i actually agree with twitch's decision to make these b/w

Issue #3405, opened 2019-03-13
Anon Chat failing


Nareese:
A user reported that Anon Chat was no longer logging him out of chat.  When I attempted to turn it on to test, I get the following stack backtrace.
```
BTTV: Failed executing listener callback TypeError: Cannot read property 'userLogin' of undefined
    at t.getUsersFromEvent (https://static.twitchcdn.net/assets/pages.channel.components.channel-root-c00f72ffff801777c66d_ttv-prio.js:1:452021)
    at Array.n.handleMessage (https://static.twitchcdn.net/assets/pages.channel.components.channel-root-c00f72ffff801777c66d_ttv-prio.js:1:450386)
    at Object.t.handleMessage (https://static.twitchcdn.net/assets/pages.channel.components.channel-root-c00f72ffff801777c66d_ttv-prio.js:1:488470)
    at Object.bttvHandleMessage [as handleMessage] (http://127.0.0.1:2888/betterttv.js:9471:62)
    at t.n.pushMessage (https://static.twitchcdn.net/assets/pages.channel.components.channel-root-c00f72ffff801777c66d_ttv-prio.js:1:493911)
    at Object.sendChatAdminMessage (http://127.0.0.1:2888/betterttv.js:8650:24)
    at AnonChatModule.changeUser (http://127.0.0.1:2888/betterttv.js:468:20)
    at AnonChatModule.join (http://127.0.0.1:2888/betterttv.js:487:18)
    at AnonChatModule.load (http://127.0.0.1:2888/betterttv.js:497:22)
    at http://127.0.0.1:2888/betterttv.js:454:26
```

Looks like maybe TMIActionTypes renumbered again so that NOTICE was renumbered and sendChatAdminMessage is failing.  Attempts at various commands that also call sendChatAdminMessage, /localsub (Unrecognized command: /localsub), and /followed (silently returns no result) would seem to support this.

PR #3409, opened 2019-03-14
Anon chat fix

itsazhuhere:
Twitch changed/updated their Action Types, breaking at least one module, #3405


itsazhuhere:
Has some extraneous additions, removing and resubmitting.

Issue #3403, opened 2019-03-11
Blacklisted words not being filtered in VoD Chat


Latias4Ever:
I've referred to this issue in #3293, and it seemed to be fixed for a while, but now it has returned. When I watch VoDs on Twitch, the chat doesn't remove posts with things I have blacklisted.

To showcase this, I decided to test with both a BTTV emote, and an official Twitch Emote:
![image](https://user-images.githubusercontent.com/24550309/54151485-d559da80-4432-11e9-8e94-591f541e02d4.png)

Even after refreshing, this isn't working:
![image](https://user-images.githubusercontent.com/24550309/54151568-05a17900-4433-11e9-9cad-328e8c946f17.png)

I'm unsure what's going on, but the issue has resurfaced.

PR #3415, opened 2019-03-18
Blacklist fix

itsazhuhere:
Blacklisting is working for live streams, but stopped working for VODs, as per #3403.  The issue seemed to be the hide function.  Blacklisted messages were once again hidden when the containing li DOM element was passed into the markBlacklisted function instead.

I also made some changes to how the onVODMessage method parsed through the message text, to make it more similar to how the onChatMessage method works.  This is because I noticed that the old way of doing it wasn't working against native Twitch emotes.


night:
fixed by 98e757d95220dc674f478161b46d8a26ba9eb3fb

PR #3416, opened 2019-03-18
[WIP] [Feature] Add option to hide emotes in chat

### LINKED ISSUE #85: Turning off all Emoticons
Maybe I'm just a huge wet blanket but I hate emoticons, it'd be nice if there was just an easy way to turn them all off so I didn't have to see all that noise mucking up the text.

whitecrownclown:
@night Fixes #85	Simple toggle of chat emotes. We could hide the entire message if it's comprised of only emotes, however that would be more than just toggling a class. Thoughts?


night:
i would prefer we don't just hide emotes as then we're creating empty messages. we should really be replacing emotes with their alt text so messages still retain their context

night:
I'm gonna close this one due to activity, but feel free to open a new pull request when we have an implementation which shows alt text!

PR #3475, opened 2019-06-04
add display: block to tooltip div for link preview

### LINKED ISSUE #3473: chat_link_preview breaks link wrapping
A user on discord (hat tip: DragooMind) reported that hovering over a link, triggering link preview, breaks the wrap on the link making it unreadable in full without horizontal scrolling.  I had also noticed this behavior before, but hadn't looked into it.

It is caused by the div that chat_link_preview.tooltip constructs having the `tw-tooltip-wrapper` class added to it.  This class contains `display: inline-flex` which is what breaks the wrapping.  That attribute is not required for the proper functioning of the tooltip, and toggling it off in chrome debugger restores the proper wrapping, while the tooltip remains functional.

Not sure what the least hacky way to resolve this is... I will test some possible solutions and possibly submit a PR later if I get to something I'm happy with.

Nareese:
fixes #3473 

Adds a `display: block` to explicitly override the `display: inline-flex` in `.tw-tooltip-wrapper` class on the outer tooltip div for link preview.  Tested with both links and image preview.  Seems fine on both.

### DIFF
--- a/src/modules/chat_link_preview/index.js
+++ b/src/modules/chat_link_preview/index.js
@@ -9,7 +9,7 @@ const CHAT_LINK_SELECTOR = '.link-fragment';
 const IMAGE_REGEX = new RegExp('(https?:\/\/.)([a-z\-_0-9\/\:\.\%\+]*\.(jpg|jpeg|png|gif|gifv|webm|mp4))', 'i');
 
 function tooltip($url, body) {
-    const $tooltip = $('<div />');
+    const $tooltip = $('<div />').css('display', 'block');
     $tooltip.addClass('tw-tooltip-wrapper').addClass('inline');
     $tooltip.html(`
         <div class="tw-tooltip tw-tooltip--up tw-tooltip--align-center" data-a-target="tw-tooltip-label" style="width: 250px;


night:
can you post a picture of what this fixes? I cannot reproduce

Nareese:
Sure.

Before
wrap-before
After
wrap-after

The right edge of the images is the edge of my browser window, no text has been cut off in the snipping process.  After you mouse into the link and trigger preview, the formatting/wrapping changes to what is shown in the after image, where some of the text is off screen and there is now a horizontal scrollbar in chat that can be moved to show that hidden text.  This particular example is just a couple of letters, but I've seen worse, and it shouldn't happen at all.  Overriding the inline-flex prevents this behavior while everything else remains functional.

OH!  While trying to figure out why you wouldn't be seeing this, what could be different, it occurred to me.... font size.  I have my font size overriden to 18pt.  If I remove the font size override, I also cannot reproduce this behavior.  I will check with the original reporter and see if he is also using a larger font size.

night:
the css looks better prior to your change. I don't know why you would want some of the link text to be cut off

Nareese:
The before and after images are not before and after my change, it is what I see before and after I mouse into the link and trigger preview. ie. showing "what this fixes".   It is intended to show the problem, not the solution. The link looks normal, until you mouse into it, as shown in the Before image.  After you mouse into it, it re-wraps it, and shows what is in the After image.  The change in this PR is intended to prevent that re-wrapping.

PR #3485, opened 2019-06-07
[Feature] Add option to hide VOD chat header and text input

andrew-anguiano:
I watch a decent amount of VODs in a tiny window and hated how much room the chat input and header took up since I never type in VOD chat anyway. Added an option to hide those so there's more room for chat replay. Based it off of `hide_room_selector` which seems like a pretty similar idea.

### DIFF
--- a/src/modules/vod_chat_hide_input/index.js
+++ b/src/modules/vod_chat_hide_input/index.js
@@ -1,0 +1,22 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+
+class VodChatHideInputModule {
+    constructor() {
+        settings.add({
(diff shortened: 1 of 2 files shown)


night:
thanks for the pull request. i apologize for the lack of review. instead of a second setting, can we combine the styles to rely on the same setting? perhaps call it 'hide chat header' instead

andrew-anguiano:
@night no worries! i can combine them. although, i saw the other day that twitch said they're taking away vod chat... so does that make this obsolete?

night:
Yea, this look obsolete now. Will close it!

PR #3514, opened 2019-07-25
Update to v3 API

Slikrick:
This doesn't touch the chat_link_preview/link unfurler which are still on v2
This also didn't modify the chat emote tip https://github.com/night/BetterTTV/blob/8d5dda3951ae96fc9a97d25a8b5ca70432ae4ce6/src/modules/channel_emotes_tip/index.js#L18 as i was struggling to get the api info from the v2 api (kept returning 500 server error)


night:
we should drop the emote tips .. that code isn't needed anymore since emotes can be resolved by id

src/modules/emotes/global-emotes.js
```diff
@@ -12,6 +12,7 @@ const provider = {
     displayName: 'BetterTTV Global Emotes',
     badge: cdn.url('tags/developer.png')
 };
+const urlTemplate = '//cdn.betterttv.net/emote/{{id}}/{{image}}';
```
night: we shouldn't duplicate constants. maybe put in https://github.com/night/BetterTTV/blob/master/src/utils/cdn.js

src/modules/settings/index.js
```diff
@@ -154,7 +154,7 @@ class SettingsModule {
         cdn.get('privacy.html').then(data => $('#bttvPrivacy').html(data));
 
         api.get('changelog')
-            .then(({changelog}) => changelog.map(({version, publishedAt, body}) => changelogEntryTemplate(version, publishedAt, body)))
+            .then(changelog => changelog.map(({version, publishedAt, body}) => changelogEntryTemplate(version, publishedAt, body)))
             .then(releases => $('#bttvChangelog .bttv-changelog-releases').html(releases.join('')));
```
night: breaking change is we cannot `.html` this as we are not serving html which is pre-escaped any longer. need to render the plain text into what we want (should hotlink the issue #s back to github and parse new lines)

src/utils/api.js
```diff
@@ -1,7 +1,7 @@
 const $ = require('jquery');
 const querystring = require('querystring');
 
-const API_ENDPOINT = 'https://api.betterttv.net/2/';
+const API_ENDPOINT = 'https://api.betterttv.net/3/cached/';
```
night: should leave off `cached` bit, since that is part of the path .. we might hit stuff not being cached

src/watcher.js
```diff
@@ -384,7 +384,7 @@ class Watcher extends SafeEventEmitter {
 
             channel = currentChannel;
 
-            api.get(`channels/${channel.name}`)
+            api.get(`users/twitch/${channel.id}`)
                 .catch(error => ({
                     bots: [],
                     emotes: [],
```
night: if an error occurs you will cause an error since this fake response data is not identical in format to the normal real response

src/modules/emotes/global-emotes.js
```diff
@@ -26,8 +27,8 @@ class GlobalEmotes extends AbstractEmotes {
 
     updateGlobalEmotes() {
         api
-            .get('emotes')
-            .then(({urlTemplate, emotes}) =>
+            .get('emotes/global')
+            .then(emotes =>
                 emotes.forEach(({id, channel, code, imageType, restrictions}) => {
```
night: `channel` was removed

src/modules/emotes/channel-emotes.js
```diff
@@ -22,10 +23,10 @@ class ChannelEmotes extends AbstractEmotes {
         return provider;
     }
 
-    updateChannelEmotes({urlTemplate, emotes}) {
+    updateChannelEmotes({channelEmotes}) {
         this.emotes.clear();
 
-        emotes.forEach(({id, code, imageType, channel}) => (
+        channelEmotes.forEach(({id, code, imageType, channel}) => (
```
night: channel is now `user` which is a dictionary. it is optional and won't exist on channelEmotes

src/modules/emotes/channel-emotes.js
```diff
@@ -22,10 +23,10 @@ class ChannelEmotes extends AbstractEmotes {
         return provider;
     }
 
-    updateChannelEmotes({urlTemplate, emotes}) {
+    updateChannelEmotes({channelEmotes}) {
```
night: you are forgetting `sharedEmotes`

src/utils/api.js
Slikrick: Should the cache path only be used for emotes i assume? accessing https://api.betterttv.net/3/users/twitch/107780543 returns 404 without the /cached part

src/modules/emotes/channel-emotes.js
Slikrick: Fixed, i didn't notice the separation in the new api. I wasn't sure if there was a more concise way to write the shared/channel emote loops, so there is a similar loop for each.

PR #3528, opened 2019-08-04
[Feature] Add "chat in fullscreen" setting

hristov296:
Mainly for users with wide and ultra-wide monitors. Streams are almost always 16:9, which leaves black space on monitors with 21:9 for example. On 16:9 monitors it also has its uses.
Screenshot from 2019-08-04 20-26-17
Screenshot from 2019-08-04 20-26-50

### DIFF
--- a/src/modules/chat_in_fullscreen/index.js
+++ b/src/modules/chat_in_fullscreen/index.js
@@ -32,9 +32,11 @@ class ChatInFullscreen {
         if (fsElement && this.chatFsActive) {
             $('body').addClass('fullscreen-active');
             $('.right-column .channel-root__right-column').appendTo(fsElement);
+            $('#emote-menu-for-twitch').appendTo(fsElement);
         } else {
             $('body').removeClass('fullscreen-active');
             $('.channel-root__right-column').appendTo('.right-column > .tw-block');
+            $('#emote-menu-for-twitch').appendTo('body');
         }
     }
 }
--- a/src/modules/chat_in_fullscreen/index.js
+++ b/src/modules/chat_in_fullscreen/index.js
@@ -1,0 +1,42 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+
+class ChatInFullscreen {
+    constructor() {
+        settings.add({
+            id: 'chatInFullscreen',
+            name: 'Chat in Fullscreen',
+            defaultValue: false,
+            description: 'Shows chat section while in fullscreen mode.'
+        });
+        this.chatFsActive = false;
+        settings.on('changed.chatInFullscreen', () => this.toggleChatInFullscreen());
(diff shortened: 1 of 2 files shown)


night:
thanks for the pull request. i know this has been sitting for a while, and i'll try to review it soon

src/modules/chat_in_fullscreen/index.js
```diff
@@ -0,0 +1,44 @@
...
+        settings.on('changed.chatInFullscreen', () => this.toggleChatInFullscreen());
+        $(document).on('fullscreenchange mozfullscreenchange MSFullscreenChange webkitfullscreenchange', () => this.onFullScreenMoveChat());
+        this.toggleChatInFullscreen();
+    }
+
+    toggleChatInFullscreen() {
+        this.chatFsActive = settings.get('chatInFullscreen');
+        if (this.chatFsActive) {
+            $('body').addClass('bttv-chat-in-fullscreen');
+            $('.video-player__container').wrapInner('<div class="chat-in-fs-wrap" />');
```
night, line 22: this looks to be outdated to how Twitch renders fullscreen now. we may need a way to embed chat alongside the video element directly

src/modules/chat_in_fullscreen/index.js
```diff
@@ -0,0 +1,44 @@
...
+            $('body').removeClass('bttv-chat-in-fullscreen');
+            $('.video-player__container > .chat-in-fs-wrap').contents().unwrap();
+        }
+    }
+
+    onFullScreenMoveChat() {
+        const fsElement = document.fullscreenElement || document.mozFullScreenElement || document.msFullscreenElement || document.webkitFullscreenElement;
+
+        if (fsElement && this.chatFsActive) {
+            $('body').addClass('fullscreen-active');
```
night, line 33: where did this class name come from. in general, we like our class names to begin with `bttv`

src/modules/chat_in_fullscreen/index.js
```diff
@@ -0,0 +1,44 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+
+class ChatInFullscreen {
+    constructor() {
+        settings.add({
```
night, line 6: rather than adding a standard setting, i would prefer a player button when a user is in fullscreen

src/modules/chat_in_fullscreen/index.js
```diff
@@ -0,0 +1,44 @@
...
+            id: 'chatInFullscreen',
+            name: 'Chat in Fullscreen',
+            defaultValue: false,
+            description: 'Shows chat section while in fullscreen mode.'
+        });
+        this.chatFsActive = false;
```
night, line 12: let's move this out of class to a local variable in this file

night:
closing this due to inactivity, but feel free to reopen if you've updated this.

PR #3579, opened 2019-09-17
fixes #3576

### LINKED ISSUE #3576: bttv's ui elements in chat are misaligned
7.2.67

bttv

Nareese:
fixes #3576 

Update the selectors to insert before the settings button instead of the chat send button, since twitch moved the settings button to the immediate left of the send button.  ~~Also clean up no longer used code for putting the container in the rooms header.~~

### DIFF
--- a/src/modules/chat_state/index.js
+++ b/src/modules/chat_state/index.js
@@ -4,9 +4,7 @@ const twitch = require('../../utils/twitch');
 
 const CHAT_STATE_ID = 'bttv-channel-state-contain';
 const CHAT_STATE_TEMPLATE = require('./template')(CHAT_STATE_ID);
-const CHAT_HEADER_SELECTOR = '.rooms-header';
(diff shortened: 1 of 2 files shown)


src/modules/chat_state/index.js
```diff
@@ -30,14 +28,9 @@ function displaySeconds(s) {
 
 function loadHTML() {
     const $stateContainer = $(`#${CHAT_STATE_ID}`);
-    const $headerSelector = $(CHAT_HEADER_SELECTOR);
-    const $chatSendButtonSelector = $(CHAT_SEND_BUTTON_SELECTOR);
-    if (!$headerSelector.length || $headerSelector.find(CHAT_HEADER_PRIVATE_ROOM_SELECTOR).length) {
```
night: the header selector is correct code to handle removing the state indicators when navigating to a private room

PR #3607, opened 2019-10-10
Fix Gulp Windows Build and Migrating Gulp

Aaron128l:
- Remove both resizable (parent of draggy) and view-list from packages, leftovers from the old viewer lists.
- Fixes Gulp on Windows. Caused by babel on `twitch-chat-emotes`, due to OS slashes.
`[22:01:00] SyntaxError: xx\node_modules\twitch-chat-emotes\script.min.js: Assigning to 'arguments' in strict mode (7:22409)`
```js
ignore: [/node_modules\/(?!draggy)/] -> ignore: [/node_modules/]
```
- Change gulp to use exports rather than the old task system and removing `gulp.task()` all together.

Functionality should be identical, but testing is encouraged.


Nareese:
git-pr'd this into a clean clone, Gulp 4.0.2, gulp watch executes without issues and everything appears functional on a twitch channel page (Checked settings, font/size change, emote menu, etc).  I did not test dist.

PR #3620, opened 2019-10-20
improve chat state

thth:
Changes:
- Fixed the slow mode icon not showing a title text when hovered over. \
(On SVG elements, `title` should not be an attribute of the `svg` element but should instead be a child element)
- Fixed Twitch's "Chat Settings" tooltip displaying over the chat state icons:

  before | after
  --- | ---
  tooltip_before | tooltip_after

  I did this by moving the chat state template two levels up with `.parent().parent().before(CHAT_STATE_TEMPLATE);`. I'm unfamiliar with how often Twitch updates their layouts, so I'm unsure if this is an appropriate way of defining where to put the template. (e.g. would be prone to breaking if they remove the first parent element)
- Added a setting for displaying/hiding the chat state icons.

  How the setting works was modeled after the [host button module](https://github.com/night/BetterTTV/blob/904fd79081344402ab3de51468893ed4a3d759b2/src/modules/host_button/index.js#L29); seems to work fine in testing.

  I don't remember if it was worse before, but right now Twitch by default seems pretty good at communicating chat modes, so these icons can be subjectively undesirable visual noise.


night:
thanks for the pull request.

> I did this by moving the chat state template two levels up with .parent().parent().before(CHAT_STATE_TEMPLATE);. I'm unfamiliar with how often Twitch updates their layouts, so I'm unsure if this is an appropriate way of defining where to put the template. (e.g. would be prone to breaking if they remove the first parent element)

This is definitely risky. We try to avoid grabbing of parent nodes like this, as it's unclear what you might be grabbing if Twitch updates their UI. You might instead try traversing down to children from a root, rather than navigating parents. This is generally safer since we're scoping our changes to a set of children.

> Twitch by default seems pretty good at communicating chat modes

can you elaborate on this? iirc they don't really communicate them well at all other than the default chat text for slow mode. i try hard to avoid adding extra toggles in settings, since it adds to the visual noise of too many choices a user can make.

src/modules/chat_state/index.js
```diff
@@ -37,7 +40,16 @@ function loadHTML() {
...
+
+    if (
+        ($chatSettingsButtonSelector.parent().siblings('.tw-tooltip').length === 1)
+        && ($chatSettingsButtonSelector.parent().parent('.tw-tooltip-wrapper').length === 1)
+    ) {
+        $chatSettingsButtonSelector.parent().parent().before(CHAT_STATE_TEMPLATE);
+    } else {
+        // fallback to inserting directly before chat settings button if Twitch UI changes
+        $chatSettingsButtonSelector.before(CHAT_STATE_TEMPLATE);
+    }
```
thth: I added a check for the assumed tooltip DOM structure and a fallback so if the UI does change then it just goes back to how it's being inserted now (directly before the chat settings button).

> can you elaborate on this? iirc they don't really communicate them well at all other than the default chat text for slow mode.

emote only:

sub only:

slow mode (time only shown while slowed):

r9k (only shows when trying to send a non-unique message):


The only information that isn't communicated by default is the slow mode duration before having sent a message and r9k mode if you haven't sent a non-unique message.

> i try hard to avoid adding extra toggles in settings, since it adds to the visual noise of too many choices a user can make.

Personally, when I added BTTV I didn't want much of the features at all and was only looking for the functionality of persisting deleted chat messages. When I first looked at the settings and saw all the toggles, I was impressed with the extent of the modularity and how it seemed that pretty much every feature could be turned off if I wanted to. The chat state icons, however, was the only feature that stood out to me as something that I couldn't disable.

IMO, the elegance in the consistency of the modularity of features and ability to turn on/off every feature in the extension outweighs the cost of visual noise of additional settings. (But of course that's a design decision that's up to you 🙂)

thth:
I [added](https://github.com/night/BetterTTV/blob/4f3b8795ef6282c3278cd7ede7437634bcb94219/src/modules/chat_state/index.js#L44-L52) a check for the assumed tooltip DOM structure and a fallback so if the UI does change then it just goes back to how it's being inserted now (directly before the chat settings button).

> can you elaborate on this? iirc they don't really communicate them well at all other than the default chat text for slow mode.

emote only:

sub only:

slow mode (time only shown while slowed):

r9k (only shows when trying to send a non-unique message):


The only information that isn't communicated by default is the slow mode duration before having sent a message and r9k mode if you haven't sent a non-unique message.

> i try hard to avoid adding extra toggles in settings, since it adds to the visual noise of too many choices a user can make.

Personally, when I added BTTV I didn't want much of the features at all and was only looking for the functionality of persisting deleted chat messages. When I first looked at the settings and saw all the toggles, I was impressed with the extent of the modularity and how it seemed that pretty much every feature could be turned off if I wanted to. The chat state icons, however, was the only feature that stood out to me as something that I couldn't disable.

IMO, the elegance in the consistency of the modularity of features and ability to turn on/off every feature in the extension outweighs the cost of visual noise of additional settings. (But of course that's a design decision that's up to you 🙂)

night:
I see. My only issue with removing this feature is that it provides visual context to moderators, who bypass all of these UI features Twitch adds for normal chatters. Perhaps we could make this state show only for mods as a middle ground given that Twitch exposes the state to normal users now.

thth:
> 
> 
> I see. My only issue with removing this feature is that it provides visual context to moderators, who bypass all of these UI features Twitch adds for normal chatters. Perhaps we could make this state show only for mods as a middle ground given that Twitch exposes the state to normal users now.

Hiding the feature from non-mods seems unnecessarily clunky to me. If you think it's worth keeping the icons around but think that it isn't particularly helpful for the majority of users, having the feature be an option that's turned off by default makes the most sense imo.

night:
gonna do a trial run of removing it with 125632b5fd303b34f0aeca96ac829686ae379b50

PR #3627, opened 2019-10-26
BetterTTV Settings Redesign

Stoupe:
A potential redesign for the BetterTTV Settings panel

The only functionality changed was going from using two radio buttons to a single checkbox for the setting toggles, other than that it's just the style changes as seen in the screenshots below

Completely open to suggestions and making any changes

Whole Page-Old

Whole Page

About

Settings-Old

Settings

Settings-Search

Channel

Changelog

Privacy Policy

Backup-Import

Backup-Import Hover


night:
thanks for the pull request. i will review this when i have some free time

night:
apologies this review has taken so long to get to. there are a lot of open PRs on here which require a lot of time/thought to review in depth.

this is a good start!

my high level thoughts:

 - the main navigation bar should remain in the header. we are working in a smaller space, so adding this sidebar is not a good use of space.
 - settings should be refactored into categories, which could use a sidebar for sub navigation within the settings page
 - backup/import don't really need to be their own page anymore
 - privacy policy should probably just link to https://betterttv.com/privacy these days
 - settings are most likely moving to react at some point, and complexity of settings will increase such that they are not just simple on/off switches. this design doesn't look like it will support such requirements

my nits:

 - the font sizes are way too large. unlike Twitch, I don't think large fonts look good on websites. we should target similar sizes
 - border radius is too large
 - the primary background is too black, we should scale back to grey
 - the close icon overlaps with settings content
 - the colors don't really match the website, so we should either change website color or extension colors to match

night:
gonna close this due to inactivity, but feel free to reopen when there's some progress made on this!

PR #3637, opened 2019-10-31
[Feature] Add picture in picture button

### LINKED ISSUE #3550: Picture In Picture
Would be nice to have a button to enable the native PiP from Chrome.

`document.querySelector('div.player-video video').requestPictureInPicture();`

If I have the time I will try to implement and open a PR for that :)

whitecrownclown:
Closes #3550

@night


src/modules/video_player/index.js
```diff
@@ -135,6 +150,43 @@ class VideoPlayerModule {
...
+            $('#bttv-pip')
+                .find('svg')
+                .html(pipEnabledSVG);
+        });
+
+        const $anchor = $('.player-controls__right-control-group');
+        const $button = $anchor
+            .children()
+            .last()
+            .clone(false, false)
```
night: this feels brittle. can we instead store a template string with the entire button html?

whitecrownclown: Done!

PR #3645, opened 2019-11-08
[Feature] Add followers count

### LINKED ISSUE #3602: Show Follower count on channel page
Life was good when I could open a twitch channel and see the amount of followers they had.
Now I have to click on the word Followers and wait for a new page to load up.

BTTV could make my Twitch experience SO MUCH BETTER if I could just get back what Twitch took away: https://cdn.discordapp.com/attachments/326719968706166785/629035968208568320/unknown.png

Thank you.

whitecrownclown:
Closes #3602.

I think we should show this by default, no need for a toggle.

@night 
Thoughts?

Thanks!

### DIFF
--- a/src/modules/followers_count/index.js
+++ b/src/modules/followers_count/index.js
@@ -1,0 +1,40 @@
+const $ = require('jquery');
+const twitchAPI = require('../../utils/twitch-api');
+const twitch = require('../../utils/twitch');
+const watcher = require('../../watcher');
+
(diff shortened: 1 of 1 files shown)


night:
this would unfortunately require a toggle .. there are millions of BTTV users so this would generate millions of extra API requests that Twitch would not be too happy about.

whitecrownclown:
@night Got it, I've made the changes. Thanks!

night:
it looks like follower count is included on channel pages now, so this is no longer needed

PR #3649, opened 2019-11-11
Hide emote-only messages

Aloisius:
Create a setting that allows filtering of messages with just emotes and no text or URL.

### DIFF
--- a/src/modules/chat_highlight_blacklist_keywords/index.js
+++ b/src/modules/chat_highlight_blacklist_keywords/index.js
@@ -161,6 +161,12 @@ function messageTextFromAST(ast) {
     }).join(' ');
 }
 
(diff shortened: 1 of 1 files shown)


Aloisius:
I wrote this to avoid the floods of emote-only messages that drown out actual chat.

Could be expanded to just a general low-quality comment filter. Maybe one that only kicks in if the message rate goes over a certain limit. As is though, it really cuts down on the junk.

night:
this should be reimplemented as a chat command local filter that restricts messages to textual content only.. we can use `/localtext` and `/localtextoff`

night:
closing this due to lack of activity. feel free to re-open this PR if the requested changes are made

PR #3681, opened 2019-12-07
[Fix] Conflict with FFZ features

### LINKED ISSUE #3676: Color correction is being applied incorrectly in combination with FFZ features
Using FrankerFaceZ, users can apply different background colors to messages highlighted for different reasons. This is achieved by adding a `background-color:` in-line style to `.chat-line__message` elements.

When BetterTTV applies its color adjustments (and specifically at https://github.com/night/BetterTTV/blob/master/src/modules/chat/index.js#L193 ), it searches the entire in-line style string for `color:` which will also match `background-color:`, resulting in unintentional behavior where a user's name color will be applied to their message text even though they did not use `/me`.

Fixing this within FrankerFaceZ is infeasible as FFZ processes chat lines before BetterTTV's own code runs.

Please ensure that only `color` itself is matched and not any other CSS properties. For a minimal change, I'd suggest something like replacing

```javascript
if (messageStyle && messageStyle.includes('color:')) {
```
with
```javascript
if (messageStyle && /(?:^|\s)color:/.test(messageStyle)) {
```

Thank you.

whitecrownclown:
Closes #3676

### DIFF
--- a/src/modules/chat/index.js
+++ b/src/modules/chat/index.js
@@ -191,7 +191,7 @@ class ChatModule {
         }
 
         const messageStyle = $element.attr('style');
-        if (messageStyle && messageStyle.includes('color:')) {
+        if (messageStyle && /(?:^|\s+)color:/.test(messageStyle)) {
             $element.css('color', color);
         }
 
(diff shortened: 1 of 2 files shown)


night:
fixed by 8d1f1641d2f7bc987bf2bb8711b0824fb793c2b6

Issue #3691, opened 2019-12-15
Left side chat is partially blocked


PairedPrototype:
Since Twitch changed some stuff on the site a week or so ago, the channel list overlays about 50px of the chat when using Left Side Chat.
![image](https://user-images.githubusercontent.com/16809256/70857019-1eb5e980-1edf-11ea-9e94-662fd7e4e8fc.png)

xgerhard:
Hiya, have you tried clearing your cache/cookies this appears to be fixed already.

PairedPrototype:
It did try that, but it did nothing at first. Turns out I left it on the clear cache from the last 24 hours. Changing that to a little later appears to have fixed it! Thanks 😊

I also wasn't aware Chrome cached extension styles 🤔 Might be a good idea to get a cache-busting method put into the build process.

PR #3692, opened 2019-12-15
Fix left side chat being obstructed by channel list

PairedPrototype:
This fixes https://github.com/night/BetterTTV/issues/3691


PairedPrototype:
This issue appears to have already been resolved, but a caching issue (as noted on the related issue https://github.com/night/BetterTTV/issues/3691#issuecomment-565803679) prevented the fix from taking effect.

PR #3705, opened 2019-12-27
Add module to hide community highlights; Fixes #3589, Fixes #3715

### LINKED ISSUE #3589: Community Gift Chests
Can we get an option to hide this community gift chest thing at the top of chat? It takes up a ton of room and doesn't go away
Screen Shot 2019-09-22 at 4 38 03 AM

### LINKED ISSUE #3715: Hide hype train
A new feature was released on Twitch recently called hype train which I guess is a way to try and make viewers to donate more for a random emote. There seems to be no limit how many times this can be triggered and it covers about 30% of the chat and is very distracting. A setting to hide this feature would be much obliged.

Brcrwilliams:
Gets rid of the very annoying Community Gift Chest toast which covers a significant portion of the chat and isn't dismiss-able. 

## Before

Before

## After

After


gertjanmaas:
This also seems to block the hype train toast. Can the wording be renamed to hide 'Community Highlights' instead of just the gift chests? @Brcrwilliams

Brcrwilliams:
Sure!

Brcrwilliams:
@night There seems to be high demand for this feature (#3705, #3715). Could we get a review / merge?

BlokkeLollie:
Hi,
I got rid of the hypetrain with this module, but had to find the solution here on GitHub. Might I suggest that this module shows up when in the searchbox for modules the query 'train*' is entered? I think it would become way more accesible to users that way.

Issue #3663, opened 2019-11-26
About page links Chrome Web Store on all browsers


Stevoisiak:
**Summary:**
The "About" page asks users to leave reviews on the Chrome Webstore, even if they are using a different web browser. (Firefox, Microsoft Edge)

**System:**
- Firefox 70.0.1
- BetterTTV 7.2.71

**Steps to reproduce:**
1. Go to https://www.twitch.tv/ while using Firefox.
2. Click top-left profile icon.
3. Open BetterTTV Settings > About.
4. Click "Drop a review"

**Expected behavior:**
When using Firefox, the link opens the [Firefox Add-on page for BetterTTV](https://addons.mozilla.org/en-US/firefox/addon/betterttv/).

**Actual behavior:**
When using Firefox, the link opens the [Chrome Web Store page for BetterTTV](https://chrome.google.com/webstore/detail/ajopnjidmegmdimjlfnijceegpefgped).

**Relevant code:**
https://github.com/night/BetterTTV/blob/e30bb3b9575c6b42ae4366152bdaaf307887866c/src/modules/settings/index.js#L55-L57

d-tara:
I am new to contributing to GitHub and I feel like this issue would be a good place to start for me. If it is alright with you I would like to fork it and try to fix it.

Stevoisiak:
@d-tara Go ahead! If you solve the issue, be sure to submit a pull request.

d-tara:
I have solved the problem and submitted a pull request. Please let me know what has to be done next.

PR #3709, opened 2019-12-31
Fixed the web store link in the about page

d-tara:
Let me know if this was satisfactory


Stevoisiak:
Resolves #3663

night:
edge reviews do not matter because they are swapping over to a new extensions marketplace which has no reviews .. we also direct chromium edge users to chrome extension now

PR #3712, opened 2020-01-06
[WIP] [Feature] Add emote tab completion to whispers

### LINKED ISSUE #3614: Add emote tab completion to whispers
Heya,

Thought it might be a good idea to add emote tab completion to whispers, seeing as it already supports rendering them.

Thanks,
Elliott

whitecrownclown:
@night 
Closes #3614

### DIFF
--- a/src/modules/chat_tab_completion/index.js
+++ b/src/modules/chat_tab_completion/index.js
@@ -6,6 +6,8 @@ const keyCodes = require('../../utils/keycodes');
 const emotes = require('../emotes');
 
 const CHAT_INPUT_SELECTOR = '.chat-input textarea';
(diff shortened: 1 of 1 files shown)


night:
it's unfortunately not this simple, or it would already be fixed. whispers work across all of twitch.

- this doesn't work on non-chat pages
- emotes are not loaded when not on a page containing chat
- emotes that tab complete include channel emotes, which should not be made available to the tab completion in whispers
- users can be tab completed, which should not be made available to the tab completion in whispers

whitecrownclown:
I'm sorry, I honestly don't use whispers and I didn't think of these cases. I'll have a look over it again and come back with an update.

PR #3718, opened 2020-01-13
[Feature] Set auto theatre only when channel is live

### LINKED ISSUE #3315: Automatic Theatre Mode When Channel is Live
When **Automatic Theatre Mode** is enabled, it always opens with Theatre Mode which seems unpractical.
Is it possible to add an option to switch to Automatic Theatre Mode only when the channel is live?

whitecrownclown:
Closes #3315


jmuhlfel:
hi, random user here, would love to see this merged. might also be worth adding a `when a channel is live` on line 12?

night:
since twitch's channel page redesign this looks to be a nonissue

PR #3730, opened 2020-01-23
[Feature] Joinable Steam Lobby Hyperlinks

Treee:
Steams offers a nice api to create links that launch the steam bootstrapper from a browser for games that are public to join. This update adds a new regex string to match the format of a steam link `steam://joinlobby/gameId/lobbyId/steamUserId` and replace the plain text with a clickable hyperlink.

### DIFF
--- a/src/modules/chat/index.js
+++ b/src/modules/chat/index.js
@@ -13,7 +13,7 @@ const splitChat = require('../split_chat');
 
 const EMOTE_STRIP_SYMBOLS_REGEX = /(^[~!@#$%\^&\*\(\)]+|[~!@#$%\^&\*\(\)]+$)/g;
 const MENTION_REGEX = /^@([a-zA-Z\d_]+)$/;
-const STEAM_LOBBY_JOIN_REGEX = /^steam:\/\/joinlobby\/[0-9]*\/[0-9]*\/[0-9]*$/;
+const STEAM_LOBBY_JOIN_REGEX = /^steam:\/\/joinlobby\/d+\/d+\/d+$/;
 const EMOTES_TO_CAP = ['567b5b520e984428652809b6'];
 const MAX_EMOTES_WHEN_CAPPED = 10;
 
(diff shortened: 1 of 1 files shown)


src/modules/chat/index.js
```diff
@@ -13,6 +13,7 @@ const splitChat = require('../split_chat');
 
 const EMOTE_STRIP_SYMBOLS_REGEX = /(^[~!@#$%\^&\*\(\)]+|[~!@#$%\^&\*\(\)]+$)/g;
 const MENTION_REGEX = /^@([a-zA-Z\d_]+)$/;
+const STEAM_LOBBY_JOIN_REGEX = /^steam:\/\/joinlobby\/[0-9]*\/[0-9]*\/[0-9]*$/;
```
EhsanKia: You can replace [0-9] with \d, and also use + instead of * since these IDs are guaranteeds non-empty

src/modules/chat/index.js
```diff
@@ -54,24 +57,23 @@ function hasNonASCII(message) {
 class ChatModule {
     constructor() {
         watcher.on('chat.message', ($element, message) => this.messageParser($element, message));
-        watcher.on('channel.updated', ({bots}) => {
+        watcher.on('channel.updated', ({ bots }) => {
```
EhsanKia: undo all these cosmetic formatting changes? They're not relevant to this PR?

src/modules/chat/index.js
```diff
@@ -157,6 +159,14 @@ class ChatModule {
                     continue;
                 }
 
+                // check steam parasbility
+                const steamJoinLink = part.match(STEAM_LOBBY_JOIN_REGEX);
+                if (part.length > 0 && part.substring(0, 18) === 'steam://joinlobby/' && steamJoinLink && steamJoinLink[0]) {
```
EhsanKia: all of these checks except the 3rd one are unnecessary. If the regex matches, then by definition 1, 2 and 4 must be true.

src/modules/chat/index.js
Treee: completed.

src/modules/chat/index.js
Treee: removed unnecessary checks

src/modules/chat/index.js
Treee: replaced [0-9]* with \d+

src/modules/chat/index.js
EhsanKia: I also meant all the other extra whitespace added.

Treee: Removed all white space, and newlines not relevant to code change.

PR #3754, opened 2020-02-14
[Feature] Add "Hide Channel Leaderboard" Option

anawww-k:
Annoyed by that channel leaderboard of top sub-gifters and bit-cheerers clogging up the top of yr chat ?

Here's a (tiny) PR for a BetterTTV setting for that !

<3

### DIFF
--- a/src/modules/hide_channel_leaderboard/index.js
+++ b/src/modules/hide_channel_leaderboard/index.js
@@ -1,0 +1,21 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+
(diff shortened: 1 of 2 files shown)


night:
this has overlap with the hide bits feature we already have, since it hides the leaderboard

anawww-k:
i understand the overlap - functionality feels distinct to me, but i understand if u feel the "hide bits" being the only option to hide the leaderboard is more appropriate

if there's any change you would recommend, lmk, but if not thats understood too thanks for maintaining BTTV <3

Issue #3424, opened 2019-03-29
Hide Squad Streaming banner


kirkdmo:
With Twitch introducing the new "Squad Streaming" feature, it's rather annoying having the banner visible across the top of the stream. Would be nice to have an option to hide the banner.

Timeeee:
@night any news on this?

billysielu:
Related: #3251

Feky:
Anytime soon this is annoying and can't hide it?

night:
it looks like twitch has made this less intrusive, so this is no longer needed

PR #3774, opened 2020-03-03
Add option to hide the purple squadstream bar above the player

chrisheib:
See Issue #3424
Just a simple implementation without much fuzz. 
My first contribution here, if there is something to be noted please tell and I will improve.

### DIFF
--- a/src/modules/hide_squadstream_bar/index.js
+++ b/src/modules/hide_squadstream_bar/index.js
@@ -1,0 +1,21 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+
+class HideSquadStreamBarModule {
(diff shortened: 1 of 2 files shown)


chrisheib:
Urg, this needs some further improvement, it seems to also kill some other icons.

chrisheib:
Now it should only be affecting the correct bar.

night:
Apologies for the delay in response. Your CSS selector does not look specific enough to classify just the squad stream bar. Because of this, it is possible you may mistakenly hide some other element with a similar pattern of animation implemented.

night:
it looks like twitch has made this less intrusive, so this is no longer needed

Issue #3828, opened 2020-04-25
twitch homepage autoplay disable doesn't work


rohankeluskar1:
Even If I set the BTTTV settings to disable homepage autoplay, it still autoplays featured streamers, it worked for a while, now it doesn't

rohankeluskar1:
working again now

PR #3830, opened 2020-04-27
Fix twitch.getCurrentPlayer

whitecrownclown:
Fixes: #3828

@night


night:
I'm not quite sure how this is the fix. Homepage auto pausing works locally for me without this patch. Is there steps to reproduce that I'm missing?

whitecrownclown:
@night 

When I opened this PR it wasn't working for me either. The `getCurrentPlayer` function was returning `{core: playerInstance}` instead of `playerInstance`.
Now it's working again.. I can only assume twitch changed things back.

PR #3862, opened 2020-05-12
Update chat list selector

whitecrownclown:
@night 

`.chat-list__lines` has been removed. We should update all occurrences with the new selector `.chat-scrollable-area__message-container`


Nareese:
The patch to chat_font_settings in this PR duplicates my PR #3857, but lacks the additional change to the name of the whispers container that is in that PR.  Since this PR is more comprehensive and integrates fixes for multiple reported and unreported issues all stemming from this same class (TBH, I should have checked this myself), if you can add that change for the whispers from my PR to yours, I will close mine.

PR #3864, opened 2020-05-12
[Fix] Picture in picture interaction with miniplayer

### LINKED ISSUE #3829: Bug with PIP button missing because of mini-player
There's a bug I've been seeing for a while now, everytime the mini-player gets activated or shown the PIP button, when you hover on the video player, would suddenly disappear until I refresh the page.

whitecrownclown:
@night 

Fixes #3829

### DIFF
--- a/src/modules/video_player/index.js
+++ b/src/modules/video_player/index.js
@@ -110,13 +110,62 @@ function togglePictureInPicture() {
     }
 }
 
+function insertPiPButton() {
+    const $anchor = $('.player-controls__right-control-group');
+    const $settingsButton = $anchor.children('div').children('div.settings-menu-button-component').parent();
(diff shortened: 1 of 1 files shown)


src/modules/video_player/index.js
```diff
@@ -110,13 +110,62 @@ function togglePictureInPicture() {
...
+
+    const miniPlayerObserver = new window.MutationObserver(mutations =>
+        mutations.forEach(mutation => {
+            if (mutation.oldValue === 'mini') {
+                insertPiPButton();
+            }
+        })
+    );
+
+    observe(miniPlayerObserver, $('.persistent-player')[0]);
```
night, line 143: lets avoid using an observer for this. there's gotta be a better way to detect this behavior

PR #3888, opened 2020-05-14
[Feature] Gifted subscription status on viewer-card

ipiv:


night:
can you explain your use case and intentions for this, as your PR comment is empty.

ipiv:
Serves same purpose as already present user stats on the viewer-card. Apart from being just a statistic, the ability to view whether the user has a gifted subscription can be helpful in moderator decision-making.

brandinarsenault:
This is useful information, I've just made a custom API to check but sometimes we don't have the time to type out a chat command just to check their sub status before making a mod decision, and a more silent way would be ideal. If this isn't a heavy load PR, I know moderators would appreciate this.

src/modules/chat_moderator_cards/moderator-card.js
```diff
@@ -86,6 +88,12 @@ const NICKNAME_CHANGE_BUTTON_TEMPLATE = `
     </button>
 `;
 
+const userGiftStatusTemplate = (gifter, giftedAt) => `
+    <div class="bttv-moderator-card-gift-status tw-flex"">
```
night, line 92: `""`

src/modules/chat_moderator_cards/moderator-card.js
```diff
@@ -86,6 +88,12 @@ const NICKNAME_CHANGE_BUTTON_TEMPLATE = `
...
+        ${Icons.GIFT}<p class="tw-c-text-overlay tw-mg-l-05 tw-mg-t-auto">Gifted by ${gifter} ${giftedAt}</p>
```
night, line 93: external data must be html escaped

src/modules/chat_moderator_cards/moderator-card.js
```diff
@@ -178,6 +187,34 @@ class ModeratorCard {
         });
     }
 
+    renderGiftStatus() {
+        const $header = this.$element.find(MODERATOR_CARD_HEADER_SELECTOR);
+        if ($header.find('.bttv-moderator-card-gift-status').length) return;
+
+        const currentChannel = twitch.getCurrentChannel();
```
night, line 194: this needs to handle null current channel

src/modules/chat_moderator_cards/moderator-card.js
```diff
@@ -178,6 +187,34 @@ class ModeratorCard {
...
+                            }
+                            isGift
+                        }
+                    }
+                }
+            }
+        }`;
+
+        twitchAPI.graphqlQuery(query).then(({data: {user: {relationship: {subscriptionBenefit}}}}) => {
+            if (!subscriptionBenefit || !subscriptionBenefit.gift.isGift) return;
```
night, line 212: this needs to handle null `subscriptionBenefit.gift`

src/modules/chat_moderator_cards/moderator-card.js
```diff
@@ -178,6 +187,34 @@ class ModeratorCard {
...
+            const giftStatusHTML = userGiftStatusTemplate(subscriptionBenefit.gift.gifter.displayName, moment(subscriptionBenefit.gift.giftDate).fromNow());
```
night, line 213: this needs to handle null `subscriptionBenefit.gift.gifter`

night:
can you also better explain why this is a benefit for moderators? why does it matter if a subscription is gifted or not?

ipiv:
> can you also better explain why this is a benefit for moderators? why does it matter if a subscription is gifted or not?

I would say it's about the fact that more then often moderators find themselves hesitating to time out/ban users with active subscription as it might increase validity. 

Simply put, to distinguish trolls - E.g. Knowing the fact that user, who has just made inappropriate comment or broken rules, could be a lucky troll who received a gift from a generous user, would solidify the fact that the same user might've not contributed anything positive to the community.

night:
Thanks for the request.

Since we now fetch from GraphQL and recently pruned views from cards, this was easy an easy add without bloating the cards: https://github.com/night/betterttv/commit/2585c13845b6fd943a5dd7d51464e1bc39e6f60f

Issue #3874, opened 2020-05-13
[BUG] BTTV emotes not working on embedded twitch


kc1r74p:
Using: 
Firefox Developer edition
BTTV 7.2.79

Since about 1-2 days ago the embeded twitch player with chat does not display the BTTV features anymore, no additional settings, no BTTV emotes displayed in chat...

Using embeded twitch like offical API docs: https://dev.twitch.tv/docs/embed/everything
Previously BTTV worked 1:1 like on the original Twitch website.

Best regards

PR #3926, opened 2020-05-27
[Fix] Embeded usage on 3rd party sites

kc1r74p:
The [Twitch.tv embeded player](https://dev.twitch.tv/docs/embed/everything) was working fine with bttv  in the past for 3rd party sites, but it broke recently when I updated bttv.

Fixes: https://github.com/night/betterttv/issues/3874


night:
it seems to be working fine without these changes

PR #3997, opened 2020-06-24
Enable GIF Emotes by default

brandinarsenault:
It is common to see chatters in Twitch Chats, the NightDev Discord and GitHub issues complaining about not being able to see some emotes. The onboarding of users to BTTV is a step in itself, having to describe how to further their experience that is engraved in the culture of Twitch and BTTV by enabling another option creates for a lengthy discussion. We should break this barrier by enabling this setting by default. Thank you for your consideration in merging this PR.

### DIFF
--- a/src/modules/emotes/index.js
+++ b/src/modules/emotes/index.js
@@ -28,7 +28,7 @@ class EmotesModule {
         settings.add({
             id: 'bttvGIFEmotes',
             name: 'BetterTTV GIF Emotes',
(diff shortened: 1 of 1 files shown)


Nareese:
I support this PR, and also think consideration should be given to extending it to the FFZ Emotes and Emote Menu options as well.

alazymeme:
👍 Looks good to me

PR #4000, opened 2020-06-25
[Feature] Display chat even if permanently banned from channel

felipekunzler:
This PR includes a new setting `Display Chat if Banned` that shows chat messages even if the user's been permanently banned from the channel.

Before:
banned

After:
unbanned

The changes required were the following:
1. Show chat react component even if banned.
2. Use anonymous credentials for chat web socket in order to retrieve chat messages.
    * If **Anon Chat** setting is enabled nothing is required here as it takes care of it.
    * If **Anon Chat** setting is not enabled:
        1. Listen to web socket banned message and switch to anonymous user in that case.
        2. Listen to part (leaving channel) message and switch back to the original user if current user is anonymous.

Tested locally and seems to be working good on all cases.

### DIFF
--- a/src/modules/banned_chat/index.js
+++ b/src/modules/banned_chat/index.js
@@ -1,0 +1,62 @@
+const settings = require('../../settings');
+const watcher = require('../../watcher');
+const twitch = require('../../utils/twitch');
+const anonChat = require('../anon_chat/index');
+
+class BannedChatModule {
+    constructor() {
+        settings.add({
+            id: 'bannedChat',
+            name: 'Display Chat if Banned',
+            defaultValue: false,
+            description: 'Show chat messages even if permanently banned from the channel'
+        });
+
+        watcher.on('load', () => this.attachBannedChatSocketListener());
+        watcher.on('load.chat', () => this.displayChatComponent());
+    }
+
+    displayChatComponent() {
+        if (settings.get('bannedChat')) {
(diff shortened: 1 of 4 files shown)


src/modules/banned_chat/index.js
```diff
@@ -0,0 +1,62 @@
...
+                    anonChat.part();
+                } else if (this.leavingChannelMessage(event)) {
+                    anonChat.join();
+                }
+            });
+        }
+    }
+
+    userBannedFromChannelMessage(event) {
+        return event.data && event.data.startsWith('@msg-id=msg_banned');
```
night, line 54: hmm .. why do we rely on raw socket message values versus the value from `isCurrentUserBanned`? what's going on there?

src/modules/banned_chat/index.js
```diff
@@ -0,0 +1,62 @@
...
+            if (!client) return;
+
+            const socket = client.connection.ws;
+            if (!socket || socket.isBannedChatListenerAttached) return;
+
+            socket.isBannedChatListenerAttached = true;
+            socket.addEventListener('message', event => {
+                if (this.userBannedFromChannelMessage(event)) {
+                    anonChat.part();
+                } else if (this.leavingChannelMessage(event)) {
```
night, line 46: rather than relying on socket for this, we could just rely on the channel change event from our watcher.

night:
closing this due to inactivity, but feel free to reopen if you've updated this.

Issue #4036, opened 2020-07-11
Left Side Chat


Zepatt:
I noticed today that the Left Side Chat setting is a bit broken, this is what twitch currently looks like with it enabled: https://gyazo.com/1723dbea8cf2f5b99be2c682641f8091
But before today the "Followed Channels" tab was moved to the right side if you had Left Side Chat enabled, meaning that the stream video player was nicely placed in the middle between the chat and the "Followed Channels" tab, and it looked like this but the chat and "Followed Channels" tab switched places: https://gyazo.com/3652e6f9490ffc7f094775887ade7136

I hope this is indeed a bug because everything was fine before and hopefully this can be fixed

night:
duplicate of #4029

Issue #4035, opened 2020-07-11
Left side chat also shows followers in theater mode


Tobbax:
Use left side chat and theater mode your followers will also show besides chat
![image](https://user-images.githubusercontent.com/44871092/87219830-e3897500-c35e-11ea-8d0e-da928a49d49d.png)

using default (right side) no followers is show in theater mode
![image](https://user-images.githubusercontent.com/44871092/87219878-4b3fc000-c35f-11ea-93a8-899348cdced1.png)

pls help🙆‍♂️

night:
duplicate of #4029

Issue #4033, opened 2020-07-10
Channel list is broken when "Left Side chat is enabled" in Theatre mode


Chromerian:
My usual set up is left side chat, theatre mode. Then if I am running more than one stream I have the stream window at it's thinnest so chat appears below the stream.
A few weeks ago, the channel list started messing with the chat, overlapping it. It wasn't a big issue but now it is even more broken all of a sudden. With left chat activated, when I go into minimised view, the channel list still appears and pushes the chat to the side so I can't access the emote menu or see the right edge of the chat. Please can you fix this and maybe even have a option to remove the channel list bar entirely as I never use it. There are options to remove offline and friends from it but not the whole bar.

Supplied are screenshots of:
Right side chat, large theatre mode, all working: https://i.imgur.com/2EtjjHL.jpg
Right side chat, small theatre mode, all working: https://i.imgur.com/XdPffmk.png
Left side chat, large theatre mode, Channel list showing (it used to stay hidden in theatre): https://i.imgur.com/x101Uh7.jpg
Left side chat, small theatre mode, chat completely broken and shifted due to channel list: https://i.imgur.com/c7fhnEQ.png

night:
duplicate of #4029

Issue #4029, opened 2020-07-10
Left-Side Chat.


FollowSonik:
Yo! Excuse me, what happened with left-side chat BetterTTV option?

![image](https://user-images.githubusercontent.com/11148271/87206300-1b1e0000-c312-11ea-9441-74b3d18d6158.png)

MoDog0:
I am having the same issue started about the same time as Sonik opened this issue.
Left side chat results in followed channels etc bar being on the left side with left side chat.
I have recommended channels set to hidden and they are showing up as-well.

PR #4037, opened 2020-07-11
fix(chat): left side chat

whitecrownclown:
@night 

They all seem to be duplicates.

Fixes: #4036, #4029, #4035, #4033


EhsanKia:
Can confirm, this fixes the issue. Thank you for the patch.

PR #4093, opened 2020-08-14
[Feature] Add chat replies setting to toggle click to reply

### LINKED ISSUE #4088: Allow disabling Click to Reply
In a recent update Twitch added a feature where users can click on any message in a chat, which then allows them to directly reply to a message, starting a new discussion thread. 

kuva

However, the feature also effectively blocks things like highlighting text in a message, and also increases the amount of misclicks (at least for me, running Firefox 79.0 on Windows 10). It would be great if the option to disable the click-to-reply feature could be considered for BTTV as well.

Sorry if this is a duplicate!

whitecrownclown:
* Added a new option to toggle the `click to reply` button in chat.

Closes #4088

### DIFF
--- a/src/modules/chat_replies/index.js
+++ b/src/modules/chat_replies/index.js
@@ -1,0 +1,21 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+
+class ChatRepliesModule {
+    constructor() {
+        settings.add({
+            id: 'chatReplies',
+            name: 'Chat Replies',
+            defaultValue: true,
+            description: 'Toggles click to reply in chat'
+        });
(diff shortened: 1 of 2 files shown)


src/modules/chat_replies/index.js
```diff
@@ -0,0 +1,21 @@
+const $ = require('jquery');
+const settings = require('../../settings');
+
+class ChatRepliesModule {
+    constructor() {
+        settings.add({
+            id: 'chatReplies',
+            name: 'Chat Replies',
```
night, line 8: can we invert this setting name, so by default users have a false option to be enabled

whitecrownclown: updated!

Issue #4095, opened 2020-08-15
Add setting for Clear Chat handler and default to false


brandinarsenault:
The /clear command is a powerful moderation tool that can be used to very quickly deal with a lot of moderation issues in a chat. We encourage many users to use BTTV, and we remove one of our moderation tools by doing so. The CLEAR_CHAT case in BTTV (index.js line 49) is hardcoded without a setting. We should add a setting toggle for this case and only prevent the default action if the setting is set to on. This setting should be toggled off by default, so we can continue using this moderation action.

Thank you for your consideration.

night:
https://github.com/night/betterttv/pull/4096#issuecomment-683368383

PR #4096, opened 2020-08-15
Add ignoreClearChat setting default to false to allow chat clears

brandinarsenault:
This PR resolves the issue I opened on #4095 to allow moderators to clear chat even when we encourage users to use BTTV. This option defaults to false and can be turned on easily anytime in the BTTV settings. This change will give moderators back the power and not hesitate to recommend the use of BTTV without losing access to a useful and powerful command from Twitch.


night:
In general we prefer to avoid adding toggles for the sake of adding optionality. Moderators have the ability to remove individual messages in chat, so I see no reason why this option needs to exist. Clearing all chat plays no role except to censor chat, which we're not advocates of.

PR #4119, opened 2020-08-31
FIX: update command for modcard ignore hotkey

Nareese:
At some point, Twitch changed the chat slash commands from /ignore and /unignore, to /block and /unblock.  The modcard hotkey "i" sends the ignore command.  I have had this updated in my local repo for a while, and somehow never remembered to submit a PR for it...

### DIFF
--- a/src/modules/chat_moderator_cards/moderator-card.js
+++ b/src/modules/chat_moderator_cards/moderator-card.js
@@ -13,7 +13,7 @@ const Commands = {
     UNMOD: '/unmod',
     TIMEOUT: '/timeout',
     PERMIT: '!permit',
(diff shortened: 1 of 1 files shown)


night:
it looks like ignore/unignore still works?

Nareese:
Hmmm.  That's possible, it didn't, for a long time (for a long time, there were noworking slash commands to manage blocklist entries at all, after they removed /ignore and before they added /block).  It's possible they added support for the old commands back.  Thanks for the merge anyway, I'm pretty sure /block|/unblock are intended to be the "official" commands, since they are documented by Twitch's /help system.

PR #4129, opened 2020-09-03
Setting to hide "similar" channels

### LINKED ISSUE #4126: Suggestion: Hide "This streamers viewers also watches:" recommendations
In the sidebar there is a " _streamerName_ s viewers also watch:"  and then it shows some streamers. 
I think there should be a hide option for this like the "Hide recommended streams in sidebar" option.

cadewey:
These are the channels shown in the section "<X> Viewers Also Watched".

Fixes #4126.

### DIFF
--- a/src/modules/hide_sidebar_elements/index.js
+++ b/src/modules/hide_sidebar_elements/index.js
@@ -4,6 +4,7 @@ const settings = require('../../settings');
 const domObserver = require('../../observers/dom');
 
 let removeFeaturedChannelsListener;
+let removeSimilarChannelsListener;
 let removeOfflineFollowedChannelsListener;
 
 class HideSidebarElementsModule {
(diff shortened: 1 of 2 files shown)


cadewey:
I'm not sold on the naming I used here for the setting/description, so open to suggestions there. The page uses the terminology "similarity channel" which I don't think it necessarily intuitive to a user, but the current text is maybe a little wordy.

night:
rather than add an additional setting, lets throw this as something under the "recommended channels" since this is still technically a kind of recommendation

cadewey:
> rather than add an additional setting, lets throw this as something under the "recommended channels" since this is still technically a kind of recommendation

I would personally really like the ability to toggle these separately. The original "recommended channels" section gives me stuff that's actually pretty useful in a lot of cases. This new one seems really bad, frequently trying to link me out to channels I'm already following (I think this will be a common complaint for people who follow multiple streamers who are friends/work together/etc).

night:
closed via 6e55ec56778dbca201d5c0acec1bf6a2635fe078

PR #4171, opened 2020-10-04
[Feature] Add theatre mode in fullscreen button

karlito40:
Hello,

I've add a button in the video player to enable the "theatre mode" in fullscreen. I think it could be useful for people. At least it is for me. 

Here is a little preview:

theater-in-fullscreen-compress

I couldn't make a more relevant icon though as i don't have access to illustrator.

### DIFF
--- a/src/modules/video_player/index.js
+++ b/src/modules/video_player/index.js
@@ -8,6 +8,7 @@ const debounce = require('lodash.debounce');
 const VIDEO_PLAYER_SELECTOR = '.video-player__container';
 const CANCEL_VOD_RECOMMENDATION_SELECTOR = '.recommendations-overlay .pl-rec__cancel.pl-button, .autoplay-vod__content-container button';
 const BTTV_PICTURE_IN_PICTURE_SELECTOR = '#bttv-picture-in-picture';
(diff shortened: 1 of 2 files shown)


night:
circling back to this now. is it possible to use twitch's theater mode icon, and then only enable this icon when already in fullscreen?

the functionality could be:

1. enter fullscreen
2. enable theater mode, which re-requests full screen with chat
3. clicking theater mode icon again re-requests full screen without chat

what do you think?

karlito40:
It was my first approach. I think it's the cleanest solution but there is a problem with that:

- a video already in full screen goes black when you request a full screen on an other element

I have no idea how to fix that at this time. I tried to exit and request another full screen operation but it's forbidden as this need to be trigger by a user interaction. 

If you want to see that in action: go to twitch, open a stream in full screen and type `document.body.requestFullscreen()`  in your devtool. You'll see that we have lost the video :(

Otherwise it will be a much better user experience than my current solution if we fix this issue

night:
thanks for the request. since this has gone stagnant I'm going to close this for now. we'd still be open to a pull request for this feature, but I don't like the idea of a different button for it in the video player (it just generates confusion for users)

PR #4233, opened 2020-11-14
Text only mode as a local chat filter

### LINKED ISSUE #144: Filter Emote Spam Messages
As the title says, I would like to be able to filter out all this muck. I like emotes, and removing all emotes would ruin the twitch experience a little too much, but I would like to be able to read the chat...

hawkins:
Hi there! I took a stab at re-implementing #3649 as a local chat filter to fix #144 .

I'm a first time user and contributor to BetterTTV for the purpose of this feature though, so I'm not really sure how to go about identifying whether a message is a text-only message or not if we implement it as a local chat filter. I'm hoping this draft PR can serve as a conversation starter where I can show some progress on the feature and ask for help on how I could identify whether a message is text-only when implementing a local chat filter.

As of opening this, it's pretty poorly functional. Some emotes are allowed, but some are not, and those that are not are not so cleanly removed (the username remains but the message does not):
image

I'm not thrilled with putting the bulk of the work in the `messageReplacer` function either, because it feels out of place since similar filters are really implemented in `messageParser`. The trouble with this though is that I could not spot a way to determine if a message was text-only at that point, but I noticed some code dealing with emotes in `messageReplacer`, so I piggy-backed off that just for a starting point.

Would love any thoughts or advice on this while I explore more of the codebase to see if anything pops out at me.

### DIFF
--- a/src/modules/chat/index.js
+++ b/src/modules/chat/index.js
@@ -44,12 +44,29 @@ const globalBots = ['nightbot', 'moobot'];
 let channelBots = [];
 let asciiOnly = false;
 let subsOnly = false;
+let textOnly = false;
 let modsOnly = false;
 
 function hasNonASCII(message) {
     for (let i = 0; i < message.length; i++) {
         if (message.charCodeAt(i) > 128) return true;
     }
+    return false;
+}
+
+function hasNonText(message) {
+    // TODO: How do we determine if the message is not plain text?
+    /*
+    previous PR did this in src/modules/chat_highlight_blacklist_keywords/index.js :
+
+        function (ast) {
+            return ast.some(node => {
+            return (node.type === 0 && node.content.trim().length > 0) || node.type === 5;
+            });
+        }
+
+    */
+    if (message) {}
     return false;
 }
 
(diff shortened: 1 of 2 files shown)


night:
Is the goal only to remove non-text, or do you want to present text representative versions of emotes?

If the former is the goal, simply hiding non message text spans will likely accomplish the goal. If the latter is the goal, you will need to iterative over all of the nodes to find their text content, and then return the text content as the new node (which would need to be done within the replacer).

hawkins:
The goal for me is to remove non-text. I'll have to play with it some more, because I tried to just remove the message node when a message is found to have an emote, but it's not behaving how I expect, so I'm doing something wrong still.

mpratt14:
@hawkins also there is a difference between removing messages that have non-text vs removing non-text from messages, where the latter is much more difficult...

bumbeen:
very interested in this!

hawkins:
I’m still interested in the feature as well, but I couldn’t manage to get the element selection down right. I think I’ll close this since it looks like I won’t be able to carry it though. Sorry!

PR #4250, opened 2020-11-29
[Fix] Hide collapsed sidebar recommended channels

### LINKED ISSUE #4183: Hide Recommended Channels not working.
Title. BTTV 7.2.89 - Chrome 86.0.4240.75.

v3g5gfA

ipiv:
The issue I noticed was that the previously used "data-test-selector" was missing on the link element in a case where the user was logged out of Twitch and sidebar was collapsed. 
Added support for "aria-label" selector to overcome this.

Fixes #4183

### DIFF
--- a/src/modules/hide_sidebar_elements/index.js
+++ b/src/modules/hide_sidebar_elements/index.js
@@ -48,7 +48,7 @@ class HideSidebarElementsModule {
         if (settings.get('hideFeaturedChannels')) {
             if (removeFeaturedChannelsListener) return;
 
-            removeFeaturedChannelsListener = domObserver.on('.side-nav-section a[data-test-selector="recommended-channel"], .side-nav-section a[data-test-selector="similarity-channel"], .side-nav-section .tw-svg__asset--navchannels', (node, isConnected) => {
+            removeFeaturedChannelsListener = domObserver.on('.side-nav-section a[data-test-selector="recommended-channel"], .side-nav-section a[data-test-selector="similarity-channel"], .side-nav-section .tw-svg__asset--navchannels, div[aria-label="Recommended Channels"]', (node, isConnected) => {
                 if (!isConnected) return;
                 $(node).addClass('bttv-hide-featured-channels');
             }, {useParentNode: true});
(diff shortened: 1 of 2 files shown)


ipiv:
On a second thought we don't actually need support for aria-label selector since we are using class selector with aria-label attribute.

Will revert the commit and suggest a change that allows selection of direct node with attribute.

src/observers/dom.js
```diff
@@ -63,7 +71,7 @@ function processObservedResults(emitter, node, results) {
 
     for (const observedType of results) {
         const {partialSelector, selector, options} = observedType;
-        const foundNode = partialSelector.includes(' ') ? node.querySelector(selector) : node;
+        const foundNode = partialSelector.split('[')[0].includes(' ') ? node.querySelector(selector) : node;
```
ipiv, line 74: Aria labels can have spaces

src/modules/hide_sidebar_elements/index.js
```diff
@@ -48,7 +48,7 @@ class HideSidebarElementsModule {
         if (settings.get('hideFeaturedChannels')) {
             if (removeFeaturedChannelsListener) return;
 
-            removeFeaturedChannelsListener = domObserver.on('.side-nav-section a[data-test-selector="recommended-channel"], .side-nav-section a[data-test-selector="similarity-channel"], .side-nav-section .tw-svg__asset--navchannels', (node, isConnected) => {
+            removeFeaturedChannelsListener = domObserver.on('.side-nav-section a[data-test-selector="recommended-channel"], .side-nav-section a[data-test-selector="similarity-channel"], .side-nav-section .tw-svg__asset--navchannels, div[aria-label="Recommended Channels"]', (node, isConnected) => {
```
night, line 51: how does this handle i18n?

ipiv: > how does this handle i18n?

right... 🤦 if they end up removing data-test-selectors, targetable classes etc. Maybe one possibility would be to search for react props/keys and identify nodes to hide that way.

PR #4270, opened 2020-12-10
[Feature] add claim a bonus watcher

yordis:
## New Feature

- Claim bonuses automatically

Screen Shot 2020-12-10 at 12 21 25 AM


night:
this already exists: https://github.com/night/betterttv/blob/master/src/modules/channel_points/index.js

PR #4286, opened 2020-12-16
[Feature] Add setting to blacklist / highlight all mentions / links

### LINKED ISSUE #4272: How to highlight all @
I was just wondering if there was a way to highlight all mentions, any chat line with the @ symbol in it

mpratt14:
Closes #4272

Thanks to @Nareese for guidance

Signed-off-by: Michael Pratt <mcpratt@pm.me>

### DIFF
--- a/src/modules/chat_highlight_blacklist_keywords/index.js
+++ b/src/modules/chat_highlight_blacklist_keywords/index.js
@@ -14,10 +14,10 @@ const REPEATING_SPACE_REGEX = /\s\s+/g;
 
 const BLACKLIST_KEYWORD_PROMPT = `Type some blacklist keywords. Messages containing keywords will be filtered from your chat.
 
-Use spaces in the field to specify multiple keywords. Place {} around a set of words to form a phrase, <> inside the {} to use exact search, and () around a single word to specify a username. Wildcards (*) are supported.`;
+Use spaces in the field to specify multiple keywords. Place {} around a set of words to form a phrase, <> inside the {} to use exact search, and () around a single word to specify a username. '$@' for all mentions, '$#' for all emotes, 'http*' for all links. Wildcards (*) are supported.`;
 const HIGHLIGHT_KEYWORD_PROMPT = `Type some highlight keywords. Messages containing keywords will turn red to get your attention.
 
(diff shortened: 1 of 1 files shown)


src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -198,6 +202,13 @@ class ChatHighlightBlacklistKeywordsModule {
             defaultValue: false
         });
 
+        settings.add({
```
night, line 205: this is not something we should make a separate setting. instead, it should be made possible to add a match for this in the existing highlights system

mpratt14:
@night ok, completely new strategy, also expanded for links and blacklisting

mpratt14:
@night I think it is unlikely enough for anyone to type "$@" or "$#" in any chat message so its sufficient...but maybe you can think of a better way...

mpratt14:
combined the if statements so the sound and pinning is included

mpratt14:
@night looks good now? or...

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -14,10 +14,10 @@ const REPEATING_SPACE_REGEX = /\s\s+/g;
 
 const BLACKLIST_KEYWORD_PROMPT = `Type some blacklist keywords. Messages containing keywords will be filtered from your chat.
 
-Use spaces in the field to specify multiple keywords. Place {} around a set of words to form a phrase, <> inside the {} to use exact search, and () around a single word to specify a username. Wildcards (*) are supported.`;
+Use spaces in the field to specify multiple keywords. Place {} around a set of words to form a phrase, <> inside the {} to use exact search, and () around a single word to specify a username. '$@' for all mentions, '$#' for all links. Wildcards (*) are supported.`;
```
night, line 17: is there a reason not to combine these with wildcards. could we instead do `https://*` for all links, and `@*` for all mentions?

mpratt14: ok I realize now that `http*` works, however I know that `@*` does not work because this filtering is done after the text is parsed into the span elements (text, mention, link, emote)

mpratt14:
updated so now `$@` for mentions and `$#` for emotes

night:
Apologies for the delays. We've recently rebuilt the settings to improve UX around managing highlights/blacklist. Are you able to update this PR using the new system? we should be able to add new types of filters now

night:
closing due to inactivity

PR #4291, opened 2020-12-18
Adds a setting for Custom Timeouts

ipiv:
Adds a setting for right click custom timeout slider. (Enabled by default)


ipiv:
Still considered helpful by multiple moderators.

prdatur:
Need this too.

night:
Thanks for the pull request. Since this has gone stagnant we are going to close this. We'd be open to accepting this request if implemented as part of our new settings system we shipped last year. This change looks to leverage our old settings system which is long gone.

PR #4320, opened 2021-01-14
Host Channel button changes

Wengiel31:
Changed Host Channel button width and changed it's content from "Host Channel" to "Host This Channel" in order for it not to look weird


night:
after reviewing the PR's changes, I do not think either is required. "This" is implied, and the button width looks to be consistent across any screen size from my testing

PR #4403, opened 2021-03-12
Fix twitch.getChatModeratorCardProps

### LINKED ISSUE #4400: Rename button(pencil icon) don't show
Looks like display cards got a layout update. But currently the rename button don't show.
Screenshot_3

### LINKED ISSUE #4402: Account creation date
I don't see account creation dates 
(i m not a moderator)
https://i.imgur.com/vuNv9io.png

Nareese:
fixes #4400, fixes #4402

Twitch changed their modcard layout, and it broke twitch.getChatModeratorCardProps.  Fix searches Children instead of Parents, and overrides maxDepth = 30 (node is currently located at depth 23).

### DIFF
--- a/src/utils/twitch.js
+++ b/src/utils/twitch.js
@@ -364,9 +364,10 @@ module.exports = {
     getChatModeratorCardProps(element) {
         let apolloComponent;
         try {
-            const node = searchReactParents(
+            const node = searchReactChildren(
                 getReactInstance(element),
-                n => n.stateNode && n.stateNode.props && n.stateNode.props.data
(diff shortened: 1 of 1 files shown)


Nareese:
Hold on this for now, this appears to be a valid fix on a channel where I have moderation rights, but appears not to work on a channel where I don't.  Continuing to investigate...

Nareese:
Closing this for now... this is not effective for non-moderators.  Cannot find the props node to a depth of 50 in either direction (Parents finds a node at 28, but it can't get targetUser from it, Children finds nothing).  Not sure how to proceed without that data, since everything in the modcard module depends on it, so... giving up. :(

PR #4425, opened 2021-03-22
[Bug Fix] Hide Recommended Channels Again

### LINKED ISSUE #4183: Hide Recommended Channels not working.

Aaron128l:
Changed the search to react rather than CSS as Twitch classes all three sections (Followed, Recommended, Friends) the same. This instead looks through all sidebar content for the `type: RECOMMENDED_SECTION` and marks that node to be hidden. Works on page load for both the full and collapsed sidebar and toggles correctly.

Fixes #4183


Aaron128l:
As a reminder this issue only occurs on the collapsed sidebar on a page load.

PR #4513, opened 2021-05-10
[Feature] Add setting to hide native twitch emotes

thesecretmaster:
A friend recently mentioned to me that certain twitch native emotes bothered them to see, this patch would add a setting to show certain native twitch emotes as text instead of images. This felt like it fit well with the blacklist/highlight module, so I put it there. I'm happy to updating the module naming to include this feature or try to build it into a new module if needed. I tried my best to match the code style although I'm still very new to this codebase, and additionally I'm not super familiar with jQuery, here's my vanilla JS implimenation of `nativeEmoteScrubber`, in case that's preferred:

```javascript
  nativeEmoteScrubber(ele, emotesToScrub) {
    const emotes = ele[0].querySelector("[data-test-selector=chat-line-message-body]").querySelectorAll("[data-test-selector=emote-button]")
    Array.prototype.map.call(emotes, emote => {
      const emote_name = emote.getElementsByTagName("img")[0].alt
      if (emotesToScrub.includes(emote_name)) {
        const new_text_frag = document.createElement("span");
        new_text_frag.classList.add("text-fragment")
        new_text_frag.dataset.ATarget = "chat-message-text"
        new_text_frag.style.backgroundColor = "rgba(255,255,0,0.2)"
        new_text_frag.textContent = emote_name
        emote.parentElement.replaceChild(new_text_frag, emote)
      }
    })
  }
```

### DIFF
--- a/src/modules/chat_highlight_blacklist_keywords/index.js
+++ b/src/modules/chat_highlight_blacklist_keywords/index.js
@@ -15,6 +15,9 @@ const REPEATING_SPACE_REGEX = /\s\s+/g;
 const BLACKLIST_KEYWORD_PROMPT = `Type some blacklist keywords. Messages containing keywords will be filtered from your chat.
 
 Use spaces in the field to specify multiple keywords. Place {} around a set of words to form a phrase, <> inside the {} to use exact search, and () around a single word to specify a username. Wildcards (*) are supported.`;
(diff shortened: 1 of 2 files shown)


night:
Apologies for the delays. This looks to duplicate some of the functionality proposed in PR #4286

PR #4600, opened 2021-06-09
[FEATURE] Add module HideChatCommands

Yahatix:
Hide chat messages starting with "!" for a cleaner chat


night:
Thanks for the PR, but I believe this can be implemented using the existing blacklist system already with wildcards.

PR #4616, opened 2021-06-16
[Feature] Add Auto Theme Setting

philipsdean:
Adds at a setting to enable matching Twitch's theme with the system's theme automatically.


night:
Apologies for the delays. Can you update this PR for the new settings menu we launched? I think we can accept it then.

src/modules/global_css/index.js
```diff
@@ -37,6 +40,16 @@ class GlobalCSSModule {
     });
   }
 
+  setAutoTheme(value) {
```
night, line 43: does the setting actually work? on initial loads this function does not check the value of the setting, meaning it will always add a listener and update the theme for browser color scheme on page load

philipsdean: No it didn't work as it should. Since it's been awhile, i've just rewrote and reimplemented it for the most recent master.

Issue #4676, opened 2021-07-18
Suggestion: Feature to opt-out of auto joining a raid


SpottedGiraffes:
This was posted 3 years ago with no update.  So I figured I'll try reposting to see if this gets any traction.  Super annoying feature that auto opts you in to joining a raid (as a viewer) unless you click a prompt.  If you are tabbed somewhere else, away from the computer at the time, or have fallen asleep to a chill streamer you like this can be very annoying.  Please help.

night:
duplicate of #2936

PR #4683, opened 2021-07-19
[Feature] Add auto join raid setting and module

### LINKED ISSUE #2936: Suggestion: Feature to opt-out of auto joining a raid

bkmak:
Adds the ability to auto leave a raid, so raiding will only occur if the user opts-in by selecting "Join".

Fixes #2936 and #4676

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -161,6 +161,7 @@ export const EmoteMenuTips = {
 export const SettingDefaultValues = {
   [SettingIds.ANON_CHAT]: false,
   [SettingIds.AUTO_THEATRE_MODE]: false,
+  [SettingIds.AUTO_JOIN_RAIDS]: true,
   [SettingIds.SPLIT_CHAT]: false,
   [SettingIds.DELETED_MESSAGES]: DeletedMessageTypes.DEFAULT,
   [SettingIds.REVERSE_CHAT_DIRECTION]: false,
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,6 +1,7 @@
 export const SettingIds = {
   ANON_CHAT: 'anonChat',
   AUTO_THEATRE_MODE: 'autoTheatreMode',
+  AUTO_LEAVE_RAIDS: 'autoLeaveRaids',
   CHANNEL_POINTS: 'channelPoints',
   SPLIT_CHAT: 'splitChat',
   REVERSE_CHAT_DIRECTION: 'reverseChatDirection',
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,7 +1,7 @@
 export const SettingIds = {
   ANON_CHAT: 'anonChat',
   AUTO_THEATRE_MODE: 'autoTheatreMode',
-  AUTO_LEAVE_RAIDS: 'autoLeaveRaids',
+  AUTO_JOIN_RAIDS: 'autoJoinRaids',
   CHANNEL_POINTS: 'channelPoints',
   SPLIT_CHAT: 'splitChat',
   REVERSE_CHAT_DIRECTION: 'reverseChatDirection',
--- a/src/modules/auto_leave_raids/index.js
+++ b/src/modules/auto_leave_raids/index.js
@@ -1,0 +1,45 @@
+import $ from 'jquery';
+import watcher from '../../watcher.js';
+import settings from '../../settings.js';
+import domObserver from '../../observers/dom.js';
+import { SettingIds } from '../../constants.js';
+
+const RAID_BANNER_SELECTOR = '[data-test-selector="raid-banner"]';
+
+let raidListener;
+
+class AutoLeaveRaidsModule {
+    constructor() {
+        watcher.on('load.chat', () => this.load());
+        settings.on(`changed.${SettingIds.AUTO_LEAVE_RAIDS}`, () => this.load());
+    }
+
+    load() {
+        if (settings.get(SettingIds.AUTO_LEAVE_RAIDS)) {
+            if (raidListener) return;
+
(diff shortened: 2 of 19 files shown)


src/constants.js
```diff
@@ -35,6 +36,7 @@ export const SettingIds = {
 export const LegacySettingIds = {
   ANON_CHAT: 'anonChat',
   AUTO_THEATRE_MODE: 'autoTheatreMode',
+  AUTO_LEAVE_RAIDS: 'autoLeaveRaids',
```
dclstn, line 39: There's no need to update the `LegacySettingIds`, this simply contains all the pre-existing settings that weren't stored in the settings dict.

src/modules/auto_leave_raids/index.js
```diff
@@ -0,0 +1,45 @@
...
+    if (!raidListener) return;
+
+    raidListener();
+    raidListener = undefined;
+  }
+
+  handleLeave(node, isConnected) {
+    if (!isConnected) return;
+
+    if (settings.get(SettingIds.AUTO_LEAVE_RAIDS)) {
```
dclstn, line 35: Assuming if the listener is attached this setting would be enabled anyway, is this lookup necessary?

src/modules/auto_leave_raids/index.js
```diff
@@ -0,0 +1,45 @@
+import $ from 'jquery';
```
dclstn, line 1: Night is looking into removing jquery, can we try converting this file to vanilla js?

src/constants.js
```diff
@@ -152,6 +154,7 @@ export const PageTypes = {
 export const DefaultValues = {
   [SettingIds.ANON_CHAT]: false,
   [SettingIds.AUTO_THEATRE_MODE]: false,
+  [SettingIds.AUTO_LEAVE_RAIDS]: false,
```
dclstn, line 157: It may be more accessible to have this default on and only when this user untoggles this setting they will be opted out of raids. e.g. `[SettingIds.AUTO_JOIN_RAIDS]: true`

night:
@bkmak Thanks for the code sample demonstrating this issue. It looks like we're using the wrong reference when creating the listener. fixed via https://github.com/night/betterttv/commit/cad9da4def344246e7c0708d4c3d59dc623a7a3c

src/constants.js
```diff
@@ -160,6 +161,7 @@ export const EmoteMenuTips = {
 
 export const SettingDefaultValues = {
   [SettingIds.ANON_CHAT]: false,
+  [SettingIds.AUTO_JOIN_RAIDS]: true,
```
night, line 164: we should not perform automatic actions by default

src/modules/settings/components/settings/AutoJoinRaids.jsx
```diff
@@ -0,0 +1,26 @@
...
+import {SettingIds, CategoryTypes} from '../../../../constants.js';
+import styles from '../../styles/header.module.css';
+
+function AutoJoinRaids() {
+  const [value, setValue] = useStorageState(SettingIds.AUTO_JOIN_RAIDS);
+
+  return (
+    <Panel header="Auto Join Raids">
+      <div className={styles.toggle}>
+        <p className={styles.description}>Untoggle to be opted out of raids.</p>
```
night, line 14: Can we update this to describe the enabled state rather than the disabled state?

src/constants.js
bkmak: Isn't the default Twitch behavior to automatically join you to a raid? If so, should we then change this back to be an "Auto Leave" setting with the default to off? Seems like most if not all BTTV settings represent default Twitch behavior when set to off.

night: you're right. i read this as inverted. this looks correct

bkmak: @night Just to clarify, you believe it is correct to leave this as "Auto Join Raids" with a default of true, instead of "Auto Leave Raids" with a default of false?

src/modules/settings/components/settings/AutoJoinRaids.jsx
eldorel:
```suggestion
        <p className={styles.description}>Joins channel raids automatically without user interaction </p>
```
Unless I misunderstand something, this text edit is the last item blocking this merge. Is that correct?

bkmak: @eldorel Thanks for the reminder. I have updated the description, removing the "without user interaction" ending as it seemed redundant with "automatically".

totof99:
I rebased the fix and it seems to work OK.
It can be found here https://github.com/totof99/betterttv/tree/auto-leave-raids
@night Should I do a new pull request ?

bkmak:
Do raids even work the same way as when I first opened this pull request? These days, Twitch makes me opt-in to join a raid, defeating the purpose of the change I made. Is it the same way for everyone else?

Yrlish:
> Do raids even work the same way as when I first opened this pull request? These days, Twitch makes me opt-in to join a raid. I originally made this pull request because raids used to require me to opt-out. Is it the same way for everyone else?

My experience is that they are opt-out.

totof99:
For me, raid are opt-out. So this feature is still required.

PR #4685, opened 2021-07-20
modal fixes

dclstn:
- fix to opening window whilst in theatre mode
- go full menu in ANY window
- only update default keywords once
- fix legacy backups

### DIFF
--- a/src/modules/chat_highlight_blacklist_keywords/index.js
+++ b/src/modules/chat_highlight_blacklist_keywords/index.js
@@ -150,55 +150,27 @@ class ChatHighlightBlacklistKeywordsModule {
 
     this.updateDefaultHighlightKeywords(highlightKeywords);
 
-    let shouldUpdateHighlightKeywords = false;
-    let shouldUpdateBlacklistKeywords = false;
+    for (const keywordsValue of [highlightKeywordsValue, blacklistKeywordsValue]) {
+      let updated = false;
 
-    const validatedHighlightKeywordsValue = Object.fromEntries(
-      Object.entries(highlightKeywordsValue)
-        .map(([key, value]) => {
-          switch (value.type) {
-            case KeywordTypes.EXACT:
-              shouldUpdateHighlightKeywords = true;
-              return [key, {...value, keyword: `<${value.keyword}>`, type: KeywordTypes.MESSAGE}];
-            case KeywordTypes.WILDCARD:
-              shouldUpdateHighlightKeywords = true;
-              return [key, {...value, type: KeywordTypes.MESSAGE}];
-            default:
-              return [key, value];
-          }
-        })
-        .filter(([, {keyword}]) => keyword.replace(/(\*|<|>)/g, '').trim())
-    );
+      for (const value of Object.values(keywordsValue)) {
+        if (![KeywordTypes.EXACT, KeywordTypes.WILDCARD].includes(value.type)) {
+          continue;
+        }
 
-    const validatedBlacklistKeywordsValue = Object.fromEntries(
-      Object.entries(blacklistKeywordsValue)
-        .map(([key, value]) => {
-          switch (value.type) {
-            case KeywordTypes.EXACT:
-              shouldUpdateBlacklistKeywords = true;
-              return [key, {...value, keyword: `<${value.keyword}>`, type: KeywordTypes.MESSAGE}];
-            case KeywordTypes.WILDCARD:
-              shouldUpdateBlacklistKeywords = true;
-              return [key, {...value, type: KeywordTypes.MESSAGE}];
-            default:
-              return [key, value];
-          }
-        })
-        .filter(([, {keyword}]) => keyword.replace(/(\*|<|>)/g, '').trim())
-    );
+        if (value.type === KeywordTypes.EXACT) {
+          value.keyword = `<${value.keyword}>`;
+        }
+        value.type = KeywordTypes.MESSAGE;
+        updated = true;
+      }
 
-    if (
-      shouldUpdateHighlightKeywords ||
-      Object.values(highlightKeywordsValue).length !== Object.values(validatedHighlightKeywordsValue).length
-    ) {
-      settings.set(SettingIds.HIGHLIGHT_KEYWORDS, validatedHighlightKeywordsValue);
-    }
-
-    if (
-      shouldUpdateBlacklistKeywords ||
-      !Object.values(blacklistKeywordsValue).length !== Object.values(validatedBlacklistKeywordsValue).length
-    ) {
-      settings.set(SettingIds.BLACKLIST_KEYWORDS, validatedBlacklistKeywordsValue);
+      if (updated) {
+        settings.set(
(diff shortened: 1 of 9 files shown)


src/modules/settings/components/Window.jsx
```diff
@@ -28,7 +27,7 @@ function Page(props) {
   }
 }
 
-const validateModal = () => window.innerWidth < 600 && isPopout();
```
night, line 31: i think we should keep the popout, but also allow it if it's a chat page.. maybe check `window.location.pathname.endsWith('/chat')`

src/modules/settings/pages/About.jsx
```diff
@@ -55,6 +55,7 @@ function About({onHide}) {
       const settingsToImport = isJSON(data);
       if (!settingsToImport) return;
       Object.keys(settingsToImport).forEach((s) => storage.set(s.split('bttv_')[1], settingsToImport[s]));
+      if (!Object.keys(settingsToImport).includes('bttv_settings')) settings.importLegacySettings();
```
night, line 58: why do we import all the settings, and then import legacy? why are we calling `Object.keys` again?

src/utils/keywords.js
```diff
@@ -85,36 +79,21 @@ export function serializeKeywords(keywords) {
...
             id: index++,
             keyword,
-            type: Types.MESSAGE,
-          };
-      }
-    })
+            type: KeywordTypes.MESSAGE,
+          }
+    )
     .filter((string) => string !== false);
```
night, line 91: instead:

```
const keywordString = computedKeywords
  .filter(string => string.length > 0)
  .map(keyword => ({id: index++, keyword, type: KeywordTypes.MESSAGE}))
```

src/utils/keywords.js
```diff
@@ -85,36 +79,21 @@ export function serializeKeywords(keywords) {
...
 
   const usersString = computedUsers.map((user) => ({
```
night, line 93: `keywordString` and `usersString` make no sense. these are arrays?

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -144,20 +144,81 @@ class ChatHighlightBlacklistKeywordsModule {
...
-        id: 0,
-        type: Types.MESSAGE,
-        status: null,
-        keyword: user.name,
-      },
-    });
+
+    settings.set(
+      SettingIds.HIGHLIGHT_KEYWORDS,
+      user != null
```
night, line 211: we should not be setting a default value in storage if user is not logged in.

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -144,20 +144,81 @@ class ChatHighlightBlacklistKeywordsModule {
     this.handleHighlightSound = this.handleHighlightSound.bind(this);
   }
 
-  updateDefaultHighlightKeywords() {
-    const value = settings.get(SettingIds.HIGHLIGHT_KEYWORDS);
-    if (Object.keys(value).length > 0) return;
+  validateKeywords() {
```
night, line 147: what you're doing is called "read repair". rename this to `readRepairKeywords`

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -144,20 +144,81 @@ class ChatHighlightBlacklistKeywordsModule {
...
+    const highlightKeywordsValue = settings.get(SettingIds.HIGHLIGHT_KEYWORDS);
+    const blacklistKeywordsValue = settings.get(SettingIds.BLACKLIST_KEYWORDS);
+
+    this.updateDefaultHighlightKeywords(highlightKeywords);
+
+    let shouldUpdateHighlightKeywords = false;
+    let shouldUpdateBlacklistKeywords = false;
+
+    const validatedHighlightKeywordsValue = Object.fromEntries(
+      Object.entries(highlightKeywordsValue)
```
night, line 157: we probably don't need to complexity of entries here, and your "when to persist" logic can actually be simplified since we know it needs to persist iff legacy types are migrated.


```
for (const keywordsValue of [highlightKeywordsValue, blacklistKeywordsValue]) {
  let updated = false;

  for (const value of Object.values(keywordsValue)) {
    if (![KeywordTypes.EXACT, KeywordTypes.WILDCARD].includes(value.type)) {
      continue;
    }

    if (value.type === KeywordTypes.EXACT) {
      value.keyword = `<${value.keyword}>`;
    }
    value.type = KeywordTypes.MESSAGE;
    updated = true;
  }

  if (updated) {
    settings.set(keywordsValue === highlightKeywordsValue ? SettingIds.HIGHLIGHT_KEYWORDS : SettingIds.BLACKLIST_KEYWORDS, keywordsValue);
  }
}

```

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -144,20 +144,52 @@ class ChatHighlightBlacklistKeywordsModule {
...
+        );
+      }
+    }
+  }
+
+  updateDefaultHighlightKeywords(value) {
+    if (value != null) return null;
     const user = twitch.getCurrentUser();
-    if (user == null) return;
+    if (user === null) return null;
```
night, line 180: why is this doing an exact match on null? `==` handles null + undefined

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -144,20 +144,52 @@ class ChatHighlightBlacklistKeywordsModule {
...
+        settings.set(
+          keywordsValue === highlightKeywordsValue ? SettingIds.HIGHLIGHT_KEYWORDS : SettingIds.BLACKLIST_KEYWORDS,
+          keywordsValue
+        );
+      }
+    }
+  }
+
+  updateDefaultHighlightKeywords(value) {
+    if (value != null) return null;
```
night, line 178: shouldn't this return `value` if it's not null?

src/modules/chat_highlight_blacklist_keywords/index.js
```diff
@@ -144,20 +144,52 @@ class ChatHighlightBlacklistKeywordsModule {
...
     const user = twitch.getCurrentUser();
-    if (user == null) return;
+    if (user == null) return null;
```
night, line 180: should this return empty object if it's not null?

dclstn: `  updateDefaultHighlightKeywords() {
    const value = settings.get(SettingIds.HIGHLIGHT_KEYWORDS);
    if (value != null) return;
    const user = twitch.getCurrentUser();
    if (user == null) return;

    settings.set(SettingIds.HIGHLIGHT_KEYWORDS, {
      0: {
        id: 0,
        type: KeywordTypes.MESSAGE,
        status: null,
        keyword: user.name,
      },
    });
  }`

I think alternatively we should keep the function like so, we shouldn't build it around another function. granted this is ineffecient on the first time a logged user uses the extension but afterwards there won't be as many reads and writes

PR #4718, opened 2021-08-03
[Feature] New Emoji-Menu

dclstn:
PR adds a new re-design to the outdated forked emote menu.

image

New Features:

- Virtualized scrolling to improve performance.
- Improved navigation with fuzzy search.
- The ability to favorite emotes and have them pinned
- See frequently/recently used emotes
- And an overall fluid design that's easy to navigate


src/modules/emote_menu/styles/menu.module.css
```diff
@@ -0,0 +1,75 @@
...
+}
+
+.header {
+  display: flex;
+  padding: 7.5px;
+}
+
+.content {
+  display: flex;
+  height: calc(360px - 52px);
```
night, line 12: we should find a way to do this without calc. you can make use of margin and padding instead.

src/modules/emotes/index.js
```diff
@@ -56,6 +56,14 @@ class EmotesModule {
 
     return null;
   }
+
+  get length() {
+    return this.emoteProviders.reduce((accumulator, {emotes}) => accumulator + emotes.size, 0);
```
night, line 61: it would be great to avoid recomputing this for every retrieval. is there a way to mark the store as dirty and recompute it one time?

src/modules/emote_menu/style.css
```diff
@@ -0,0 +1,13 @@
+#bttvEmoteMenuContainer {
```
night, line 1: thoughts on using css modules for this instead?

src/modules/emote_menu/stores/favourite-emotes.js
```diff
@@ -0,0 +1 @@
+// WIP
```
night, line 1: nit: prefer `favorite`. i know it's different in the UK

webpack.config.js
```diff
@@ -57,6 +57,20 @@ function jsonTransform(emojis) {
...
+function jsonTransformCategories(emojis) {
+  const result = {};
+  for (const emojiData of Object.values(emojis)) {
+    const char = convertEmojiToolkitCodePointToChar(emojiData.code_points.fully_qualified);
+    const data = {
+      char,
+      slug: emojiData.shortname.replace(/:/g, ''),
+    };
+
+    result[emojiData.category] = {...result[emojiData.category], [data.slug]: data};
```
night, line 69: it would be great to avoid iterating over the same dataset twice. can we re-use the same loop to generated these both?

src/modules/emote_menu/components/Button.jsx
```diff
@@ -0,0 +1,27 @@
...
+      trigger="click"
+      placement="topEnd"
+      speaker={
+        <Popover className={styles.popover} full>
+          <EmoteMenu triggerRef={triggerRef} />
+        </Popover>
+      }
+      triggerRef={triggerRef}>
+      <button type="button" className={styles.button}>
+        <FontAwesomeIcon icon={faSmile} />
```
night, line 23: this icon doesn't appear as the same size as twitch's native one. im also curious if there's a way to make use of twitch's?

src/modules/emote_menu/search.js
```diff
@@ -0,0 +1,18 @@
...
+
+class EmoteSearch {
+  constructor() {
+    this.options = {
+      keys: ['code'],
+    };
+    this.search = new Fuse(emotes.getEmotes(), this.options);
+
+    watcher.on('channel.updated', () => {
+      this.search = new Fuse(emotes.getEmotes(), this.options);
```
night, line 13: rather than reinstantiating again, is there a way to update it with a new dataset? maybe `createIndex`?

i'm also curious how performance on this is. would it be more performant to remove emotes from the old channel and add emotes from the new channel?

src/modules/emote_menu/stores/emoji-categories.js
```diff
@@ -0,0 +1,125 @@
...
+  return count;
+}
+
+function parseEmotes(emojiBySlug, provider) {
+  return Object.values(emojiBySlug)
+    .filter((emoji) => emojiBlacklist.indexOf(emoji.char) === -1 && countEmojis(emoji) === 1)
+    .map((emoji, index) => {
+      let url;
+
+      twemoji.parse(emoji.char, {
```
night, line 23: we should avoid loading these twice, as it's a lot of additional parsing. is there a way to deduplicate this work?

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,83 @@
...
+    return this.rows.length;
+  }
+
+  loadProviders() {
+    this.providers = [
+      {
+        provider: {
+          id: 'bttv',
+          displayName: 'BetterTTV',
+          icon: Icons.THUMBS_UP,
```
night, line 38: are these icons are placeholder for actual logos?

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,83 @@
...
+  getRow(index) {
+    return this.rows[index];
+  }
+
+  getHeaders() {
+    return Object.values(this.headers);
+  }
+
+  isHeader(index) {
+    return index in this.headers;
```
night, line 79: use `Object.prototype.hasOwnProperty` instead

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,83 @@
...
+    this.headers = {};
+    this.rows = [];
+
+    for (const provider of this.providers) {
+      if (provider.emotes.length === 0) {
+        continue;
+      }
+
+      this.headers[this.totalRows] = provider.provider;
+      this.rows = [...this.rows, provider.provider, ...chunkArray(provider.emotes, COLOUMN_COUNT)];
```
night, line 64: what is the performance like reconstructing this array multiple times with spreading? have you considered concatenating arrays instead?

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+    this.emit('loaded');
+  }
+
+  getRow(index) {
+    return this.rows[index];
+  }
+
+  getHeaders() {
+    return Object.values(this.headers);
```
night, line 75: what is the performance like retrieving a new list of values on every call? should we store this reference instead?

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,83 @@
...
+  loadEmotes() {
+    this.headers = {};
+    this.rows = [];
+
+    for (const provider of this.providers) {
+      if (provider.emotes.length === 0) {
+        continue;
+      }
+
+      this.headers[this.totalRows] = provider.provider;
```
night, line 63: we seem to store header twice, once in the headers list and once in the rows. i'm curious, have you considered looking at the row's data to determine if it's a header instead? couldn't the row have a type?

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+const style = (index, rowHeight) => ({
+  position: 'absolute',
+  top: `${index * rowHeight}px`,
+  height: `${rowHeight}px`,
+  width: '100%',
+});
+
+const headerStyle = (rowHeight) => ({
+  position: 'sticky',
```
night, line 11: unless you need a value defined in js, css values could be stored in a css module rather than hardcoding them

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,76 @@
...
+    const {scrollTop} = wrapperRef.current;
+    const scrollBottom = scrollTop + windowHeight;
+
+    const startIndex = Math.floor(scrollTop / rowHeight);
+    const endIndex = Math.min(totalRows - 1, Math.floor(scrollBottom / rowHeight));
+
+    const rowsVisible = [];
+
+    for (let i = startIndex; i <= endIndex; i++) {
+      rowsVisible.push(i);
```
night, line 45: curious, why not use `slice` for this?

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,76 @@
...
+    return () => {
+      wrapperRef.current.removeEventListener('scroll', isInViewportListener, false);
+    };
+  }, [isInViewport]);
+
+  return (
+    <div {...restProps} style={{height: windowHeight, overflowY: 'scroll'}} ref={wrapperRef}>
+      <div style={{position: 'relative', height: listHeight}}>
+        {data.rows.map((value) => renderRow({key: `row-${value}`, index: value, style: style(value, rowHeight)}))}
+        {data.header != null &&
```
night, line 69: use a ternary, and if false return null.

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,76 @@
...
+    const newHeader = headers != null ? headers.filter((value) => value < startIndex + 1)[0] : null;
+
+    setData({
+      header: newHeader,
+      rows: rowsVisible,
+    });
+  }, [totalRows, rowHeight, windowHeight]);
+
+  useEffect(() => {
+    const isInViewportListener = isInViewport;
```
night, line 57: im not sure why you save this to a variable. isn't `isInViewport` equivalent?

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,76 @@
...
+    const startIndex = Math.floor(scrollTop / rowHeight);
+    const endIndex = Math.min(totalRows - 1, Math.floor(scrollBottom / rowHeight));
+
+    const rowsVisible = [];
+
+    for (let i = startIndex; i <= endIndex; i++) {
+      rowsVisible.push(i);
+    }
+
+    const newHeader = headers != null ? headers.filter((value) => value < startIndex + 1)[0] : null;
```
night, line 48: curious, why not use array `find`?

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,76 @@
...
+  height: `${rowHeight}px`,
+  width: '100%',
+});
+
+function VirtualizedList(
+  {totalRows, rowHeight, renderRow, windowHeight, stickyRows = [], onHeaderChange = () => {}, ...restProps},
+  ref = null
+) {
+  const listHeight = useMemo(() => rowHeight * totalRows, [totalRows, rowHeight]);
+  const headers = useMemo(() => stickyRows.sort((a, b) => b - a), [stickyRows]);
```
night, line 22: curious: if the rows are build in the same direction as settings, wouldn't the sort be identical?

src/modules/emote_menu/components/Preview.jsx
```diff
@@ -0,0 +1,45 @@
...
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <div {...restProps} key={emote.code}>
+      <div className={styles.content}>
+        <div className={styles.emoji}>
+          <img alt={emote.name} src={emote.images['3x'] || emote.images['2x'] || emote.images['1x']} />
```
night, line 29: we ideally do not want to to load sizes like this. instead, you can make use of srcset which takes into account highdpi screens

src/modules/emote_menu/components/Preview.jsx
```diff
@@ -0,0 +1,45 @@
...
+        </div>
+        <div>
+          <p>{emote.code}</p>
```
night, line 32: nit: don't use `p` as it comes with default styles. use `div` and apply some styling instead

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,47 @@
...
+  const [search, setSearch] = useState('');
+  const [emote, setEmote] = useState(null);
+  const [focus, setFocus] = useState({
+    eventKey: emoteStore.getHeaders()[0].id,
+    scrollTo: false,
+  });
+
+  const onHide = () => triggerRef.current.close();
+
+  useEffect(() => setSearch(''), [focus]);
```
night, line 20: is this correct? why would focus clear your input?

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+  function handleClick(newEmote) {}
```
night, line 22: i assume this is a TODO which will insert into the message input?

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,89 @@
...
+  );
+
+  const headerChange = useCallback((index) => {
+    const header = emoteStore.headers[index];
+    if (header) onFocus(header.id);
+  });
+
+  useEffect(() => {
+    if (!focus.scrollTo) return;
+    const index = Object.keys(emoteStore.headers).find((key) => emoteStore.headers[key].id === focus.eventKey);
```
night, line 40: it is a bit of a code smell that we are considering headers both here and in the child component. is there a way to pass this information to the child and have it compute its own scroll position?

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,89 @@
...
+          {row.map((emote) => (
+            <Emote emote={emote} onClick={() => onClick(emote)} onMouseOver={() => onSelect(emote)} />
+          ))}
+        </div>
+      );
+    },
+    [onSelect, onClick]
+  );
+
+  const headerChange = useCallback((index) => {
```
night, line 33: nit: try to keep naming consistent. you seem to use `onHeaderChange`, but `headerChange` here and `onSelect` above. let's normalize on `onName` for props and `handleName` for functions we're passing in.

```
onHeaderChange={handleHeaderChange}
```

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,89 @@
...
+    const header = emoteStore.headers[index];
+    if (header) onFocus(header.id);
```
night, line 35: nit: prefer non-falsey and non-truthy code moving forward, as it leads to less bugs. `header != null`

additionally, always prefer brackets unless it's a bare control statement (return, continue, break, etc)

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  useEffect(() => {
+    if (!focus.scrollTo) return;
+    const index = Object.keys(emoteStore.headers).find((key) => emoteStore.headers[key].id === focus.eventKey);
+    if (index) wrapperRef.current.scrollTo(0, index * ROW_HEIGHT);
```
night, line 41: brackets

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,89 @@
...
+  const emotes = EmoteSearch.search.search(search);
+
+  const renderRow = useCallback(
+    ({key, style, index}) => {
+      const row = emotes.slice(index * TOTAL_COLS, (index + 1) * TOTAL_COLS).map(({item}) => [item.code, item]);
+
+      return (
+        <div key={key} style={style} className={styles.row}>
+          {row.map((emote) => (
+            <Emote emote={emote} onClick={() => onClick(emote)} onFocus={() => onSelect(emote)} />
```
night, line 68: does this need `key`?

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,89 @@
...
+    />
+  );
+}
+
+function SearchedEmotes({search, onSelect, onClick}) {
+  const emotes = EmoteSearch.search.search(search);
+
+  const renderRow = useCallback(
+    ({key, style, index}) => {
+      const row = emotes.slice(index * TOTAL_COLS, (index + 1) * TOTAL_COLS).map(({item}) => [item.code, item]);
```
night, line 63: curious, why do we map this to an array?

src/modules/emote_menu/stores/index.js
dclstn: Yes

src/modules/emote_menu/components/VirtualizedList.jsx
dclstn: What would we be slicing? we're pushing the index

src/modules/emote_menu/components/Emotes.jsx
dclstn: In `VirtualizedList.jsx` we're mapping rows to a list of components with this func, so I believe so yes

src/modules/emote_menu/components/VirtualizedList.jsx
dclstn: This is included for reusability sake, i.e if the `stickyRows` were ever explicitly defined unordered but I get why you would see this as being redundant.

src/modules/emote_menu/components/Menu.jsx
dclstn: Yes.

src/modules/emote_menu/components/Menu.jsx
dclstn: Yes, and handle toggling favorites with `alt` and keeping the menu open with `shift`

src/modules/emote_menu/EmoteMenu.jsx
```diff
@@ -0,0 +1,53 @@
...
+  async embedEmoteMenu() {
+    if ($(BTTV_EMOTE_PICKER_BUTTON_CONTAINER_SELECTOR).length) return;
+    const panel = document.createElement('div');
+    panel.setAttribute('id', 'bttvEmoteMenuContainer');
+    $(CHAT_INPUT_ICONS_SELECTOR).append(panel);
+    ReactDOM.render(<EmoteMenu appendText={this.appendText} />, panel);
+  }
+
+  appendText(string) {
+    const textArea = $(CHAT_TEXT_AREA);
```
night, line 33: WeirdChamp

src/modules/emotes/emojis.js
```diff
@@ -68,6 +74,15 @@ class Emojis extends AbstractEmotes {
           },
         });
 
+        if (!emoji.isAlternative) {
+          // eslint-disable-next-line no-prototype-builtins
+          if (this.emotesByCategory.hasOwnProperty(emoji.category)) {
+            this.emotesByCategory[emoji.category].push(emote);
```
night, line 80: we're performing the same lookup here multiple times:

```
let categoryEmotes = this.emotesByCategory[emoji.category];
if (categoryEmotes == null) {
  categoryEmotes = [];
  this.emotesByCategory[emoji.category] = categoryEmotes;
}
categoryEmotes.push(emote);
```

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+        score: calcScore(1, [Date.now()]),
+      };
+    }
+
+    this.updateIds();
+
+    storage.set('emotes', this.emoteStore);
+  }
+
+  updateIds() {
```
night, line 108: this doesn't need to be a public method.. we can prob move this to a private function in this file

src/modules/emote_menu/components/Preview.jsx
```diff
@@ -0,0 +1,36 @@
...
+          <img alt={emote.name} src={emote.images['3x'] || emote.images['2x'] || emote.images['1x']} />
+        </div>
+        <div>
+          <div className={styles.emoteCode}>{emote.code}</div>
+          <p>
+            from <b>{emote.provider.displayName}</b>
+          </p>
+        </div>
+      </div>
+      {emoteStorage.isFavorite(emote.id) && (
```
night, line 23: avoid `&&` when doing react renders. otherwise you're sending `false` to the renderer. you are meant to send `null` or a react component.

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+        newRecentUses.pop();
+      }
+
+      this.emoteStore.usageHistory[id] = {
+        totalUses: totalUses + 1,
+        recentUses: newRecentUses,
+        score: calcScore(newTotalUses, newRecentUses),
+      };
+    } else {
+      this.emoteStore.usageHistory[id] = {
```
night, line 96: similar feedback to another spot where you are defining a default. invert this operation to avoid duplicate lookups.

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+
+  getFavorites() {
+    return this.emoteStore.favorites;
+  }
+
+  isFavorite(id) {
+    return this.emoteStore.favorites.includes(id);
+  }
+
+  incrementEmote(emote) {
```
night, line 75: `trackUsage`

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+    let {id} = emote;
+    id = String(id);
+
+    // eslint-disable-next-line no-prototype-builtins
+    if (this.emoteStore.usageHistory.hasOwnProperty(id)) {
+      const {totalUses, recentUses} = this.emoteStore.usageHistory[id];
+
+      const newTotalUses = totalUses + 1;
+      const newRecentUses = [Date.now(), ...recentUses];
```
night, line 84: append instead to avoid making a new array

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+
+      if (newRecentUses.length >= MAX_TIMESTAMPS) {
+        newRecentUses.pop();
```
night, line 87: you can use shift in place with above comment

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+    storage.set('emotes', this.emoteStore);
+  }
+
+  updateIds() {
+    this.ids = Object.entries(this.emoteStore.usageHistory)
+      .sort(([, {score: a}], [, {score: b}]) => b - a)
+      .map(([id]) => id);
+  }
+
+  getRecents() {
```
night, line 114: this naming sucks. why are we storing frecent ids in `.ids` .. but this store also works for favorites? maybe worth separating these out to different stores, or at the least making the two kinds of emotes stored more deliberate in naming?

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+        usageHistory: {},
+        favorites: [],
+      };
+    }
+
+    this.updateAllScores();
+    this.updateIds();
+  }
+
+  updateAllScores() {
```
night, line 50: this is used once, but is a public method. we prob can either inline this or make it a private function

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+    for (const [id, {totalUses, recentUses}] of Object.entries(this.emoteStore.usageHistory)) {
+      this.emoteStore.usageHistory[id].score = calcScore(totalUses, recentUses);
+    }
+  }
+
+  toggleFavorite(emote) {
```
night, line 56: maybe `setFavorite` and accept a deliberate value instead? otherwise toggling favorites across windows could result in unexpected behavior?

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+      : [...this.emoteStore.favorites, id];
+
+    storage.set('emotes', this.emoteStore);
+  }
+
+  getFavorites() {
+    return this.emoteStore.favorites;
+  }
+
+  isFavorite(id) {
```
night, line 71: why do we have `is` methods for stuff like this. could we just `getFavorites().includes(id)`?

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,119 @@
...
+    return this.emoteStore.favorites.includes(id);
```
night, line 72: `includes` is slow iirc. do we expect there to be many favorites?

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,121 @@
...
+    }
+    emoteStore.on('updated', callback);
+
+    return () => {
+      emoteStore.off('updated', callback);
+    };
+  }, []);
+
+  useEffect(() => {
+    if (!section.scrollTo) return;
```
night, line 55: nit: use a compare against null. `!= null`

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,121 @@
...
+  if (emotes.length === 0) {
+    return (
+      <div className={styles.empty}>
+        <p>We could not find any results...</p>
+        <Button
+          appearance="primary"
+          target="_blank"
+          rel="noreferrer"
+          href={`https://betterttv.com/emotes/shared/search?query=${search}`}>
+          Search Shared Emotes
```
night, line 102: is this an actionable cta? what would a user be able to do from there? if it's so they can add to their own channel, maybe only surface it when in that context.. or Pro user.. etc.

src/modules/emote_menu/components/Preview.jsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <div {...restProps} key={emote.code}>
+      <div className={styles.content}>
+        <div className={styles.emoji}>
+          <img alt={emote.name} src={emote.images['3x'] || emote.images['2x'] || emote.images['1x']} />
+        </div>
+        <div>
+          <div className={styles.emoteCode}>{emote.code}</div>
+          <p>
+            from <b>{emote.provider.displayName}</b>
```
night, line 19: nit: avoid `b` and other builtins. try to make use of css.

src/modules/emote_menu/components/Preview.jsx
```diff
@@ -0,0 +1,36 @@
...
+function PreviewEmote({emote, ...restProps}) {
+  return (
+    <div {...restProps} key={emote.code}>
+      <div className={styles.content}>
+        <div className={styles.emoji}>
+          <img alt={emote.name} src={emote.images['3x'] || emote.images['2x'] || emote.images['1x']} />
+        </div>
+        <div>
+          <div className={styles.emoteCode}>{emote.code}</div>
+          <p>
```
night, line 18: similar with `p` (as below)

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,71 @@
...
+    wrapperRef.current.addEventListener('scroll', isInViewport, false);
+    isInViewport();
+    return () => {
+      if (wrapperRef.current) wrapperRef.current.removeEventListener('scroll', isInViewport, false);
+    };
+  }, [isInViewport]);
+
+  return (
+    <div {...restProps} style={{height: windowHeight, overflowY: 'scroll'}} ref={wrapperRef}>
+      <div style={{position: 'relative', height: listHeight, width: '100%'}}>
```
night, line 61: use css classes

src/modules/emote_menu/components/Emotes.jsx
dclstn: scrollTo is either `True` or `False`, the ambiguity is probably not ideal. It decides when the emotes list gets scrolled to on section change.

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,121 @@
...
+import styles from '../styles/emotes.module.css';
+import Emote from './Emote.jsx';
+
+const ROW_HEIGHT = 36;
+const WINDOW_HEIGHT = 308;
+const TOTAL_COLS = 7;
+
+function Emotes({onClick, onHover, section, onSection}) {
+  const wrapperRef = useRef(null);
+  const [, setUpdated] = useState(false);
```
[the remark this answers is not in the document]
dclstn: eslint doesn't like that

src/modules/emote_menu/styles/button.module.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.button:focus {
+  background-color: #616163;
+}
+
+.popover {
+  width: 320px;
+  height: 420px;
+  margin-left: 5px;
```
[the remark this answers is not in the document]
dclstn: This is to shift the popover to be directlly above the chat input box, it's not for styling

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,170 @@
...
+      if (emotes.length === 0) continue;
+
+      this.headers.push(this.rows.length);
+      this.rows = this.rows.concat([provider, ...chunkArray(emotes, COLOUMN_COUNT)]);
+    }
+
+    this.emit('updated');
+  }
+
+  loadConditionalEmotes() {
```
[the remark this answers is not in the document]
dclstn: dependable?

src/modules/emote_menu/components/Emotes.jsx
dclstn: Probably wasn't a good idea anyway, if a user is searching for an emote they probably know what they're looking for.. not exploring emotes.

DorCoMaNdO:
I'd like to argue against this change, the simplicity and density of the current BTTV emote menu is why I prefer to use it over the default Twitch and the FFZ emote menus and one of the reasons I keep BTTV installed alongside FFZ.
There are a bunch of people who prefer the BTTV menu as it is over other menus, at the very least I'd hope for an option to revert to the current menu.

The increased padding an added "navigation" features take up a lot of space and display much fewer emotes within the same area. With the chat and emote menu default width of 340px (at least at on a 1080p monitor) the default emote menu displays 6 emotes per row, the FFZ and the preview of the new BTTV menus display 7, as opposed to the current BTTV menu which displays 10.
I can't speak for others' uses for the menu but I find it most useful for third-party emotes (BTTV, FFZ, 7TV via FFZ) as it typically displays them all at the top (technically alphabetically but they all tend to be near the top, and quite literally all of them since the menu defaults to chat height), while I use emote autocomplete for first-party Twitch emotes and less frequently for third-party emotes.

The current menu isn't perfect, but it does it's job well, but if you're looking for improvements to make then I've got 3 I came up with:
1. The option to collapse/expand groups (not just hide).
2. The option to reorder/pin groups to top (to address exceptions to "third-party emotes appear at the top" when you're subscribed to a channel that appears ahead of any of the services alphabetically, and to make more frequently accessed groups more accessible)
3. Favorites (though I question the usefulness of this for users with similar usage to mine, since "favorites" just end up being the emotes I use most frequently via autocomplete).

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  setFavorite(emote, bool) {
+    let {id} = emote;
+    id = String(id);
+
+    if (bool) {
+      this.emoteStore.favorites.push(id);
+    } else {
+      this.emoteStore.favorites = this.emoteStore.favorites.filter((favoriteId) => favoriteId !== id);
```
night, line 96: why not use a Set instead?

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,176 @@
...
+import SafeEventEmitter from '../../../utils/safe-event-emitter.js';
+import watcher from '../../../watcher.js';
+import emojiCategories from './emoji-categories.js';
+import emotes from '../../emotes/index.js';
+import Icons from '../components/Icons.jsx';
+import emoteStorage from './emote-storage.js';
+import {loadTwitchEmotes} from './twitch-emotes.js';
+import cdn from '../../../utils/cdn.js';
+
+export const COLOUM_COUNT = window.location.pathname.endsWith('/chat') ? 7 : 9;
```
night, line 11: `COLOUM`?

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,176 @@
...
+    this.colsCount = cols;
+    this.createRows();
+  }
+
+  createRows() {
+    this.rows = [];
+    this.headers = [];
+
+    for (const {provider, emotes: providerEmotes} of [
+      ...this.dependableProviders,
```
night, line 119: rather than having 3 kinds of providers, have you considered just having a single object for all providers and updating providers by key? why do they need to be separate?

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -0,0 +1,135 @@
...
+}
+`;
+
+const TWITCH_EMOTE_CDN = (id, size) => `https://static-cdn.jtvnw.net/emoticons/v2/${id}/default/dark/${size}`;
+
+function getForcedProviderToChannels(key) {
+  switch (key) {
+    case 0:
+      return {
+        id: 0,
```
night, line 33: is there a way to avoid using a union type here? can id be only a string or an integer?

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -0,0 +1,135 @@
...
+  try {
+    const tempSets = {};
+    const {data} = await twitchApi.graphqlQuery(EMOTE_SET_QUERY);
+
+    for (const {owner, id, emotes} of data.currentUser.emoteSets) {
+      let provider = getForcedProviderToChannels(id);
+
+      if (provider.id === -1 && owner != null) {
+        provider = {
+          id: owner?.id,
```
night, line 92: if owner is not equal to null, why do we null check these?

src/modules/emote_menu/components/Button.jsx
```diff
@@ -0,0 +1,36 @@
...
+  }, []);
+
+  return (
+    <Whisper
+      trigger="click"
+      placement="topEnd"
+      speaker={
+        <Popover
+          className={styles.popover}
+          style={{width: window.location.pathname.endsWith('/chat') ? 320 : 380}}
```
night, line 25: can we avoid hardcoding this in js? is it possible to use css to handle the window size < 380?

src/modules/emote_menu/styles/button.module.css
```diff
@@ -0,0 +1,40 @@
...
+.button:hover {
+  background-color: #6a6a6c;
+}
+
+.button:focus {
+  background-color: #616163;
+}
+
+.popover {
+  width: 380px;
```
night, line 37: we overwrite this in js right now, but should do it all in css.

src/modules/emote_menu/styles/emotes.module.css
```diff
@@ -0,0 +1,78 @@
...
+  margin-top: 16px;
+  align-items: center;
+  width: 100%;
+  flex-direction: column;
+  row-gap: 8px;
+}
+
+.emotesContainer {
+  width: 100%;
+  padding: 0 10px;
```
night, line 53: dimensions should be multiples of 4px, so maybe try 8px

src/modules/emote_menu/styles/emotes.module.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.emotesContainer::-webkit-scrollbar {
+  border-radius: 6px;
+  width: 5px;
+}
+
+.emotesContainer::-webkit-scrollbar-thumb {
+  background: #191c23;
+  border-radius: 5px;
```
night, line 67: 4px

src/modules/emote_menu/styles/emotes.module.css
```diff
@@ -0,0 +1,78 @@
...
+  width: 100%;
+  padding: 0 10px;
+  overflow-y: scroll;
+  overflow-x: hidden;
+  scrollbar-width: thin;
+  scrollbar-color: #191c23 transparent;
+}
+
+.emotesContainer::-webkit-scrollbar {
+  border-radius: 6px;
```
night, line 61: 4px

src/modules/emote_menu/styles/menu.module.css
```diff
@@ -0,0 +1,37 @@
...
+  width: 100%;
+  padding: 0 10px;
+  overflow-y: scroll;
+  overflow-x: hidden;
+  scrollbar-width: thin;
+  scrollbar-color: #191c23 transparent;
+}
+
+.emojis::-webkit-scrollbar {
+  border-radius: 6px;
```
night, line 26: rather than copy/pasting scrollbar styles, can we standardize these? you can use `composes`

night: can you also explain why this wrapping component has a scrollbar, and the emojis container also has a scrollbar? nesting scrollbars sounds like a bug

src/modules/emote_menu/styles/sidebar.module.css
```diff
@@ -0,0 +1,11 @@
+.sidebar {
+  width: 52px;
+  height: 100%;
+  overflow-y: scroll;
+  background-color: #191c23;
+  scrollbar-width: none;
+}
+
+.sidebar::-webkit-scrollbar {
```
night, line 9: can you explain this? overflow-y is why you have a scrollbar, and it sounds like you need to turn that off.

src/modules/emotes/emojis.js
```diff
@@ -68,6 +74,18 @@ class Emojis extends AbstractEmotes {
           },
         });
 
+        if (!emoji.isAlternative) {
+          // eslint-disable-next-line no-prototype-builtins
+          let categoryEmotes = this.emotesByCategory[emoji.category];
+
+          if (categoryEmotes == null) {
+            categoryEmotes = [];
+            this.emotesByCategory[emoji.category] = categoryEmotes;
```
night, line 83: isnt this an off by one?

```
if (categoryEmotes == null) {
  categoryEmotes = [];
  this.emotesByCategory[emoji.category] = categoryEmotes;
}
categoryEmotes.push(emote);
```

webpack.config.js
```diff
@@ -39,21 +39,29 @@ function convertEmojiToolkitCodePointToChar(codePoint) {
...
 function jsonTransform(emojis) {
   const result = {};
+
   for (const emojiData of Object.values(emojis)) {
     const char = convertEmojiToolkitCodePointToChar(emojiData.code_points.fully_qualified);
     const data = {
       char,
       slug: emojiData.shortname.replace(/:/g, ''),
+      category: emojiData.category,
+      isAlternative: false,
```
night, line 49: so we can save a little extra space in memory, can we only specify alternatives if they are true?

src/modules/emote_menu/styles/sidebar.module.css
dclstn: how are you meant to scroll otherwise?

src/modules/emote_menu/stores/index.js
dclstn: For the order they appear in the menu

src/modules/emote_menu/EmoteMenu.jsx
```diff
@@ -0,0 +1,99 @@
...
+    let selectionEnd = element.selectionStart + text.length;
+    const currentValue = element.value;
+    const beforeText = currentValue.substring(0, element.selectionStart);
+    const afterText = currentValue.substring(element.selectionEnd, currentValue.length);
+
+    if (beforeText !== '' && beforeText.substr(-1) !== ' ') {
+      text = ` ${text}`;
+    }
+
+    text = `${beforeText + text} ${afterText}`;
```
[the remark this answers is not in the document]
dclstn: It does not, the above logic checks this I believe

Limekys:
Круто, мне нравится!

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,193 @@
...
+
+    settings.on(`changed.${SettingIds.EMOTES}`, () => {
+      this.loadExtensionEmotes();
+      this.loadDependableEmotes();
+      this.createRows();
+    });
+  }
+
+  async loadFixedEmotes() {
+    if (this.providers.fixed.length > 0) return; // constants already been loaded
```
night, line 68: if a user buys a subscription, their new emotes will never populate outside of a full page refresh.

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -0,0 +1,135 @@
...
+      return {
+        id: -1,
+        displayName: 'Unlocked',
+        icon: Icons.UNLOCK,
+      };
+  }
+}
+
+export async function loadTwitchEmotes() {
+  try {
```
night, line 83: why is this entire operation wrapped in a try catch? wouldn't it just be the http call?

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -0,0 +1,135 @@
...
+        id,
+        displayName
+        profileImageURL(width: 300)
+      }
+    }
+  }
+}
+`;
+
+const TWITCH_EMOTE_CDN = (id, size) => `https://static-cdn.jtvnw.net/emoticons/v2/${id}/default/dark/${size}`;
```
night, line 27: would we want `dark` to be variable based on theme?

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    this.frecentIds = sortHistory(this.emoteStore.usageHistory);
+  }
+
+  trackHistory(emote) {
+    let {id} = emote;
+    id = String(id);
+
+    let emoteHistory = this.emoteStore.usageHistory[id];
```
night, line 65: this usage history looks to be unbounded in size. should there be an upper limit?

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -0,0 +1,135 @@
...
+      };
+    case '19194':
+      return {
+        id: 'twitch-gaming',
+        displayName: 'Twitch Gaming',
+        icon: Icons.TWITCH_GAMING,
+      };
+    default:
+      return {
+        id: -1,
```
night, line 75: id should not be mixed type. use a sentinel value string instead

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -0,0 +1,135 @@
...
+  try {
+    const tempSets = {};
+    const {data} = await twitchApi.graphqlQuery(EMOTE_SET_QUERY);
+
+    for (const {owner, id, emotes} of data.currentUser.emoteSets) {
+      let provider = getForcedProviderToChannels(id);
+
+      if (provider.id === -1 && owner != null) {
+        provider = {
+          id: owner.id,
```
night, line 92: using a bare id for provider ids could result in conflicts between other potential category ids. we should make a `EmoteMenuCategories` constant for all of the categories. One of them can be `EmoteMenuCategories.TWITCH_CHANNEL`, and we can make this be a template string instead: `EmoteMenuCategories.TWITCH_CHANNEL-${owner.id}`

src/modules/emote_menu/styles/menu.module.css
```diff
@@ -0,0 +1,19 @@
...
+
+.header {
+  display: flex;
+  padding: 8px;
+  column-gap: 4px;
+}
+
+.content {
+  display: flex;
+  height: 308px;
```
night, line 13: this is a very specific number. how is this obtained? is there a way to get a rounder number?

src/modules/settings/styles/_rsuite.less
```diff
@@ -1,104 +1,112 @@
-@theme: 'dark'; //set to "default" for light theme
```
night, line 1: can you explain the changes in this file?

webpack.config.js
```diff
@@ -39,21 +39,28 @@ function convertEmojiToolkitCodePointToChar(codePoint) {
...
     result[data.slug] = data;
+
     for (const alternativeShortName of emojiData.shortname_alternates) {
       // :tf: is a legacy betterttv global emote
       if (alternativeShortName === ':tf:') {
         continue;
       }
+
+      data.isAlternative = true;
       result[alternativeShortName.replace(/:/g, '')] = data;
```
night, line 60: im not sure this works as you intend, and potentially only emotes without alternatives are rendering.

data on line 60 is equal to data on line 51, which means data is also equal in the resulting `result`

src/modules/emote_menu/stores/emote-storage.js
dclstn: ` for (const [id, emote] of Object.entries(this.emoteStore.usageHistory)) {
      const score = calcScore(emote);

      if (score <= PURGE_SCORE) {
        delete this.emoteStore.usageHistory[id];
      } else {
        this.emoteStore.usageHistory[id].score = score;
      }
    }`
    
    An elegant way is that we can purge all scores below a threshold when updated the score on load, meaning everything "irrelevant" gets removed

src/modules/settings/styles/_rsuite.less
dclstn: for what ever reason `prettier --write .` updated it

src/modules/emote_menu/components/Emote.jsx
```diff
@@ -0,0 +1,23 @@
...
+    <button
+      key={emote.code}
+      onClick={() => onClick(emote)}
+      onMouseOver={() => onMouseOver(emote)}
+      onFocus={() => onMouseOver(emote)}
+      type="button"
+      className={styles.emote}>
+      <img
+        className={styles.emoteImage}
+        srcSet={`${emote.images['1x']} 1x, ${emote.images['2x']} 2x, ${emote.images['3x']} 4x`}
```
night, line 15: image
image

src/modules/emote_menu/stores/index.js
```diff
@@ -0,0 +1,197 @@
...
+
+  loadExtensionEmotes() {
+    this.categories.extension = [
+      {
+        provider: {
+          id: emotesCategoryIds.BETTERTTV,
+          displayName: 'BetterTTV',
+          icon: Icons.IMAGE(cdn.url('/assets/logos/mascot.png'), 'BetterTTV'),
+        },
+        emotes: emotes.getEmotesByProviders(['bttv-channel', 'bttv-personal', 'bttv']),
```
night, line 93: should we separate the betterttv personal emotes?

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -0,0 +1,148 @@
...
+      });
+    });
+
+    // twitch seperates emotes by tier, so we merge them into one set
+    // eslint-disable-next-line no-prototype-builtins
+    if (tempSets.hasOwnProperty(provider.id)) {
+      tempSets[provider.id].emotes = [...tempSets[provider.id]?.emotes, ...providerEmotes];
+      continue;
+    }
+
```
night, line 143: should we sort current channel closer to the top of the list?

src/modules/emote_menu/components/VirtualizedList.jsx
```diff
@@ -0,0 +1,111 @@
...
+    setData({
+      header: isStuck
+        ? {
+            current,
+            top: 0,
+            position: 'sticky',
+          }
+        : {
+            current,
+            top: (next - 1) * rowHeight,
```
night, line 62: the header logic is pretty buggy, so you will need to iron this out. it looks like it doesn't update until scroll stop. the animations also only work one direction (scrolling down).

src/modules/emote_menu/components/Emote.jsx
night: can we also show a loading state of some kind?

src/modules/emote_menu/components/Button.jsx
```diff
@@ -0,0 +1,35 @@
...
+    });
+  }, []);
+
+  return (
+    <Whisper
+      trigger="active"
+      placement="topEnd"
+      speaker={
+        <Popover className={styles.popover} full>
+          <EmoteMenu triggerRef={triggerRef} appendToChat={appendToChat} />
```
night, line 28: image

the popout arrow does not align to the button

src/modules/emote_menu/components/Emote.jsx
night: search is also for some reason unpredictable with rendering urls 
image

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    return () => {
+      window.removeEventListener('keydown', buttonPressCallback, false);
+      window.removeEventListener('keyup', buttonPressCallback, false);
+    };
+  }, []);
+
+  useEffect(() => {
+    function handleEnterCallback(event) {
+      if (event.keyCode === 13) {
```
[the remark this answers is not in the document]
night: we also really need arrow key support too for navigation.

src/modules/emote_menu/EmoteMenu.jsx
```diff
@@ -0,0 +1,136 @@
...
+  hide() {
+    const bttvEmotePicker = document.querySelector(BTTV_EMOTE_PICKER_BUTTON_SELECTOR);
+    const emotePicker = document.querySelector(EMOTE_PICKER_BUTTON_SELECTOR);
+
+    emotePicker.style.display = 'inline-flex';
+    bttvEmotePicker.style.display = 'none';
+  }
+
+  show() {
+    const bttvEmotePicker = document.querySelector(BTTV_EMOTE_PICKER_BUTTON_SELECTOR);
```
night, line 99: we need to also render the old button, and re-educate users of it being moved. otherwise we will have a ton of confused users.

src/modules/emote_menu/EmoteMenu.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+  render() {
+    return <EmoteMenu {...this.props} />;
+  }
+}
+
+export default class EmoteMenuModule {
+  constructor() {
+    domObserver.on(CHAT_INPUT_ICONS_SELECTOR, () => this.load());
+    settings.on(`changed.${SettingIds.CLICK_TWITCH_EMOTES}`, () => this.load());
```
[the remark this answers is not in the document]
dclstn: This class only gets created once however?

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,143 @@
...
+          down: event.keyCode === keycodes.DownArrow,
+          right: event.keyCode === keycodes.RightArrow,
+          left: event.keyCode === keycodes.LeftArrow,
+        });
+
+        event.preventDefault(); // prevent scrolling
+      }
+    }
+
+    window.addEventListener('keydown', buttonPressCallback, false);
```
night, line 75: it would be great to only create one listener for all the key presses

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,143 @@
...
+
+    function arrowPressCallback(event) {
+      if (
+        event.keyCode === keycodes.UpArrow ||
+        event.keyCode === keycodes.DownArrow ||
+        event.keyCode === keycodes.RightArrow ||
+        event.keyCode === keycodes.LeftArrow
+      ) {
+        setArrowKeys({
+          top: event.keyCode === keycodes.UpArrow,
```
night, line 65: these do not appear to ever get set back to false since we only listen to keydown. listening to key down is fine for all of this, but we likely should be handling button presses directly for keyboard navigation. for example, a user holding down an arrow key will generate multiple arrow press events but we are already set to `true`. if we're setting state we miss that nuance.

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    onHide();
+  }, []);
+
+  useEffect(() => setSearch(''), [section]);
+
+  useEffect(() => {
+    function buttonPressCallback(event) {
+      altKeyPressed = event.altKey;
```
night, line 53: should we be storing these states with a `useState` hook? multiple mounts of a component could cause these variables to be unexpectedly changed during runtime.

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,143 @@
...
+
+  useEffect(() => {
+    function handleEnterCallback(event) {
+      if (event.keyCode === keycodes.Enter) {
+        event.preventDefault();
+        handleClick(selected);
+      }
+    }
+
+    window.addEventListener('keydown', handleEnterCallback, false);
```
night, line 111: similar comment regarding multiple keydowns. we really only should have 1 in this component.

src/modules/emote_menu/components/Sidebar.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+export default function Sidebar({section, onChange, providers}) {
+  const containerRef = useRef(null);
+
+  useEffect(() => {
+    if (section.eventKey == null) return;
+
+    clearTimeout(timer);
+
+    timer = setTimeout(() => {
```
night, line 18: similar comment regarding global state in components. stateful data should not be stored outside of the component context unless it's a store.

src/modules/emote_menu/components/Sidebar.jsx
```diff
@@ -0,0 +1,52 @@
...
+      }
+
+      if (depth + ITEM_HEIGHT > top + SIDEBAR_HEIGHT) {
+        containerRef.current.scrollTo({
+          top: depth - SIDEBAR_HEIGHT + 36,
+          left: 0,
+          behavior: 'smooth',
+        });
+      }
+    }, 100);
```
night, line 40: i think you're intending to use this as a debounce, but this forces a required delay on changes. I think we'd want to use a throttle library for this, which allows us to call this logic immediately, and then again after function debounce.

src/modules/emote_menu/components/Sidebar.jsx
```diff
@@ -0,0 +1,52 @@
...
+          top: depth,
+          left: 0,
+          behavior: 'smooth',
+        });
+
+        return;
+      }
+
+      if (depth + ITEM_HEIGHT > top + SIDEBAR_HEIGHT) {
+        containerRef.current.scrollTo({
```
night, line 34: we wrote this logic twice, but it seems like we instead want to compute `top` as a variable, and pass that for both.

```
let newTop;
if (depth < top) {
  newTop = depth;
}
if (depth + ITEM_HEIGHT > top + SIDEBAR_HEIGHT) {
  newTop = depth - SIDEBAR_HEIGHT + 36; // where does 36 come from? why isnt that a constant?
}
if (newTop == null) {
  return;
}

containerRef.current.scrollTo(...);
```

src/modules/emote_menu/components/Sidebar.jsx
```diff
@@ -0,0 +1,52 @@
...
+        });
+      }
+    }, 100);
+  }, [section]);
+
+  return (
+    <div className={styles.sidebar} ref={containerRef}>
+      <Nav vertical appearance="subtle" onSelect={onChange} activeKey={section.eventKey}>
+        {providers.map((provider) => (
+          <Nav.Item key={provider.id} eventKey={provider.id} icon={provider.icon} />
```
night, line 47: there's a hover state for icons, but not for avatars.

we also prob should add tooltips

src/modules/emote_menu/EmoteMenu.jsx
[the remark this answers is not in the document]
night: right, but that doesn't stop someone from importing it later. we should export a singleton if that's the use case

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,121 @@
...
+
+const MAX_TIMESTAMPS = 100;
+const MAX_FRECENTS = 250;
+
+function sortHistory(history, limit) {
+  const sortedHistory = new Set();
+
+  Object.entries(history)
+    .sort(([, {score: a}], [, {score: b}]) => b - a)
+    .slice(0, limit)
```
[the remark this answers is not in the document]
dclstn: Whenever the `frecents` or `favorites` changed this will update localstorage, or should it persist whenever this func is ran?

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,303 @@
...
+    newYColumnCount = rowColumnCounts[y];
+  }
+
+  return {
+    x: Math.min(newYColumnCount - 1, x),
+    y: newY,
+  };
+}
+
+function travelDown(rowColumnCounts, {x, y}, numBlocks = 1) {
```
night, line 36: nit: pull this all out to a custom react hook. we can call it something like `useGridKeyboardNavigation`

src/modules/emote_menu/components/Header.jsx
```diff
@@ -0,0 +1,48 @@
...
+import Icon from 'rsuite/lib/Icon/index.js';
+import InputGroup from 'rsuite/lib/InputGroup/index.js';
+import Input from 'rsuite/lib/Input/index.js';
+import IconButton from 'rsuite/lib/IconButton/index.js';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import {faSearch} from '@fortawesome/free-solid-svg-icons/faSearch';
+import {faTimes} from '@fortawesome/free-solid-svg-icons/faTimes';
+
+function Header({value, onChange, onHide, selected, ...restProps}) {
+  const ref = useRef(null);
```
night, line 11: nit: `searchInputRef`

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    function handleArrowKeys(event) {
+      if (
+        event.keyCode === keycodes.UpArrow ||
+        event.keyCode === keycodes.DownArrow ||
+        event.keyCode === keycodes.RightArrow ||
+        event.keyCode === keycodes.LeftArrow
+      ) {
+        setArrowKeys({
```
night, line 77: I think I commented previously we should stop storing state for these. instead, either add a listener where we need them or pass a callback function in that allows the nested component to hook into these events on mount

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,303 @@
...
+        break;
+      case arrowKeys.down:
+        newCords = travelDown(rowColumnCounts, cords);
+        break;
+      case arrowKeys.right:
+        newCords = travelRight(rowColumnCounts, cords);
+        break;
+      case arrowKeys.left:
+        newCords = travelLeft(rowColumnCounts, cords);
+        break;
```
night, line 273: we don't seem to support other keys, like home or end. we can we support more of the keyboard nav buttons?

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,121 @@
...
+
+    this.emoteStore.usageHistory[id] = emoteHistory;
+
+    this.updateFrecents();
+    this.save();
+  }
+
+  updateFrecents() {
+    this.emoteStore.usageHistory = sortHistory(this.emoteStore.usageHistory, MAX_FRECENTS);
+    this.frecents = new Set(Object.keys(this.emoteStore.usageHistory));
```
night, line 93: this attribute is being lazily defined on the class. we should set default values in the constructor to make this class more readable.

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -0,0 +1,121 @@
...
+
+    this.save();
+  }
+
+  setFavorite(emote, bool) {
+    let {id} = emote;
+    id = String(id);
```
night, line 100: we can't use `id` alone because `id` can be re-used between emote providers. you need to audit all usage of `id` in your code.

src/modules/settings/styles/_rsuite.less
```diff
@@ -111,3 +115,9 @@
     width: 100%;
   }
 }
+
+.@{ns}nav-item-content {
+  svg {
+    width: 100% !important;
```
night, line 121: does this stretch the images?

src/utils/twitch.js
```diff
@@ -133,6 +137,20 @@ export default {
     });
   },
 
+  async getCurrentProfile() {
+    if (currentProfile != null) {
+      return currentProfile;
+    }
+
+    try {
+      const res = await twitchAPI.get(`users?login=${currentUser.displayName}`);
```
night, line 146: 1. using a display name as the login is not reliable. they are distinctly different fields, and some users exist where `login != displayName.toLowerCase()`
2. using a login is brittle, since users can change their name. always use a twitch user id for resolution.

additionally, should we be using gql for this since it's onsite? twitch api v5 is deprecated

src/modules/emote_menu/components/Sidebar.jsx
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    containerRef.current.scrollTo({
+      top: newTop,
+      left: 0,
+      behavior: 'smooth',
+    });
+  });
+
+  useEffect(() => throttleFunc(), [section]);
```
night, line 39: are you sure this is doing what you intend?

In the react lifecycle, every time a component "rerenders" it is calling this function again.

react does some magic around storing references to stop hooks from being re-ran, but when you compute functions like `throttleFunc` in the render function it will be a completely new throttle reference every time the render function runs.

as-is, anytime `section` changes the throttle function being called is a new reference.

src/modules/settings/styles/_rsuite.less
dclstn: It doesn't, the svgs just won't will the bounding box otherwise. this fixes it

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -0,0 +1,363 @@
...
+import Emote from './Emote.jsx';
+import Icons from './Icons.jsx';
+import {NavigationModeTypes, RowHeight, WindowHeight} from '../../../constants.js';
+import keycodes from '../../../utils/keycodes.js';
+
+const MAX_COLUMN_COUNT = 6;
+
+let navigationMode = NavigationModeTypes.MOUSE;
+
+function travelUp(rowColumnCounts, {x, y}, numBlocks = 1) {
```
night, line 16: migrate grid stuff to a util file please

src/utils/emotes.js
```diff
@@ -0,0 +1,22 @@
...
+    case providerId.startsWith('emoji'):
+    case providerId.startsWith('bttv'):
+      return `${emotesCategoryIds.BETTERTTV}-${emoteId}`;
+    case providerId.startsWith('ffz'):
+      return `${emotesCategoryIds.FRANKERFACEZ}-${emoteId}`;
+    default:
+  }
+
+  switch (providerId) {
+    case emotesCategoryIds.FAVORITES:
```
night, line 15: since favorites and frecents are not true emote providers, we may want to throw an error instead and prevent ids from being generated for them

src/utils/emotes.js
```diff
@@ -0,0 +1,22 @@
+import {emotesCategoryIds} from '../constants.js';
+
+/* eslint-disable import/prefer-default-export */
+export function getEmoteIdFromProvider(emoteId, providerId) {
```
night, line 4: maybe we should call these `emoteCanonicalId`

src/modules/emote_menu/stores/emote-storage.js
```diff
@@ -64,9 +64,15 @@ class EmoteStorage extends SafeEventEmitter {
...
 
   trackHistory(emote) {
-    const id = getEmoteIdFromProvider(emote.id, emote.provider.id);
+    let emoteCanonicalId = null;
 
-    let emoteHistory = this.emoteStore.usageHistory[id];
+    try {
+      emoteCanonicalId = getEmoteIdFromProvider(emote.id, emote.provider.id);
+    } catch (e) {
+      emoteCanonicalId = emote.id;
```
night, line 72: curious: why are the emotes in `favorites` and `frecents` classified as those providers? shouldn't `emote.provider` be their original provider? see the code at `src/modules/emote_menu/stores/index.js` line 129, 133, where emote is an existing reference.

src/modules/emote_menu/stores/index.js
```diff
@@ -188,8 +188,15 @@ class EmoteStore extends SafeEventEmitter {
...
 
   toggleFavorite(emote, forceUpdate = false) {
-    const id = getEmoteIdFromProvider(emote.id, emote.provider.id);
-    emoteStorage.setFavorite(emote, !emoteStorage.favorites.has(id));
+    let emoteCanonicalId = null;
+
+    try {
+      emoteCanonicalId = getEmoteIdFromProvider(emote.id, emote.provider.id);
+    } catch (e) {
+      emoteCanonicalId = emote.id;
```
night, line 196: similar comment here

src/modules/emote_menu/components/Sidebar.jsx
```diff
@@ -0,0 +1,57 @@
...
+import Tooltip from 'rsuite/lib/Tooltip/index.js';
+import {RowHeight, WindowHeight} from '../../../constants.js';
+import styles from '../styles/sidebar.module.css';
+
+const ITEM_HEIGHT = 42;
+
+export default function Sidebar({section, onChange, providers}) {
+  const containerRef = useRef(null);
+
+  function throttleFunc() {
```
night, line 13: nit: inline this in the useEffect hook

src/modules/emote_menu/components/GridKeyboardNavigation.jsx
```diff
@@ -0,0 +1,154 @@
+import {useCallback, useEffect, useState} from 'react';
+import {NavigationModeTypes} from '../../../constants.js';
+import keycodes from '../../../utils/keycodes.js';
+
+const MAX_COLUMN_COUNT = 6;
```
night, line 5: column counts are dynamic, and can be 6 or 9

src/modules/emote_menu/components/GridKeyboardNavigation.jsx
```diff
@@ -0,0 +1,154 @@
...
+    if (newY <= y) {
+      return {x, y};
+    }
+    return {x: newX, y: newY};
+  }
+
+  return {x: x + 1, y};
+}
+
+function travelBottom(rowColumnCounts, {x}) {
```
night, line 78: this should be traveling to the end:

```
function travelEnd(rowColumnCounts, {x, y}) {
  const newY = rowColumnCounts.length - 1;
  const newYColumnCount = rowColumnCounts[newY];

  // if the end has no columns, travel up from the end instead
  if (newYColumnCount === 0) {
    const {x: proposedX, y: proposedY} = travelUp(rowColumnCounts, {x: MAX_COLUMN_COUNT, y: newY});
    // if y didn't actually changed, original coords get returned
    if (proposedY >= newY) {
      return {x, y};
    }

    return {
      x: proposedX,
      y: proposedY,
    };
  }

  return {
    x: newYColumnCount - 1,
    y: newY,
  };
}
```

src/modules/emote_menu/components/GridKeyboardNavigation.jsx
```diff
@@ -0,0 +1,154 @@
...
+  const newY = rowColumnCounts.length - 1;
+  const newYColumnCount = rowColumnCounts[newY];
+
+  return {
+    x: Math.min(newYColumnCount - 1, x),
+    y: newY,
+  };
+}
+
+function travelTop(rowColumnCounts, {x}) {
```
night, line 88: this should be traveling to the home:

```
function travelHome(rowColumnCounts, {x, y}) {
  const newY = 0;
  const newYColumnCount = rowColumnCounts[0];

  // if the home has no columns, travel down from the home instead
  if (newYColumnCount === 0) {
    const {x: proposedX, y: proposedY} = travelDown(rowColumnCounts, {x: 0, y: newY});
    // if y didn't actually change, original coords get returned
    if (proposedY <= newY) {
      return {x, y};
    }

    return {
      x: proposedX,
      y: proposedY,
    };
  }

  return {
    x: 0,
    y: newY,
  };
}
```

src/modules/emote_menu/components/GridKeyboardNavigation.jsx
```diff
@@ -0,0 +1,154 @@
...
+  const [cords, setCords] = useState({x: 0, y: 0});
+
+  const handleKeyPress = useCallback(
+    (event) => {
+      setNavigationMode(NavigationModeTypes.ARROW_KEYS);
+
+      let newCords = null;
+
+      switch (event.keyCode) {
+        case keycodes['8numpad']:
```
night, line 114: remove these number keys

src/modules/emote_menu/components/GridKeyboardNavigation.jsx
```diff
@@ -0,0 +1,154 @@
...
+          break;
+        case keycodes.End:
+          newCords = travelBottom(rowColumnCounts, cords);
+          event.preventDefault();
+          break;
+        case keycodes.Home:
+          newCords = travelTop(rowColumnCounts, cords);
+          event.preventDefault();
+          break;
+        default:
```
night, line 142: can we support tab and shift + tab?

night: pressing enter also does not appear to place the emote in chat

night (commented):
popout alignment appears to be off:

image

image

image


there is no selected state for non-native icons in the category selector:

image

image

src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -43,6 +43,7 @@ const Emotes = React.forwardRef(
           <div key={key} style={style} className={classNames(className, styles.row)}>
             {row.map((emote, x) => (
               <Emote
+                key={emote.id}
```
night, line 46: this may need to be canonical ids, as two ids can be shared across providers

PR #4722, opened 2021-08-06
[Bug Fix] Fix PIP button being large when page load

Matsukii:
Picture-in-Picture button is larger while page loads (image) and gets smaller after finish loading.
Added max width and height property to #bttv-picture-in-picture

Screenshot_265
*Disclaimer: I only forked and changed the CSS as done with devTools, didn't actually downloaded and tested*

### DIFF
--- a/src/modules/video_player/style.css
+++ b/src/modules/video_player/style.css
@@ -16,6 +16,8 @@
 
 #bttv-picture-in-picture {
   display: inline-flex !important;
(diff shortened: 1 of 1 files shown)


night:
adding additional styles will not fix flashing due to styles being lazily loaded in the current builds. this is fixed via 12b52400cd61a6f0a346d44cee9b270f43912b0c instead.

PR #4752, opened 2021-08-18
ties '/clear' behavior to message delete setting

lhocke:
it's become necessary for a lot of streamers to add "/clear" as a panic response to malicious raids and bttv's override behavior keeps streamers from protecting their audiences from it. this PR ties the override to the "Deleted Messages" setting instead so users have to opt in to seeing things a streamer is trying to clear.


dclstn:
I like this idea however it doesn't seem wise to bind it to the `showDeletedMessages` flag, if this were to exist it would most likely be its own setting that's default on as currently ALL users have this enabled.

lhocke:
yeah, ultimately I would prefer it in it's own setting this was more of a quick fix coded on my lunch break. I can put together something a little cleaner over the weekend. Any suggestions and/or preference for where that setting would be (inside another category, it's own category at the top or bottom, etc.)?

lhocke:
making a couple changes, i'll close this PR and open a new one with with an explicit setting

PR #4753, opened 2021-08-20
Creates setting for clear command behavior

lhocke:
I've updated my previous PR to instead have an explicit toggle for how bttv handles the /clear command. Same reasoning applies, sometimes chat needs to be cleared to remove hateful and/or threatening messages so viewers don't have to read them. This PR creates a toggle with the default behavior set to allow chat clear.


src/modules/chat_deleted_messages/index.js
```diff
@@ -33,10 +33,13 @@ class ChatDeletedMessagesModule {
   }
 
   handleMessage({message, preventDefault}) {
+    const clearChatBehavior = settings.get(SettingIds.CLEAR_BEHAVIOR);
```
dclstn, line 36: `handleMessage` gets called on every new message, may be more efficient to do this setting lookup inside the case statement to avoid unnecessary computation.

src/modules/settings/components/settings/PreventChatClear.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Panel from 'rsuite/lib/Panel/index.js';
+import Toggle from 'rsuite/lib/Toggle/index.js';
+import {registerComponent, useStorageState} from '../Store.jsx';
+import {CategoryTypes, SettingIds} from '../../../../constants.js';
+import styles from '../../styles/header.module.css'
```
dclstn, line 6: Have you run `npm run lint` on this file? looks like there is some formatting issues.

lhocke: the lint script itself is throwing errors for me, but you're right. looked at it again and ran prettier on it's own to make sure it was fixed

src/modules/chat_deleted_messages/index.js
lhocke: eslint isn't happy with declarations in case statements, I can move the lookup into the if statement though since the variable isn't actually used again

night:
thanks for the pull request, but we intentionally prevent chat clears to prevent group censorship. moderators and bots are still able to remove individual messages.

PR #4765, opened 2021-08-28
Add multi-pasting functionality to tables

dclstn:
`Happy\nLittle\nThings` will now convert to 3 new rows in a table

### DIFF
--- a/src/modules/settings/components/Table.jsx
+++ b/src/modules/settings/components/Table.jsx
@@ -69,23 +69,48 @@ function CustomWhisper(props) {
   );
 }
 
-function useOutsideAlerter(ref, onClick, rowData) {
+function EditCell({rowData, dataKey, onChange, onMouseOver, onMouseLeave, onClick, onPaste, ...props}) {
+  const wrapperRef = useRef(null);
+
   useEffect(() => {
     function handleClickOutside(event) {
-      if (ref.current && !ref.current.contains(event.target)) {
+      if (wrapperRef.current && !wrapperRef.current.contains(event.target)) {
         onClick(rowData.id);
         document.removeEventListener('mousedown', handleClickOutside);
       }
     }
+
+    function handlePasteCallback(event) {
+      onPaste(event, rowData.id, dataKey);
+    }
+
+    wrapperRef.current.addEventListener('paste', handlePasteCallback);
     document.addEventListener('mousedown', handleClickOutside);
-  }, [ref]);
+
+    return () => {
+      wrapperRef.current.removeEventListener('paste', handlePasteCallback);
+    };
+  }, []);
+
+  return (
+    <Cell {...props}>
+      <input
+        ref={wrapperRef}
+        className={classNames({'bttv-rs-input': true}, styles.tableContentEditing)}
+        defaultValue={rowData[dataKey]}
+        onChange={(event) => onChange(rowData.id, dataKey, event.target.value)}
+      />
+    </Cell>
+  );
 }
 
-function EditCell({rowData, dataKey, onChange, onMouseOver, onMouseLeave, onClick, onPaste, ...props}) {
+function CustomCell(props) {
+  const {rowData, dataKey, onChange, onMouseOver, onMouseLeave, onClick, onPaste, ...restProps} = props;
+
   switch (rowData.status) {
     case Status.HOVERING:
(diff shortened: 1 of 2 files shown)


src/modules/settings/components/Table.jsx
```diff
@@ -171,22 +182,47 @@ function EditTable({options, setValue, value, ...props}) {
     return parseInt(Object.keys(data)[Object.keys(data).length - 1], 10) + 1 || 0;
   }
 
-  function addRow() {
+  function createRow() {
     const id = nextId();
-    const newRow = {id, status: Status.EDIT};
-    for (const header of options) {
-      switch (header.type) {
+    const row = {};
```
night, line 187: why did we remove the initial attributes from this instantiation?

src/modules/settings/components/Table.jsx
```diff
@@ -171,22 +182,47 @@ function EditTable({options, setValue, value, ...props}) {
...
+
+  function addEmptyRow() {
+    const newRow = createRow();
+    data[newRow.id] = newRow;
+    setData({...data});
+  }
+
+  function handlePaste(event, id, dataKey) {
+    const paste = (event.clipboardData || window.clipboardData).getData('text');
+    const words = paste.split('\n');
```
night, line 216: these are lines, not words.

we also want to likely trim and filter the results:

```
const lines = paste.split('\n').map(line => line.trim()).filter(line => line.length > 0);
```

src/modules/settings/components/Table.jsx
```diff
@@ -171,22 +182,47 @@ function EditTable({options, setValue, value, ...props}) {
...
         default:
           break;
       }
     }
-    data[id] = newRow;
+
+    return row;
+  }
+
+  function addEmptyRow() {
```
night, line 208: nit: `handleAddRowClick`

src/modules/settings/components/Table.jsx
```diff
@@ -97,6 +97,17 @@ function EditCell({rowData, dataKey, onChange, onMouseOver, onMouseLeave, onClic
     case Status.EDIT: {
       const wrapperRef = useRef(null);
       useOutsideAlerter(wrapperRef, onClick, rowData);
+
+      useEffect(() => {
```
night, line 101: im surprised we don't have linting errors, but you cannot reliably use hooks unless they are the first thing being executed in a function. https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

night: this applies also to the ones above.

src/modules/settings/components/Table.jsx
```diff
@@ -69,23 +69,48 @@ function CustomWhisper(props) {
...
-function useOutsideAlerter(ref, onClick, rowData) {
+function EditCell({rowData, dataKey, onChange, onMouseOver, onMouseLeave, onClick, onPaste, ...props}) {
+  const wrapperRef = useRef(null);
+
   useEffect(() => {
     function handleClickOutside(event) {
-      if (ref.current && !ref.current.contains(event.target)) {
+      if (wrapperRef.current && !wrapperRef.current.contains(event.target)) {
         onClick(rowData.id);
         document.removeEventListener('mousedown', handleClickOutside);
```
night, line 79: do we need to clean this up in the unmount?

dclstn: when the `onClick` callback is called it changes status anyway and will unmount

src/modules/settings/components/Table.jsx
```diff
@@ -89,6 +88,7 @@ function EditCell({rowData, dataKey, onChange, onMouseOver, onMouseLeave, onClic
 
     return () => {
       wrapperRef.current.removeEventListener('paste', handlePasteCallback);
```
night, line 90: `wrapperRef` can change, so you may need to break this out to its own effect:

```
useEffect(() => {
  function handlePasteCallback(event) {
    onPaste(event, rowData.id, dataKey);
  }
  const currentWrapperRef = wrapperRef.current;
  currentWrapperRef.addEventListener('paste', handlePasteCallback);
  return () => currentWrapperRef.removeEventListener('paste', handlePasteCallback);
}, [wrapperRef.current])
```

PR #4822, opened 2021-09-15
Fix for preview not updating in search mode

dclstn:


src/modules/emote_menu/components/Emotes.jsx
```diff
@@ -103,15 +103,17 @@ const SearchedEmotes = React.forwardRef(
       [navigationMode]
     );
 
+    useEffect(() => setSelected(emotes[cords.y][cords.x].item), [cords]);
```
night, line 106: nit: can we go and mass update `cords` to be `coords` (coordinates)

PR #4840, opened 2021-09-17
Add 7TV global and channel emotes

DmitryScaletta:


night:
Thanks for the pull request. Unfortunately since 7TV emotes do not abide by the Twitch Terms of Service and our emote guidelines, we not willing to allow this content on our extension.

Fittiboy:
How are they breaking ToS exactly, and is that something that can change? Or is that a fundamental 7tv issue?

robmonte:
> Thanks for the pull request. Unfortunately since 7TV emotes do not abide by the Twitch Terms of Service and our emote guidelines, we not willing to allow this content on our extension.

7TV is not as loose with emotes as it seems. They do actually remove emotes they find too edgy or questionable, even if it follows both rules they have stated on the website (the rules being: is safe to show on stream and doesn't violate copyright). 

Saying they don't abide by Twitch ToS is fairly misleading though. If you want to get pedantic, BTTV doesn't follow Twitch ToS either because there's copyrighted emotes in BTTV which Twitch ordinarily wouldn't allow.

barnumbirr:
7tv support in BTTV would indeed be awesome!
None of the other extensions even come close to BTTV in terms of amount of features or polish.

PieterVMB:
7TV is following ToS as far as I can see, 7tv is getting more popular and there will be more demand to have it added in BTTV, if you so afraid at all you should add a disclaimer that the emotes that will show in the 7tv section falls under the streamer their choice and that the streamer will be held responsible for breaking any ToS etc.

sheikhijat:
7TV support is extremely needed. The reasoning of TOS stated seems really sus.

PR #4843, opened 2021-09-17
[Fix] Blurry Emotes on Zoom

dclstn:
Emotes will rerender on zoom change which will prevent blurry appearances.


dclstn:
redundant pr

PR #4845, opened 2021-09-17
[Feature] Prepopulate user's frecent emotes

dclstn:


src/constants.js
```diff
@@ -174,6 +174,17 @@ export const emotesCategoryIds = {
   FRECENTS: 'frecents',
 };
 
+export const defaultFrequentEmotes = {
+  'twitch-global-25': {recentUses: [Date.now()], totalUses: 1, score: 100}, // Kappa
+  'twitch-global-305954156': {recentUses: [Date.now()], totalUses: 1, score: 100}, // PogChamp
+  'bttv-566c9f6365dbbdab32ec0532': {recentUses: [Date.now()], totalUses: 1, score: 100}, // ConcernDoge
```
dclstn, line 180: note: if we decide to merge #4846 then some of these prefixes will need to be updated

dclstn: e.g. `bttv-566c9f6365dbbdab32ec0532` will become `bttv-global-566c9f6365dbbdab32ec0532`

src/constants.js
```diff
@@ -174,6 +174,17 @@ export const emotesCategoryIds = {
   FRECENTS: 'frecents',
 };
 
+export const defaultFrequentEmotes = {
```
night, line 177: constants should never be camelCase.

for enum values, use `PascalCase`.
for everything else, use `SNAKE_CASE`.

src/constants.js
night: because emotes are unique per provider and can be duplicated between contexts, I'm not sure why we would want to put "global" in the key?

PR #4846, opened 2021-09-17
[Feature] Seperate Global and Channel BetterTTV emotes

dclstn:


src/modules/emote_menu/stores/index.js
```diff
@@ -73,14 +73,23 @@ class EmoteStore extends SafeEventEmitter {
...
+      twitch.getCurrentChannelProfilePicture(),
+    ]);
 
     emoteProviderCategories = [
       createCategory(
-        emotesCategoryIds.BETTERTTV,
-        'BetterTTV',
+        emotesCategoryIds.BETTERTTV_CHANNEL,
+        'BetterTTV Channel',
+        Icons.IMAGE(channelPicture == null ? cdn.url('/assets/logos/mascot.png') : channelPicture, 'BetterTTV'),
```
night, line 85: it is confusing to use the same image for the channel all over the place. we need to have distinguishable icons

dclstn: any ideas on what the image could be instead?

night: maybe there's some way we can build a component which renders the channel icon with a provider icon overlaid in the bottom right corner.

night:
superseded by cbdb9e03a24df6d04078503b0c133ce8c95c8d1d

PR #4847, opened 2021-09-17
[Feature] Smart tool-tips

dclstn:
PR adds tooltips to make the new emote menu easier to navigate

image


src/constants.js
```diff
@@ -174,6 +174,11 @@ export const emotesCategoryIds = {
   FRECENTS: 'frecents',
 };
 
+export const tipIds = {
+  EMOTE_MENU_ALT_FAVORITE_EMOTE: 'clickAltToFavoriteEmote',
```
night, line 178: why do we need key codes in these identifiers? just name them what they are for?

```
emoteMenuFavorite
emoteMenuPreventClose
```

src/modules/emote_menu/components/Menu.jsx
```diff
@@ -90,7 +95,9 @@ export default function EmoteMenu({triggerRef, appendToChat}) {
 
   return (
     <>
-      <Header className={styles.header} value={search} onChange={setSearch} onHide={onHide} selected={selected} />
+      <TipWhisper>
```
night, line 98: rather than making a tooltip, i'd prefer to designate a part of the menu container to showing tips. We could allocate a small 16px box at the bottom to show tips in, if available.

src/modules/emote_menu/components/TipWhisper.jsx
```diff
@@ -0,0 +1,67 @@
...
+  const [tooltip, setTooltip] = useState(null);
+
+  useEffect(() => {
+    const mostValuableTooltip = calcMostValuableTooltip([
+      tips.getTip(tipIds.EMOTE_MENU_SHIFT_MULTIPLE_EMOTES),
+      tips.getTip(tipIds.EMOTE_MENU_ALT_FAVORITE_EMOTE),
+    ]);
+
+    // we either don't show the tooltip cause there is none or randomly don't as to avoid being obnoxious
+    if (mostValuableTooltip == null && Math.random() < RANDOM_DONT_SHOW_TOOLTIP) {
```
night, line 40: I don't think we should randomly show the tip or not. the point is to get a user to try it, and since it goes away once they do I think that's good enough.

src/modules/emote_menu/components/TipWhisper.jsx
```diff
@@ -0,0 +1,67 @@
...
+      return;
+    }
+
+    setTooltip(mostValuableTooltip);
+    const timeout = setTimeout(() => triggerRef.current.open(), DELAY_SHOW_TOOLTIP);
```
night, line 45: I don't think we need a delay if we're allocating space to it.

src/utils/tips.js
```diff
@@ -0,0 +1,48 @@
+import storage from '../storage.js';
+
+class Tips {
+  constructor() {
+    this.tips = storage.get('tips');
```
night, line 5: do we need a whole storage for this? could we store individual storage keys for these?

PR #4849, opened 2021-09-18
[Feature] Show current channel's emotes

dclstn:
image
image

### DIFF
--- a/src/modules/emote_menu/stores/twitch-emotes.js
+++ b/src/modules/emote_menu/stores/twitch-emotes.js
@@ -12,6 +12,18 @@ import {getCurrentChannel} from '../../../utils/channel.js';
 
 const AVAILABLE_EMOTES_FOR_CHANNEL_QUERY = `
   query AvailableEmotesForChannel($channelID: ID!) {
+    user(id: $channelID) {
+      id,
+      displayName,
+      profileImageURL(width: 300)
+      subscriptionProducts {
+        emotes {
+          id,
+          token,
+          type
+        }
+      }
+    }
     channel(id: $channelID) {
       self {
         availableEmoteSets {
@@ -96,6 +108,7 @@ export async function loadTwitchEmotes() {
     const [availableEmotesResponse] = await twitchApi.graphqlQuery([
       {query: AVAILABLE_EMOTES_FOR_CHANNEL_QUERY, variables: {channelID: currentChannel.id}},
     ]);
+
     data = availableEmotesResponse.data;
   } catch (e) {
     debug.log('failed to fetch twitch emotes', e);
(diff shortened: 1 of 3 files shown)


src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -105,7 +118,23 @@ export async function loadTwitchEmotes() {
...
+    }
+  }
+
+  const subscriptionProducts = {
+    id: user.id,
+    owner: user,
+    emotes: concatSubscriptionProducts,
+  };
+
+  for (const {owner, id, emotes} of [...data.channel.self.availableEmoteSets, subscriptionProducts]) {
```
night, line 137: what happens if a user is subscribed to the emote set already? is there duplicates? do we override?

dclstn: We override it, as seen in the second image

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -105,7 +118,23 @@ export async function loadTwitchEmotes() {
...
+    for (const emote of set.emotes) {
+      emote.locked = true;
+      concatSubscriptionProducts.push(emote);
+    }
+  }
+
+  const subscriptionProducts = {
+    id: user.id,
+    owner: user,
+    emotes: concatSubscriptionProducts,
```
night, line 134: could we simply this logic to:

```
const subscriptionProducts = user.subscriptionProducts
  .filter(({emoteSetID}) => emoteSetID != null)
  .map(
    ({emoteSetID, emotes}) => ({
      id: emoteSetID,
      owner: user,
      emotes: emotes.map((emote) => ({...emote, locked: true}))
    })
  )
```

src/modules/emote_menu/stores/twitch-emotes.js
night: is this a reliable mechanism? I think this relies on `uniqby`

src/modules/emote_menu/stores/twitch-emotes.js
```diff
@@ -105,7 +119,17 @@ export async function loadTwitchEmotes() {
...
 
-  for (const {owner, id, emotes} of data.channel.self.availableEmoteSets) {
+  const {user} = data;
+
+  const subscriptionSets = user.subscriptionProducts
+    .filter(({id}) => id != null)
+    .map(({id, emotes}) => ({
+      id,
+      owner: user,
+      emotes: emotes.map((emote) => ({...emote, locked: true})),
```
night, line 129: we need to figure out a way to dynamically render locks when rendering. as-is a user subscribing or following a channel would not show the locked emotes followed by the then unlocked emotes when they subscribe or follow.

PR #5046, opened 2021-11-26
[Feature] Add AlwaysSourceQuality Setting 

### LINKED ISSUE #5044: A setting that automatically sets the resolution of streams to source quality
Could you please add a setting that automatically sets the resolution of streams to source quality. This is already accomplished with the script referred to as "Enhance twitch" (https://greasyfork.org/en/scripts/375467-enhance-twitch) on greasyfork.com but adding this feature to betterttv minimizes the need to install extra scripts to use this feature.

whitecrownclown:
Hello,

I've implemented the setting mentioned in #5044. 

- It updates the localStorage with the twitch equivalent for the `source` quality. `{default: 'chunked'}`. Twitch will then use this value when opening a new page / navigating away.
- For the initial switch, if you're not navigating/refreshing, we retrieve the player qualities and search for the source (highest bitrate). We could assume it's always the first in the list or search by name, but I thought it would be safer this way.
- I'm also not entirely sure of the naming / description, suggestions are welcomed.

If this is something you want added, please provide feedback. @night 

Thanks!

Closes #5044

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -12,6 +12,7 @@ export const SettingIds = {
   TAB_COMPLETION_EMOTE_PRIORITY: 'tabCompletionEmotePriority',
   WHISPERS: 'whispers',
   SHOW_DIRECTORY_LIVE_TAB: 'showDirectoryLiveTab',
+  ALWAYS_SOURCE_QUALITY: 'alwaysSourceQuality',
   CHANNEL_POINTS_MESSAGE_HIGHLIGHTS: 'channelPointsMessageHighlights',
   EMOTE_MENU: 'clickTwitchEmotes',
   DARKENED_MODE: 'darkenedMode',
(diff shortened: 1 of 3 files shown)


32423d:
Thank you for creating this feature. I appreciate it. The setting could be called "Always Use Source Quality".  A possible description could be "Forces stream quality to use source quality".

whitecrownclown:
> Thank you for creating this feature. I appreciate it. The setting could be called "Always Use Source Quality". A possible description could be "Forces stream quality to use source quality".

Thanks! I've updated the name & description.

night:
I'm confused how this will actually help. BetterTTV persists settings in the same place Twitch does. Twitch already remembers your configured quality and applies it for you.

32423d:
> I'm confused how this will actually help. BetterTTV persists settings in the same place Twitch does. Twitch already remembers your configured quality and applies it for you.

It seems the browser saves the configured quality in the browser cookies and not in the account settings of Twitch. If the browser cookies were to be cleared it would also clear the betterttv configuration that is saved but if you were to open a save file for the betterttv configuration with the AlwaysSourceQuality saved in the betterttv configuration you wouldn't have to set the quality to source yourself which would save a few clicks.

night:
In the current state I don't think this option makes sense since the settings would be wiped on cookie clear. We can re-evaluate should this change down the road.

whitecrownclown:
Thanks for the update, I'll close this PR for now.

PR #5047, opened 2021-11-26
[Feature] Add Show Homepage Carousel Setting

### LINKED ISSUE #4987: Highlighted Streamers Frontpage
hi all!

please give us an option to blackbox or completely disable the highlighted streamers on the top of the frontpage. not recommended ones further below. just the highlights. there is stop autoplay, but that only does half the trick.

once the paid-option arrives to highlight streams it becomes inevitable anyway to have such a function i guess.

https://ibb.co/j3SVB4K

i usually jump on the frontpage to get a grip on what's going on atm on twitch. recommended channels etc.

but the wild mix of featured streamers, that i actually never ever click on, is a bit annoying.

if it's recommendations based on my likings: it's really bad. if it's not, it has no justification to be there for me.

thank you!

(tried to grey out the streamers and their names - not shitting on anyone in particular)

whitecrownclown:
Hello,

- Added a new setting `Show Homepage Carousel` (defaults true)
- Grouped together with the Auto Play setting for the homepage video player (same thing as the carousel) in a Homepage module
- Fixed an issue regarding the disabled Auto Play setting for the homepage video player (by muting it as well), you can hear the stream for a brief moment before it stops.

Closes #4987 

Thanks!

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -30,6 +30,7 @@ export const SettingIds = {
   CHAT: 'chat',
   AUTO_PLAY: 'autoPlay',
   USERNAMES: 'usernames',
+  SHOW_HOMEPAGE_CAROUSEL: 'showHomepageCarousel'
 };
 
 export const CategoryTypes = {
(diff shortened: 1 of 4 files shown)


night:
Thanks for the PR. Unfortunately this is not a change we can permit. Please see my response at https://github.com/night/betterttv/issues/4987#issuecomment-981152462

PR #5058, opened 2021-11-30
[Feature] Mark Chat As Read

### LINKED ISSUE #2049: Feature Request: Hotkey to mark chat as read
Would be super helpful, as a streamer, to be able to have a keybind that marks your entire chat as read (grays it out?), or places some sort of marker, which makes it easier to quickly glance at chat and find the first new message.

whitecrownclown:
Hello,

- Added a new setting that allows for double click on the chat list in order to "mark" the current messages as read. Currently done by lowering the opacity.

Closes #2049


night:
I don't think this implementation makes sense at this time. The notion of a read message doesn't really have much use given how ephemeral messages are. The usefulness of this is limited to smaller chats, but having to manually mark messages makes this feature a pain to use rather than solves a problem. Typical chat clients rely on scroll state to determine read state, but because streamers are in-game we don't really know what they have read. Really only a hotkey style approach may work here, as otherwise we'd require them to leave their window context to chat context.

PR #5059, opened 2021-11-30
[Feature] Configurable number of pinned highlights

### LINKED ISSUE #901: Pinned highlighed messages: Set maximum amount pinned
Currently there is the ability to make pinned messages time out after 1 minute... but it would be really useful to set how many you want pinned.

I use several chat windows while streaming and some are too small to have more than 5 pinned. One is significantly bigger and I could use having ~15 pinned.

whitecrownclown:
Hello,

- Added a new setting where you can input the number of pinned higlights instead of the current 10.

Closes #901


src/modules/settings/components/settings/PinnedHighlights.jsx
```diff
@@ -1,22 +1,35 @@
...
   const [value, setValue] = useStorageState(SettingIds.PINNED_HIGHLIGHTS);
+  const [pinnedHighlights, setPinnedHighlights] = useStorageState(SettingIds.NUMBER_OF_PINNED_HIGHLIGHTS);
 
   return (
     <Panel header="Pinned Highlights">
       <div className={styles.toggle}>
-        <p className={styles.description}>Pins your last ten highlighted messages above chat</p>
+        <p className={styles.description}>
+          Pins your last{' '}
+          <Input
```
night, line 18: i'm not a fan of inlining settings into the description. Please use two setting rows within a Pinned Highlights panel instead.

PR #5075, opened 2021-12-11
[Feature] Show Locked Channel Emotes

dclstn:
Show locked emotes within the emote menu


src/modules/emote_menu/utils/youtube-emotes.js
```diff
@@ -80,12 +76,15 @@ export async function loadYouTubeEmotes() {
...
     const categoryEmotes = categoryEmojis[channelId].map(
-      ({emojiId: emoteId, image, shortcuts}) =>
+      ({emojiId: emoteId, image, shortcuts, isLocked = false}) =>
         new Emote({
           id: emoteId,
           category,
           channel: categoryName,
           code: shortcuts.find((shortcut) => !shortcut.startsWith(':_')) || shortcuts[0],
+          metadata: {
+            locked: isLocked,
```
night, line 86: `locked` here should be a predicate which checks if the user is subscribed or not on render.

src/modules/emote_menu/utils/twitch-emotes.js
```diff
@@ -123,7 +162,10 @@ export async function loadTwitchEmotes() {
           '2x': TWITCH_EMOTE_CDN(emoteId, '2.0', isDark),
           '4x': TWITCH_EMOTE_CDN(emoteId, '3.0', isDark),
         },
-        metadata: {type},
+        metadata: {
+          type,
+          locked,
```
night, line 167: locked here should be a predicate checking if the user is subscribed or not, or following or not (depends on emote type) when rendering.

src/modules/emote_menu/utils/twitch-emotes.js
```diff
@@ -102,9 +116,34 @@ export async function loadTwitchEmotes() {
...
+  // so we don't rely on uniqBy https://github.com/night/betterttv/pull/4849#discussion_r711787270
+  const emoteSets = data.channel.self.availableEmoteSets;
+
+  const availableChannelEmotes = flatten(
+    data.channel.self.availableEmoteSets
+      .filter(({owner}) => owner?.id === data.user.id)
+      .map((channel) => channel.emotes)
+  );
+
+  const unavailableChannelEmotes = flatten(data.user.subscriptionProducts.map(({emotes}) => emotes))
```
night, line 128: we spoke about this offline, but can we use uniqby to do this but explicitly with a comment?

src/modules/emote_menu/utils/twitch-emotes.js
```diff
@@ -102,9 +116,34 @@ export async function loadTwitchEmotes() {
...
+    .filter(({id}) => availableChannelEmotes.find(({id: emoteId}) => emoteId === id) == null)
+    .map((emote) => ({...emote, locked: true}));
+
+  if (unavailableChannelEmotes.length > 0) {
+    emoteSets.push({
+      emotes: unavailableChannelEmotes,
+      id: EmoteCategories.UNAVAILABLE_CHANNEL_EMOTES,
```
night, line 135: rather than use a separate category, could we instead sort them within the same category?

src/modules/emote_menu/components/EmoteMenu.jsx
```diff
@@ -34,6 +34,10 @@ export default function EmoteMenu({toggleWhisper, appendToChat, onSetTip}) {
         return;
       }
 
+      if (emote.metadata && emote.metadata.isLocked()) {
```
night, line 37: if it's locked, why are you able to click on it or hover it when using keyboard shortcuts. ideally clicking on locked elements should not result in a click event unless you're undoing some action, like unfavoriting.

src/modules/emote_menu/components/Preview.jsx
```diff
@@ -1,14 +1,20 @@
...
 
-export default function Preview({emote, isFavorite}) {
+export default function Preview({emote}) {
   if (emote == null) return null;
 
+  let icon = null;
+  if (emote.metadata && emote.metadata.isLocked()) {
+    icon = Icons.UNLOCK;
+  }
+  if (icon == null && emoteMenuViewStore.hasFavorite(emote)) {
```
night, line 14: this could be an else if

src/modules/emote_menu/utils/twitch-emotes.js
```diff
@@ -113,6 +170,20 @@ export async function loadTwitchEmotes() {
         newToken = emoteToken;
       }
 
+      let predicate = () => false;
+      if (['SUBSCRIPTIONS', 'FOLLOWER'].includes(type)) {
+        // eslint-disable-next-line react/function-component-definition
+        predicate = () => {
```
night, line 176: rather than defining a new method for each emote in memory, we could write a single predicate that works for all of them, and then pass in an emote id, right?

src/modules/emote_menu/utils/twitch-emotes.js
```diff
@@ -113,6 +170,20 @@ export async function loadTwitchEmotes() {
         newToken = emoteToken;
       }
 
+      let predicate = () => false;
+      if (['SUBSCRIPTIONS', 'FOLLOWER'].includes(type)) {
```
night, line 174: is this logic future proof? should we be checking for any emote which isn't in the available emote sets?

src/modules/emote_menu/utils/twitch-emotes.js
```diff
@@ -81,6 +113,16 @@ function getCategoryForSet(setId, owner) {
   }
 }
 
+function isLocked(setId, locked) {
```
night, line 116: nit: `defaultValue`

src/modules/emote_menu/utils/youtube-emotes.js
```diff
@@ -33,16 +33,26 @@ function getCategoryForChannelId(channelId, categoryName) {
   }
 }
 
+function getLiveChat() {
+  const liveChatContinuation = window.ytInitialData?.continuationContents?.liveChatContinuation;
+  const liveChatRenderer = window.ytInitialData?.contents?.liveChatRenderer;
```
night, line 38: i believe `window.ytInitialData` may go stale. we should maybe prefer line 39

night (approved):

PR #5076, opened 2021-12-12
[Feature] Emote Autocomplete

dclstn:
### Things that need fixed
1. Not mounting correctly on chat reload (twitch)
2. Preserve twitch input state

https://user-images.githubusercontent.com/43322006/155929112-5b55756e-4889-4965-a148-86046c2d3d04.mp4


src/modules/emote_autocomplete/components/Emotes.jsx
```diff
@@ -0,0 +1,92 @@
...
+const shortenName = (displayName) => displayName.split(' ')[0];
+
+let navigationCallback;
+function setNavigationCallback(callback) {
+  navigationCallback = callback;
+}
+
+export default function Emotes({chatInputElement, repositionPopover, autocomplete}) {
+  const [emotes, setEmotes] = useChatInput(chatInputElement);
+  const shortEmotes = useMemo(() => emotes.slice(0, calcMaxHeight()), [emotes]);
```
night, line 33: what happens when a resize occurs? do we care?

src/modules/emote_autocomplete/hooks/ChatInput.jsx
```diff
@@ -0,0 +1,39 @@
+import {useEffect, useState} from 'react';
+import emoteMenuViewStore from '../../../common/stores/emote-menu-view-store.js';
+import {getAutocompletable} from '../../../utils/autocomplete.js';
+
+function handleChatInput() {
+  const focusedWord = getAutocompletable();
+
+  if (focusedWord != null) {
+    const strippedFocusedWord = focusedWord.replace(/\s|:/g, '');
```
night, line 9: is this reliable? should this be returned parsed from the function above?

src/modules/emote_autocomplete/hooks/ChatInput.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+export default function useChatInput(chatInputElement) {
+  const [emotes, setEmotes] = useState([]);
+
+  useEffect(() => {
+    function keydownCallback() {
+      const updateEmotes = () => setEmotes(handleChatInput());
+
+      if (!emoteMenuViewStore.isLoaded()) {
+        emoteMenuViewStore.once('updated', updateEmotes);
```
night, line 25: every key down will subscribe to this, which adds a ton of additional listeners that will call updateEmotes when the load occurs. is there a way to detect this and load once only on hook mount? should the hook be listening itself?

src/modules/emote_autocomplete/twitch/EmoteAutcomplete.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+      mountedNode = whisperContainer;
+    }
+
+    this.show();
+  }
+
+  show() {
+    const autocompleteContainer = document.querySelector(EMOTE_AUTOCOMPLETE_CONTAINER_SELECTOR);
+    const emoteMenuEnabled = settings.get(SettingIds.EMOTE_AUTOCOMPLETE);
```
night, line 65: we seem to load the setting twice, when we could pass this in as a variable

src/modules/emote_autocomplete/twitch/EmoteAutcomplete.jsx
```diff
@@ -0,0 +1,102 @@
...
+      return;
+    }
+
+    listener = domObserver.on(TWITCH_AUTOCOMPLETE_CONTAINER_SELECTOR, (node, isConnected) => {
+      if (!isConnected) {
+        return;
+      }
+
+      if (isEmoteAutocompletable()) {
+        node.parentElement.remove();
```
night, line 98: should this use css instead?

src/modules/emote_autocomplete/twitch/EmoteAutcomplete.jsx
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    autocompleteContainer.classList.toggle(styles.hideEmoteAutocomplete, !emoteMenuEnabled);
+  }
+
+  autocomplete({code}) {
+    const currentValue = twitch.getChatInputValue();
+    const parts = currentValue.split(' ');
+    parts.pop();
+    parts.push(code);
```
night, line 78: can you explain this logic? is it safe/reliable to pop the last value?

src/modules/emote_autocomplete/components/EmoteWhisper.jsx
```diff
@@ -0,0 +1,22 @@
...
+        open
+        placement={null}
+        speaker={
+          <AutocompletePopover
+            chatInputElement={chatInputElement}
+            boundingQuerySelector={boundingQuerySelector}
+            autocomplete={autocomplete}
+          />
+        }>
+        <span />
```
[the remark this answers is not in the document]
dclstn: rsuite requires a html element as a child

src/modules/emote_autocomplete/components/Emotes.jsx
dclstn: do you think the user will resize the window whilst writing a message with autocomplete open?

src/modules/emote_autocomplete/components/Emotes.jsx
```diff
@@ -0,0 +1,120 @@
...
+  }, [emotes, selected]);
+
+  if (!open || emotes.length === 0) {
+    return null;
+  }
+
+  return (
+    <div className={styles.emotes} ref={localRef}>
+      <EmotesHeader getAutocomplete={getAutocomplete} />
+      <AutoSizer disableHeight>
```
night, line 97: we do we need this? could we show the first n results instead? as a user types more characters the list becomes more filtered, which gives them choices closer to what they want. we do not want people to browse the entire emote list here, but rather for them to press an arrow key or tab no more than a few times to get where they need to go

src/modules/emote_autocomplete/components/Emotes.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+  if (currentHeight > window.innerHeight - BOTTOM_OFFSET) {
+    currentHeight = window.innerHeight - BOTTOM_OFFSET;
+  }
+
+  return currentHeight;
+}
+
+let navigationCallback;
+function setNavigationCallback(callback) {
```
night, line 30: could we set this in the component itself rather than as a singleton?

src/modules/emote_autocomplete/components/AutocompletePopover.jsx
```diff
@@ -0,0 +1,58 @@
...
+import classNames from 'classnames';
+import Popover from 'rsuite/Popover';
+import mergeRefs from 'react-merge-refs';
+import styles from './AutocompletePopover.module.css';
+import Emotes from './Emotes.jsx';
+import repositionPopover from '../../../utils/popover.js';
+import useResize from '../../../common/hooks/Resize.jsx';
+
+const TOP_PADDING = 2;
+const DEFAULT_POPOVER_WIDTH = 300;
```
night, line 11: why 300? should the popover be more dynamically in width? have we considered different sized embeds?

src/modules/chat_tab_completion/index.js
```diff
@@ -55,6 +56,10 @@ class ChatTabcompletionModule {
   }
 
   onKeydown(e, includeUsers = true) {
+    if (emoteAutocomplete.isEnabled()) {
```
night, line 59: it isn't really `enabled`, but rather it is active

src/modules/emote_autocomplete/twitch/EmoteAutocomplete.jsx
```diff
@@ -0,0 +1,90 @@
...
+        whisperContainer
+      );
+
+      mountedNode = whisperContainer;
+    }
+
+    this.show();
+  }
+
+  show() {
```
night, line 54: this function name does not perform the action which is occurring.

src/modules/emote_autocomplete/twitch/EmoteAutocomplete.jsx
```diff
@@ -0,0 +1,90 @@
...
+    const emoteMenuEnabled = settings.get(SettingIds.EMOTE_AUTOCOMPLETE);
+
+    if (!emoteMenuEnabled && mountedNode != null) {
+      ReactDOM.unmountComponentAtNode(mountedNode);
+    }
+  }
+
+  isEnabled() {
```
night, line 62: `isActive`

src/modules/emote_autocomplete/twitch/EmoteAutocomplete.jsx
```diff
@@ -0,0 +1,90 @@
...
+    return settings.get(SettingIds.EMOTE_AUTOCOMPLETE) && this.getAutocomplete() != null;
+  }
+
+  autocomplete({code}) {
+    const currentValue = twitch.getChatInputValue();
+
+    const lastSpaceIndex = currentValue.lastIndexOf(' ');
```
night, line 69: what happens when there is not a space? isn't that a negative index?

src/modules/emote_autocomplete/twitch/EmoteAutocomplete.jsx
```diff
@@ -0,0 +1,90 @@
...
+    if (!emoteMenuEnabled && mountedNode != null) {
+      ReactDOM.unmountComponentAtNode(mountedNode);
+    }
+  }
+
+  isEnabled() {
+    return settings.get(SettingIds.EMOTE_AUTOCOMPLETE) && this.getAutocomplete() != null;
+  }
+
+  autocomplete({code}) {
```
night, line 66: `replaceChatInputPartialEmote` might be clearer, as it is replacing the partial emote code in the chat input

src/modules/emote_autocomplete/twitch/EmoteAutocomplete.jsx
```diff
@@ -0,0 +1,90 @@
...
+    const currentValue = twitch.getChatInputValue();
+
+    const lastSpaceIndex = currentValue.lastIndexOf(' ');
+    const newValue = `${currentValue.substring(0, lastSpaceIndex)} ${code}`;
+
+    twitch.setChatInputValue(newValue, true);
+  }
+
+  getAutocomplete() {
```
night, line 75: this name is not super clear. the logic is really `getChatInputPartialEmote`

src/modules/emote_autocomplete/hooks/ChatInput.jsx
```diff
@@ -0,0 +1,48 @@
...
+    return [];
+  }
+
+  const strippedFocusedWord = focusedWord.replace(/\s|:/g, '');
+  const foundEmotes = emoteMenuViewStore.search(strippedFocusedWord);
+
+  return foundEmotes.map((emote) => emote.item);
+}
+
+export default function useChatInput(chatInputElement, getAutocomplete) {
```
night, line 15: this hook name doesn't make much sense. `useChatInput` implies you are retrieving the chat input, but that is not what is happening here. I would recommend refactoring this logic:

1. rename to `useChatInputPartialEmote`
2. make this hook responsible solely for propagating and returning input changes corresponding to partial emotes in the chat.
3. move store listening logic to another hook, if need be. more ideally, consider making a hook for listening to a store and unlistening from it on unmount and using it where we need updates to the store.

usage could look like:
```
const partialEmoteCode = useChatInputPartialEmote(chatInputElement, getChatInputPartialEmoteCode);
const emotes = useEmoteMenuViewStore((emotes) => emotes.search(partialEmoteCode).map((emote) => emote.item))
```

src/modules/emote_menu/components/EmoteMenuPopover.jsx
```diff
@@ -21,47 +23,10 @@ const EmoteMenuPopover = React.forwardRef(
...
-
-    useEffect(() => {
-      function handleResize() {
-        repositionPopover();
-        // Twitch animates chat moving on zoom changes
-        setTimeout(repositionPopover, 500);
-      }
-
-      window.addEventListener('resize', handleResize);
+    const reposition = () => repositionPopover(localRef, boundingQuerySelector, TOP_PADDING);
```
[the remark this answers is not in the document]
dclstn: imo this is too specific to this component as we also call `reposition()` on emotes (matches) length change

src/modules/emote_menu/components/EmoteMenuPopover.jsx
```diff
@@ -37,11 +31,7 @@ const EmoteMenuPopover = React.forwardRef(
           full>
           <EmoteMenu
             toggleWhisper={toggleWhisper}
-            appendToChat={(...args) => {
-              const result = appendToChat(...args);
-              repositionPopover();
```
dclstn, line 77: this was doing nothing

src/modules/emote_autocomplete/twitch/EmoteAutocomplete.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+  isActive() {
+    return settings.get(SettingIds.EMOTE_AUTOCOMPLETE) && this.getChatInputPartialEmote() != null;
+  }
+
+  replaceChatInputPartialEmote({code}) {
+    const currentValue = twitch.getChatInputValue();
+
+    const lastSpaceIndex = currentValue.lastIndexOf(' ');
+    const newValue = lastSpaceIndex < 0 ? code : `${currentValue.substring(0, lastSpaceIndex)} ${code}`;
```
night, line 64: im curious: why not use the same regex for getting the input, and replacing the input?

```
const PARTIAL_EMOTE_REGEX = /:([a-z0-9-_.:]+?)(:|$)/i;
...
retest.exec(':test:') -> [':test:', 'test', ':', index: 0, input: ':test:', groups: undefined]
'test :testa:'.replace(retest, 'lol') -> 'test lol'
```

src/modules/emote_autocomplete/components/Emotes.jsx
```diff
@@ -0,0 +1,152 @@
...
+  useEffect(() => {
+    function handleOutsideClick(event) {
+      if (localRef.current == null) {
+        return;
+      }
+
+      setOpen(localRef.current.contains(event.target));
+    }
+
+    window.addEventListener('mousedown', handleOutsideClick);
```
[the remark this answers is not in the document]
dclstn: this is intentional

PR #5111, opened 2022-01-01
[Feature] Add Settings Menu to Youtube

dclstn:
- [x] Embed a BetterTTV Settings button into the chat app dropdown
- [x] Fix styling for modal & standalone menu
- [x] Migrate settings to be platform specific
- [x] Apply site styling

Notes:
- Where is the ideal placement for the settings menu button?
- If a category in the settings menu has no settings available to it, should we hide the category?
- If the button was to be located in the chat window, should it open the settings menu within the chat window or outer?
- `loadModuleForPlatforms` is now being used for react components, should this be renamed to just `loadForPlatforms` for less ambiguity?
- Should `./src/modules/settings/components/settings/Youtube.jsx` still be accessible on twitch or only youtube's menu?
- Unlike twitch, the emote menu is enabled by default. Should this default be reverted with the addition of this menu?

https://user-images.githubusercontent.com/43322006/149255080-d1647993-82ce-4931-8a81-25d12e404fa3.mp4


src/modules/settings/components/Window.jsx
```diff
@@ -28,12 +28,15 @@ function Page(props) {
 }
 
 function isWindowSmallStandaloneChat() {
-  return window.innerWidth < 600 && window.location.pathname.endsWith('/chat');
+  return (
```
night, line 31: you can't do this. site specific logic should be in a youtube/twitch component folder (see emote menu)

src/modules/settings/components/Window.jsx
```diff
@@ -28,12 +28,15 @@ function Page(props) {
...
+    (window.innerWidth < 600 && window.location.pathname.endsWith('/chat')) ||
+    window.location.pathname.endsWith('/live_chat')
+  );
 }
 
 function Window({setHandleOpen}) {
   const [page, setPage] = useState(PageTypes.CHAT_SETTINGS);
-  const [open, setOpen] = useState(false);
+  const [open, setOpen] = useState(true);
```
night, line 39: is this really what we want? in general a modal component should be loaded but not open until its prop changes to open, no?

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,76 @@
+import $ from 'jquery';
```
night, line 1: dont use jquery in new files

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,76 @@
...
+      nativeButtonContainer.insertBefore(buttonContainer, nativeButtonContainer.firstChild);
+
+      if (mountedNode != null) {
+        ReactDOM.unmountComponentAtNode(mountedNode);
+      }
+
+      ReactDOM.render(
+        <Button
+          onClick={() => {
+            console.log('here');
```
night, line 61: ?

src/modules/settings/components/settings/BlacklistKeywords.jsx
```diff
@@ -47,9 +48,13 @@ function BlacklistKeywords() {
   );
 }
 
-export default registerComponent(BlacklistKeywords, {
-  settingId: SettingIds.BLACKLIST_KEYWORDS,
-  name: 'Blacklist Keywords',
-  category: CategoryTypes.CHAT,
-  keywords: ['black', 'list', 'keywords', 'banned', 'remove', 'hide'],
-});
+loadModuleForPlatforms([
```
night, line 51: since a majority of settings apply specifically to Twitch, does it make sense to have twitch-specific settings in a twitch folder instead of having this boilerplate for each?

night (commented):
> Where is the ideal placement for the settings menu button?

probably in the same places as on twitch. user dropdown and the settings dropdown in chat.

> If a category in the settings menu has no settings available to it, should we hide the category?

ideally we should only show settings and categories applicable to youtube

> If the button was to be located in the chat window, should it open the settings menu within the chat window or outer?

given the limited functionality, it seems reasonable to only show it within the iframe

> Should ./src/modules/settings/components/settings/Youtube.jsx still be accessible on twitch or only youtube's menu?

it isnt clear what this means. in general when the module (settings in this case) loads, only files related to that loads

> Unlike twitch, the emote menu is enabled by default. Should this default be reverted with the addition of this menu?

if users are not complaining we can leave it alone. the big thing is toggling it off should leave it off.

src/modules/settings/components/Window.jsx
dclstn: for testing*

src/modules/emote_menu/youtube/EmoteMenu.jsx
```diff
@@ -90,15 +91,16 @@ export default class EmoteMenuModule {
...
 
   show(visible) {
-    // TODO: take into account emote menu setting in the future
-    const nativeContainer = document.querySelector(NATIVE_EMOTE_MENU_BUTTON_CONTAINER_SELECTOR);
-    if (nativeContainer != null) {
-      nativeContainer.classList.toggle(styles.hideEmoteMenuButton, visible);
-    }
+    const emoteMenuEnabled = settings.get(SettingIds.EMOTE_MENU);
+
+    // const nativeContainer = document.querySelector(NATIVE_EMOTE_MENU_BUTTON_CONTAINER_SELECTOR);
```
night, line 96: what is going on here?

src/modules/settings/components/Sidenav.jsx
```diff
@@ -19,36 +19,38 @@ import FontAwesomeSvgIcon from '../../../common/components/FontAwesomeSvgIcon.js
...
-            icon={<Icon as={FontAwesomeSvgIcon} fontAwesomeIcon={faColumns} />}>
-            <p>Emote Dashboard</p>
-          </Nav.Item>
-          <Nav.Item eventKey={PageTypes.CHANGELOG} icon={<Icon as={FontAwesomeSvgIcon} fontAwesomeIcon={faBox} />}>
-            <p>Changelog</p>
-          </Nav.Item>
-        </Nav>
-        <Nav pullRight activeKey={value} onSelect={onChange} className={styles.bottom}>
+      <Sidenav.Body className={styles.body}>
+        <div>
```
night, line 23: why is this in a div?

src/modules/settings/components/settings/twitch/AutoPlay.jsx
```diff
@@ -2,40 +2,40 @@ import React from 'react';
...
       <div className={styles.setting}>
-        <p className={styles.description}>Control Auto Play on various Twitch video players.</p>
+        <p>Control Auto Play on various Twitch video players.</p>
         <CheckboxGroup
           value={Object.values(AutoPlayFlags).filter((value) => hasFlag(autoplay, value))}
           onChange={(value) => setAutoplay(value.length > 0 ? value.reduce((a, b) => a | b) : 0)}>
           <Checkbox key="fpPlayer" value={AutoPlayFlags.FP_VIDEO}>
             <p>Front-page Player</p>
-            <p className={styles.description}>Enable auto play for homepage video player</p>
+            <p>Enable auto play for homepage video player</p>
```
night, line 22: why did we lose description css on all these? does the title not now match the description format?

src/modules/settings/components/settings/youtube/YouTube.jsx
```diff
@@ -1,10 +1,10 @@
 import React, {useEffect} from 'react';
```
night, line 1: this file is a twitch setting, and must remain there.

src/modules/settings/index.js
```diff
@@ -1,5 +1,9 @@
 import {PlatformTypes} from '../../constants.js';
 import {loadModuleForPlatforms} from '../../utils/modules.js';
-import SettingsModule from './Settings.jsx';
+import TwitchSettingsModule from './twitch/Settings.jsx';
+import YoutubeSettingModule from './youtube/Settings.jsx';
 
-export default loadModuleForPlatforms([PlatformTypes.TWITCH, () => new SettingsModule()]);
+export default loadModuleForPlatforms(
```
night, line 6: isn't the `Settings.jsx` file already doing this?

src/modules/settings/styles/_rsuite.less
```diff
@@ -20,6 +20,20 @@
   width: 800px;
   height: 500px;
 
+  color: var(--bttv-rs-text-primary);
+  font-size: 1.3rem;
+  line-height: 1.5;
+
+  p {
```
night, line 27: we should not set styles for html elements. in general this will cause pain. the goal of css modules is to prevent styles from bleeding between components.

src/modules/settings/youtube/DropdownButton.module.css
```diff
@@ -0,0 +1,38 @@
+.button {
+  width: 290px;
```
night, line 2: why are we setting width and height of this? width and height should really only be set for fixed dimension containers or images/icons. button height ideally is derived based on its contents + padding + margin

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,160 @@
...
+    if (bttvPanel == null) {
+      const panel = document.createElement('div');
+      panel.setAttribute('id', 'bttvSettingsPanel');
+      const chatApp = document.querySelector('#chat');
+
+      if (chatApp == null) {
+        return;
+      }
+
+      chatApp.append(panel);
```
night, line 61: why do we append this to chat? shouldn't we append this to body so it doesn't get dropped on a rerender?

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,160 @@
...
+  constructor() {
+    this.renderSettings();
+
+    // TODO: re-enable this when we have settings like theatre mode...
+    // domObserver.once(AVATAR_BUTTON_SELECTOR, (node, isConnected) => {
+    //   if (!isConnected) {
+    //     return;
+    //   }
+
+    //   node.addEventListener('click', this.loadDropdownButton);
```
night, line 37: this logic looks error prone, as observing once will not account for re-renders

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,160 @@
...
+      ReactDOM.render(<Modal setHandleOpen={setHandleOpen} />, panel);
+      mountedPanel = panel;
+    }
+  }
+
+  loadChatAppMenuButton() {
+    const buttonContainer = document.querySelector(BTTV_CHAT_DROPDOWN_BUTTON_CONTAINER_SELECTOR);
+
+    if (buttonContainer == null) {
+      const itemsContainer = document.querySelector(CHAT_SETTINGS_DROPDOWN_ITEMS_SELECTOR).parentElement;
```
night, line 76: isn't this already included in the `node` from the callback of the dom observer?

src/utils/window.js
```diff
@@ -47,3 +47,7 @@ export function isPopout() {
     return true;
   }
 }
+
+export function isExternalWindow() {
+  return window.location.pathname.endsWith('/chat') || window.location.pathname.endsWith('/live_chat');
```
night, line 52: we should probably separate these by platform rather than combining them

src/modules/settings/components/Sidenav.jsx
dclstn: for `display: flex` to group the logo and top navigation options

src/modules/settings/index.js
dclstn: ?

src/modules/global_css/rsuite.less
```diff
@@ -20,6 +20,15 @@
   width: 800px;
   height: 500px;
 
+  color: var(--bttv-rs-text-primary);
+  font-size: 1.3rem;
+  line-height: 1.5;
+
+  a {
+    color: var(--color-text-link);
```
night, line 28: whose variable is this?

src/modules/settings/components/Window.jsx
```diff
@@ -29,7 +30,7 @@ function Page(props) {
 }
 
 function isWindowSmallStandaloneChat() {
-  return window.innerWidth < 600 && window.location.pathname.endsWith('/chat');
+  return window.innerWidth < 600 && isExternalWindow();
```
night, line 33: we should pass in a predicate instead, rather than making a util that works for all sites. this also isn't checking for an external window, it's checking for a standalone chat window

src/modules/settings/components/Settings.jsx
```diff
@@ -1,15 +1,40 @@
...
-import {Components} from './Store.jsx';
 import FontAwesomeSvgIcon from '../../../common/components/FontAwesomeSvgIcon.jsx';
+import {loadModuleForPlatforms} from '../../../utils/modules.js';
+import {PlatformTypes} from '../../../constants.js';
 
-const settings = Object.values(Components).sort((a, b) => a.name.localeCompare(b.name));
+let settings = [];
+
+(async () => {
+  await loadModuleForPlatforms(
```
night, line 14: rather than loading these here, load them from within the platform's settings file and pass in settings as a prop.

`twitch/Settings.jsx` and `youtube/Settings.jsx`

src/modules/emote_menu/youtube/EmoteMenu.jsx
```diff
@@ -90,15 +91,16 @@ export default class EmoteMenuModule {
   }
 
   show(visible) {
-    // TODO: take into account emote menu setting in the future
+    const emoteMenuEnabled = settings.get(SettingIds.EMOTE_MENU);
+
     const nativeContainer = document.querySelector(NATIVE_EMOTE_MENU_BUTTON_CONTAINER_SELECTOR);
     if (nativeContainer != null) {
-      nativeContainer.classList.toggle(styles.hideEmoteMenuButton, visible);
+      nativeContainer.classList.toggle(styles.hideEmoteMenuButton, visible && emoteMenuEnabled);
```
night, line 98: can you explain these changes? why accept `visible` at all if we're simply overwriting its value? why are we getting the setting from storage twice in this path (load -> settings.get -> show -> settings.get)?

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,160 @@
...
+let mountedDropdownButton;
+let mountedChatDropdownButton;
+
+let menuItemsListener = null;
+
+export default class SettingsModule {
+  constructor() {
+    this.renderSettings();
+
+    // TODO: re-enable this when we have settings like theatre mode...
```
night, line 31: if the code is not in use, then we should delete it.

src/modules/settings/youtube/Settings.jsx
night: bump?

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,160 @@
...
+        return;
+      }
+
+      menuItemsListener();
+      menuItemsListener = null;
+
+      const dropdownButton = document.querySelector(BTTV_DROPDOWN_BUTTON_CONTAINER_SELECTOR);
+
+      if (dropdownButton == null || dropdownButton.hasAttribute('hidden')) {
+        const items = document.querySelectorAll(DROPDOWN_MENU_ITEMS_SELECTOR)[1];
```
night, line 126: why do we pull the first instead of the 0th? if there are multiple items, is this a reliable selector?

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,160 @@
...
+        ReactDOM.unmountComponentAtNode(mountedChatDropdownButton);
+      }
+
+      ReactDOM.render(
+        <DropdownButton
+          onClick={() => {
+            handleOpen(true);
+
+            // close the dropdown menu when the modal is opened
+            const menuButton = document.querySelector('#overflow #button');
```
night, line 97: why do we inline some selectors but not others?

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,160 @@
...
+      ReactDOM.render(<Modal setHandleOpen={setHandleOpen} />, panel);
+      mountedPanel = panel;
+    }
+  }
+
+  loadChatAppMenuButton(node) {
+    const buttonContainer = document.querySelector(BTTV_CHAT_DROPDOWN_BUTTON_CONTAINER_SELECTOR);
+
+    if (buttonContainer == null) {
+      const itemsContainer = node.parentElement;
```
night, line 76: is this a reliable selector? how do we know the parent is what we want?

src/modules/settings/youtube/Settings.module.css
```diff
@@ -0,0 +1,82 @@
...
+    .bttv-rs-btn-primary:focus .bttv-rs-btn-primary.bttv-rs-btn-active,
+    .bttv-rs-btn-icon-with-text.bttv-rs-btn-primary:hover > .bttv-rs-icon,
+    .bttv-rs-btn-icon-with-text.bttv-rs-btn-primary:focus > .bttv-rs-icon {
+      background-color: rgba(255, 255, 255, 1);
+      color: rgba(100, 100, 100, 1);
+    }
+
+    .bttv-rs-checkbox-checked .bttv-rs-checkbox-wrapper .bttv-rs-checkbox-inner::before,
+    .bttv-rs-checkbox-indeterminate .bttv-rs-checkbox-wrapper .bttv-rs-checkbox-inner::before {
+      background-color: rgba(255, 255, 255, 1);
```
night, line 72: why do we use rgba for 100% opacity?

src/modules/settings/youtube/Settings.module.css
```diff
@@ -0,0 +1,82 @@
...
+      font-family: 'Helvetica Neue', Helvetica, Arial;
+    }
+
+    .bttv-rs-toggle-checked .bttv-rs-toggle-presentation {
+      background-color: rgba(255, 255, 255, 0.5);
+    }
+
+    .bttv-rs-sidenav-default .bttv-rs-sidenav-item.bttv-rs-sidenav-item-active,
+    .bttv-rs-sidenav-default .bttv-rs-dropdown-item.bttv-rs-dropdown-item-active,
+    .bttv-rs-sidenav-default .bttv-rs-dropdown.bttv-rs-dropdown-selected-within .bttv-rs-dropdown-toggle-icon {
```
night, line 51: can we parent these selectors?

```
.bttv-rs-sidenav-default {
  ...
}
```

src/modules/settings/youtube/Settings.module.css
```diff
@@ -0,0 +1,82 @@
...
+
+    .bttv-rs-btn-primary,
+    .bttv-rs-btn-icon-with-text.bttv-rs-btn-primary > .bttv-rs-icon {
+      background-color: rgba(255, 255, 255, 0.8);
+      color: rgba(100, 100, 100, 1);
+    }
+
+    .bttv-rs-btn-primary:active,
+    .bttv-rs-btn-primary:hover,
+    .bttv-rs-btn-primary:focus .bttv-rs-btn-primary.bttv-rs-btn-active,
```
night, line 63: is this selector correct? we have a primary button inside a primary, active button?

src/modules/settings/youtube/Settings.module.css
```diff
@@ -0,0 +1,82 @@
...
+  margin-right: 16px;
+}
+
+.chatAppDropdownButton {
+  position: relative;
+  bottom: 10px;
+  height: 36px;
+}
+
+/* TODO: swap to chakra :) */
```
night, line 38: rm. this is not a useful comment

src/modules/emote_menu/youtube/EmoteMenu.jsx
dclstn: This is copied logix from twitch/Settings.jsx, I'm unsure why we even call this.show(true) at the end of renderSettings() as we call this.show(true) anyway in the onMount prop on line 39.

was this final this.show(true) a fail-safe?

src/modules/settings/youtube/Settings.module.css
dclstn: You tell me, this is what was done in `youtube/EmoteMenu.module.css`, I'll update them all to use rgb when 100%

src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -85,19 +85,16 @@ export default class EmoteMenuModule {
       );
       mountedNode = buttonContainer;
     }
-
-    this.show(true);
   }
 
   show(visible) {
     const legacyContainer = document.querySelector(LEGACY_BTTV_EMOTE_PICKER_BUTTON_CONTAINER_SELECTOR);
-    const emoteMenuEnabled = settings.get(SettingIds.EMOTE_MENU);
```
night, line 94: why did you remove this (and the above)? there is no way to disable the emote menu as-is. we can probably update the above to be:

```
if (!emoteMenuEnabled) {
  this.show(false);
}

```

src/modules/emote_menu/youtube/EmoteMenu.jsx
```diff
@@ -85,12 +86,9 @@ export default class EmoteMenuModule {
 
       mountedNode = buttonContainer;
     }
-
-    this.show(true);
```
night, line 89: same issue as above

src/modules/global_css/rsuite.less
```diff
@@ -20,6 +20,10 @@
   width: 800px;
   height: 500px;
 
+  a {
+    text-decoration: none;
```
night, line 24: does the underline still appear when you hover links? i think the intended behavior is that links by default do not have underscores, but do upon hover.

src/modules/settings/components/settings/twitch/DarkMode.jsx
```diff
@@ -19,9 +19,9 @@ function DarkMode() {
   );
 }
 
-export default registerComponent(DarkMode, {
+registerComponent(DarkMode, {
   settingId: SettingIds.DARKENED_MODE,
   name: 'Dark Mode',
-  category: CategoryTypes.CHANNEL,
```
night, line 25: should this remain under channel?

src/modules/settings/pages/Changelog.jsx
```diff
@@ -77,7 +77,7 @@ function Changelog({onClose}) {
       <div className={styles.content}>
         <PanelGroup>
           <Panel header={<h3>Changelog</h3>}>
-            <p className={styles.description}>A list of recent updates and patches to BetterTTV.</p>
+            <p>A list of recent updates and patches to BetterTTV.</p>
```
night, line 80: why was this classname removed

src/modules/settings/styles/popout.module.css
```diff
@@ -1,12 +1,15 @@
-.default {
-  position: relative;
-  background-color: #292d33;
+.standaloneChatWindow {
+  position: absolute;
+  top: 0px;
+  left: 0px;
+  background-color: var(--bttv-rs-bg-overlay);
   height: 100vh;
```
night, line 6: instead of setting height and width, let's just set top, left, right, bottom to `0`

src/utils/modules.js
```diff
@@ -4,8 +4,8 @@ import {getPlatform} from './window.js';
 
 const currentPlatform = getPlatform();
 
-export function loadModuleForPlatforms(...platformConfigurations) {
-  for (const [platformType, callback] of platformConfigurations) {
+export async function loadModuleForPlatforms(...platformConfigurations) {
```
night, line 7: we can undo these changes

src/modules/settings/components/Settings.jsx
```diff
@@ -1,15 +1,17 @@
...
 import InputGroup from 'rsuite/InputGroup';
 import AutoComplete from 'rsuite/AutoComplete';
 import * as faSearch from '@fortawesome/free-solid-svg-icons/faSearch';
-import {Components} from './Store.jsx';
 import FontAwesomeSvgIcon from '../../../common/components/FontAwesomeSvgIcon.jsx';
 
-const settings = Object.values(Components).sort((a, b) => a.name.localeCompare(b.name));
+let settings = [];
+
+export async function importStoreCallback() {
```
night, line 11: instead of exporting this, can we instead use `useEffect` to load this below?

```
let settings = null;
async function loadSettings() {
  if (settings != null) {
    return settings;
  }
  const {Components} = await import('./Store.jsx');
  settings = Object.values(Components).sort((a, b) => a.name.localeCompare(b.name));
}

...

const [settings, setSettings] = useState([]);
useEffect(() => {
  loadSettings.then(settings => setSettings(settings));
});
```

src/modules/settings/twitch/Settings.jsx
```diff
@@ -13,12 +14,21 @@ let mountedNode;
...
     domObserver.on('a[data-test-selector="user-menu-dropdown__settings-link"]', () => {
       this.renderSettingsMenuOption();
     });
   }
 
+  async load() {
+    // eslint-disable-next-line import/no-unresolved
+    await import('../components/settings/global/*.jsx');
+    // eslint-disable-next-line import/no-unresolved
+    await import('../components/settings/twitch/*.jsx');
```
night, line 27: should these be loaded by `'../components/Settings.jsx'`?

src/modules/settings/youtube/Settings.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+      this.loadChatAppMenuButton(node);
+    });
+  }
+
+  async load() {
+    // eslint-disable-next-line import/no-unresolved
+    await import('../components/settings/global/*.jsx');
+    // eslint-disable-next-line import/no-unresolved
+    await import('../components/settings/youtube/*.jsx');
```
night, line 38: ditto

src/modules/settings/twitch/Settings.jsx
dclstn: https://github.com/night/betterttv/pull/5111#discussion_r789347202

Issue #5203, opened 2022-02-12
Idea: Preview stream when hovering sidebar streamers


Far-Se:
Feature Preview:

https://user-images.githubusercontent.com/20853986/153699626-d776d8c8-cc70-452b-abf7-f6b116238121.mp4

My code for it, the random size is to dodge cache so you see current stream and not outdated:
```js
setInterval( ()=>{
    //Tooltip img
	let el = document.querySelector('.online-side-nav-channel-tooltip__body');
	if(el) {
		let user = document.querySelector('[data-test-selector="recommended-channel"]:hover') ||  document.querySelector('[data-test-selector="followed-channel"]:hover') ;
		if(user) {
			user = user.getAttribute('href').split('/')[1];
            if(!el.querySelector('img')) {
                const size = 235 + Math.floor(Math.random() * 13);
                el.querySelector('div>p').insertAdjacentHTML('afterBegin',`<img src="https://static-cdn.jtvnw.net/previews-ttv/live_user_${user}-440x${size}.jpg" data-ping="248" height="102" width="187">`);
                1 || el.querySelector('div>p').insertAdjacentHTML('afterBegin',`<iframe src="https://player.twitch.tv/?channel=${user}&muted=true&parent=www.twitch.tv&controls=false&quality=mobile" height="150" width="280" allowfullscreen="true"></iframe>`);
            }
		}
	}
},200);
```
The iframe line works but  twitch gives a random 1min screen with some advert to watch it on twitch, so you can remove it, I left it there for an alternative.

PR #5205, opened 2022-02-12
[Feature] Stream Hover Preview

dclstn:
in response to #5203


src/modules/stream_preview/Preview.module.css
```diff
@@ -0,0 +1,14 @@
+.previewContainer {
+  margin-bottom: 4px;
+  margin-top: 4px;
+
+  img {
+    border-radius: 4px;
+  }
+
+  background-color: var(--bttv-rs-bg-card);
```
night, line 9: why is bg color all alone at the bottom when it could be at the top

src/modules/stream_preview/Preview.module.css
```diff
@@ -0,0 +1,14 @@
...
+}
+
+.placeholder {
```
night, line 12: this doesnt look used

src/modules/stream_preview/Preview.jsx
```diff
@@ -0,0 +1,18 @@
+import React, {useEffect, useState} from 'react';
+import styles from './Preview.module.css';
+
+function createPreviewImage(user) {
+  const dateTime = new Date().toISOString(); // cache prevention
+  return `https://static-cdn.jtvnw.net/previews-ttv/live_user_${user}-440x248.jpg?dateTime=${dateTime}`;
```
night, line 6: can we get this from react instead? I believe there exists a stream object in the parent chain

src/modules/stream_preview/StreamPreview.jsx
```diff
@@ -0,0 +1,53 @@
...
+      }
+
+      const hoveredElement = document.querySelector(HOVERED_SIDEBAR_SELECTOR);
+      const user = hoveredElement.getAttribute('href').split('/')[1];
+      const panel = document.createElement('div');
+      panel.setAttribute('id', PREVIEW_SELECTOR);
+      ReactDOM.render(<StreamPreview username={user} usernameCallback={usernameCallback} />, panel);
+      node.prepend(panel);
+
+      tooltipObserver.observe(node, {
```
night, line 43: instead of manually creating a tooltip observer, could we instead improve our dom watcher selector so that it can detect changes in the tooltip's child.

src/modules/stream_preview/Preview.jsx
dclstn: couldn't find a prop with this data

night:
Twitch used to have this on their site. Honestly I don't think it's all that useful in its current form, as folks that aren't sure what to watch are likely browsing thumbnails in the directory. Channels you already follow you're already accustomed to what to expect.

PR #5227, opened 2022-02-22
[Feature] Draggle & Droppable Emoji Categories

dclstn:
QA Testing Checklist

- [x] Twitch Dark Mode Chat + Emote Menu
- [x] Twitch Light Mode Chat + Emote Menu
- [x] Twitch Dark Mode Settings Menu (from Chat Settings + from User Dropdown)
- [x] Twitch Light Mode Settings Menu (from Chat Settings + from User Dropdown)
- [x] Twitch POPOUT Dark Mode Chat + Emote Menu + Chat Settings
- [x] Twitch POPOUT Light Mode Chat + Emote Menu + Chat Settings
- [x] Twitch Moderation Chat Dark Mode Chat + Emote Menu + Chat Settings
- [x] Twitch Moderation Chat Light Mode Chat + Emote Menu + Chat Settings
- [x] Twitch Dashboard Chat Dark Mode Chat + Emote Menu + Chat Settings
- [x] Twitch Dashboard Chat Light Mode Chat + Emote Menu + Chat Settings
- [x] YouTube Dark Mode Chat + Emote Menu
- [x] YouTube Light Mode Chat + Emote Menu
- [x] YouTube Dark Mode Settings Menu (from Chat Settings)
- [x] YouTube Light Mode Settings Menu (from Chat Settings)
- [x] YouTube POPOUT Dark Mode Chat + Emote Menu + Chat Settings
- [x] YouTube POPOUT Light Mode Chat + Emote Menu + Chat Settings
- [ ] YouTube Studio Dark Mode Chat + Emote Menu + Chat Settings

PR #5305, opened 2022-03-30
[Fix]: restrict icon size for youtube channel icon categories

rileymiller:
This fixes an issue with the Youtube emote categories introduced during https://github.com/night/betterttv/pull/5227

The previous pull request changed some of the styles used by the `BrandedImage` component and removed the brand image icon's width/height restrictions.

This had the side effect of allowing the channel image size to be unbounded resulting in the following behavior for Ludwig's channel:
image

Which is pulling the channel image avatar, a 300x300 image which causes the overflow, from here: https://api.betterttv.net/3/cached/users/youtube/UCrPseYLGpNygVi34QpGNqpA

Steps to Repro:
1. Install BetterTTV v7.4.24
2. Head to this scheduled Ludwig stream: https://www.youtube.com/watch?v=Xe5KJINZGRA
3. Enable the BetterTTV menu
4. Open the Emote menu

Here's a screenshot of the emote picker after the fix:
image

### DIFF
--- a/src/modules/emote_menu/components/Icons.jsx
+++ b/src/modules/emote_menu/components/Icons.jsx
@@ -32,8 +32,8 @@ const twitchGamingLogo = {
 
 function BrandedImage({src, alt, brandSrc}) {
   return (
-    <div className={styles.icon}>
-      <img src={src} alt={alt} className={styles.iconBorderRadius} />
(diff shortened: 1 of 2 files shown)


xKito15:
have the same problem

PR #5472, opened 2022-07-05
Shadow Ban Detection

Saalvage:
Adds the ability to toggle message confirmations. Uses an additional, anonymous, `WebSocket` to this end.

I'm not entirely happy with how messages are being selected for confirmation (simply using the latest message with equal content). Using ids would be very greatly preferable, but on the react side of things the ids appear to be simple timestamps and it seems like I'm unable to get appropriate absolute timestamps from the `WebSocket`'s events. Should still be good enough.

Easy way to trigger a shadow ban for testing purposes: Spam two alternating messages quickly. Good idea to use a VPN + throwaway account though as shadow bans can affect IP and account (even then, they shouldn't stick around for too long though).

P.S.: The .gitattributes file makes sure that no CRLF line endings come into existence on Windows, because those make the linter go insane with errors.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto
+*.js text eol=lf
+*.jsx text eol=lf
--- a/src/constants.js
+++ b/src/constants.js
@@ -26,6 +26,7 @@ export const SettingIds = {
   DELETED_MESSAGES: 'deletedMessages',
   BLACKLIST_KEYWORDS: 'blacklistKeywords',
   HIGHLIGHT_KEYWORDS: 'highlightKeywords',
+  SHADOW_BAN_DETECTOR: 'shadowBanDetector',
   SIDEBAR: 'sidebar',
   EMOTES: 'emotes',
   CHAT: 'chat',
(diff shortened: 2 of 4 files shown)


Saalvage:
Shadow bans are clearly a very user-unfriendly feature, tricking users into thinking they are partaking in chat while their messages are simply being filtered out.
From what I've heard there is no one to one correlation between spamming and shadow bans, they seem to be given out for arbitrary reasons.
This features does also not aid spamming, as it doesn't allow for circumventing shadow bans, but merely alerting users to the fact that their messages are not being received.
If you really want to argue that this counteracts functionality Twitch deliberately puts into place then the same argument could be made against preventing chat clearing, showing deleted messages and auto-claiming channel points.

night:
Thanks again for the request. I haven't personally seen evidence of shadow bans on Twitch over the years. If any exists it could be indicative of Twitch chat dropping messages due to service unavailability or being held by automod in streams for manual moderator review. Twitch already actively prevents users from sending messages too quickly, which avoids any kind of automated chat bans which occur on the platform level. Because of this, I'm not sure how necessary a change like this is in BetterTTV.

PR #5615, opened 2022-09-11
Add option to show and highlight deleted messages

maxwlang:
This is something I've been wanting for a while in BTTV. I think FFZ or some other extension had something similar to this a while ago and I always found it pretty useful.

Switched an `if` and `if else` block into a `switch` block with a fall-through to reduce code duplication with this change, but fine with reverting that and writing a little extra code.

Hoping this is something that could be added, but if not no sweat. Thanks!

### DIFF
--- a/src/modules/chat_deleted_messages/index.js
+++ b/src/modules/chat_deleted_messages/index.js
@@ -62,9 +62,11 @@ class ChatDeletedMessagesModule {
 
   handleDelete(name, targetMessageId) {
     const deletedMessages = settings.get(SettingIds.DELETED_MESSAGES);
-    if (deletedMessages !== DeletedMessageTypes.HIDE
-        && deletedMessages !== DeletedMessageTypes.SHOW
-        && deletedMessages !== DeletedMessageTypes.HIGHLIGHT) {
+    if (
+      deletedMessages !== DeletedMessageTypes.HIDE &&
+      deletedMessages !== DeletedMessageTypes.SHOW &&
+      deletedMessages !== DeletedMessageTypes.HIGHLIGHT
+    ) {
       return false;
     }
     const messages = findAllUserMessages(name, targetMessageId);
@@ -76,6 +78,7 @@ class ChatDeletedMessagesModule {
           break;
         case DeletedMessageTypes.HIGHLIGHT:
           $message.css('background-color', '#dc2626');
+        // eslint-disable-next-line no-fallthrough
         case DeletedMessageTypes.SHOW:
           $message.toggleClass(CHAT_LINE_DELETED_CLASS, true);
           /* eslint-disable-next-line func-names */
@@ -84,6 +87,7 @@ class ChatDeletedMessagesModule {
             $link.removeAttr('href');
           });
           $message.find(CHAT_LINE_CLIP_CARD_SELECTOR).remove();
+          break;
         default:
           break;
       }
--- a/src/modules/chat_deleted_messages/index.js
+++ b/src/modules/chat_deleted_messages/index.js
@@ -10,6 +10,7 @@ const CHAT_LINE_LINK_SELECTOR = 'a.link-fragment';
 const CHAT_LINE_CLIP_CARD_SELECTOR = '.chat-card';
 const CHAT_LINE_DELETED_CLASS = 'bttv-chat-line-deleted';
 
+
 function findAllUserMessages(name, targetMessageId) {
   return Array.from(document.querySelectorAll(CHAT_LINE_SELECTOR)).filter((node) => {
     const message = twitch.getChatMessageObject(node);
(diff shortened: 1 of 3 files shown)


src/modules/chat_deleted_messages/index.js
```diff
@@ -57,24 +57,34 @@ class ChatDeletedMessagesModule {
...
-          $link.removeAttr('href');
-        });
-        $message.find(CHAT_LINE_CLIP_CARD_SELECTOR).remove();
+      switch (deletedMessages) {
+        case DeletedMessageTypes.HIDE:
+          $message.hide();
+          break;
+        case DeletedMessageTypes.HIGHLIGHT:
+          $message.css('background-color', '#dc2626');
+        // eslint-disable-next-line no-fallthrough
```
dclstn, line 76: nit: why not this? to avoid using an eslint-ignore

```js
case DeletedMessageTypes.HIGHLIGHT:
case DeletedMessageTypes.SHOW:
   if (deletedMessages === DeletedMessageTypes.HIGHLIGHT) {
      $message.css('background-color', '#dc2626');
   }
   ...

Comment (maxwlang) [reply src/modules/chat_deleted_messages/index.js]: Good call, change made. Thanks!

File: src/modules/chat_deleted_messages/index.js [hunk 2]
```diff
@@ -57,24 +57,34 @@ class ChatDeletedMessagesModule {
...
-          const $link = $(this);
-          $link.removeAttr('href');
-        });
-        $message.find(CHAT_LINE_CLIP_CARD_SELECTOR).remove();
+      switch (deletedMessages) {
+        case DeletedMessageTypes.HIDE:
+          $message.hide();
+          break;
+        case DeletedMessageTypes.HIGHLIGHT:
+          $message.css('background-color', '#dc2626');
```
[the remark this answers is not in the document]
maxwlang: I believe my changes should have implemented this correctly using `.addClass`. Thanks!

src/modules/chat_deleted_messages/index.js
```diff
@@ -10,6 +10,7 @@ const CHAT_LINE_LINK_SELECTOR = 'a.link-fragment';
 const CHAT_LINE_CLIP_CARD_SELECTOR = '.chat-card';
 const CHAT_LINE_DELETED_CLASS = 'bttv-chat-line-deleted';
 
+
```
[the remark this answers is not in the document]
maxwlang: Removed, thanks!

src/modules/chat_deleted_messages/index.js
```diff
@@ -57,24 +57,35 @@ class ChatDeletedMessagesModule {
 
   handleDelete(name, targetMessageId) {
     const deletedMessages = settings.get(SettingIds.DELETED_MESSAGES);
-    const showDeletedMessages = deletedMessages === DeletedMessageTypes.SHOW;
-    const hideDeletedMessages = deletedMessages === DeletedMessageTypes.HIDE;
-    if (!hideDeletedMessages && !showDeletedMessages) {
+    if (
+      deletedMessages !== DeletedMessageTypes.HIDE &&
+      deletedMessages !== DeletedMessageTypes.SHOW &&
+      deletedMessages !== DeletedMessageTypes.HIGHLIGHT
```
night, line 63: `![DeletedMessageTypes.HIDE, DeletedMessageTypes.SHOW, DeletedMessageTypes.HIGHLIGHT].includes(deletedMessages)`

src/modules/chat_deleted_messages/index.js
```diff
@@ -57,24 +57,35 @@ class ChatDeletedMessagesModule {
...
+          }
+          $message.toggleClass(CHAT_LINE_DELETED_CLASS, true);
+          /* eslint-disable-next-line func-names */
+          $message.find(CHAT_LINE_LINK_SELECTOR).each(function () {
+            const $link = $(this);
+            $link.removeAttr('href');
+          });
+          $message.find(CHAT_LINE_CLIP_CARD_SELECTOR).remove();
+          break;
+        default:
```
night, line 87: nit: we don't need this at all

src/modules/chat_deleted_messages/index.js
```diff
@@ -57,24 +57,35 @@ class ChatDeletedMessagesModule {
...
-        });
-        $message.find(CHAT_LINE_CLIP_CARD_SELECTOR).remove();
+      switch (deletedMessages) {
+        case DeletedMessageTypes.HIDE:
+          $message.hide();
+          break;
+        case DeletedMessageTypes.HIGHLIGHT:
+        case DeletedMessageTypes.SHOW:
+          if (deletedMessages === DeletedMessageTypes.HIGHLIGHT) {
+            $message.addClass('bttv-highlighted');
```
night, line 77: we should call `markHighlighted` from the highlight module, which avoids compatibility issues in future module updates.

night (changes requested):

src/modules/chat_deleted_messages/index.js
maxwlang: Hah, yep.

src/modules/chat_deleted_messages/index.js
maxwlang: Interesting approach, changed.

src/modules/chat_deleted_messages/index.js
maxwlang: How would you recommend accessing the highlight module? I could export the `ChatHighlightBlacklistKeywordsModule` class, but I feel like there's a better way to do this. Possibly something relating to accessing the module loaded by `loadModuleForPlatforms`? Thanks!

night (approved):

PR #5652, opened 2022-10-02
Reverse VOD chat direction when setting is enabled

### LINKED ISSUE #5612: Feature Request - Reverse chat for VODs
I think it'd be neat if the "reverse chat" feature applied to VODS as well,

MartijnWelker:
Fixes #5612

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build
 dist
 .DS_Store
 .vscode
+.idea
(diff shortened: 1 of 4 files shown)


night:
Thanks for the pull request. It doesn't appear to work properly. While the ordering is changed in the style level, new chat lines do not appear to populate in reverse order when playing the feed. In addition, scrolling in the opposite direction seems to pause chat replay.

MartijnWelker:
Ah good catch! I accidentally reverted a bit too much of my code I believe. I’ll fix it up tomorrow and re-request a review 🙂

MartijnWelker:
@night I might be misunderstanding how reversing of the chat should work, but with this feature enabled new chat messages should appear at the top, right? That is also what I'm observing on the PR branch. 
image

MartijnWelker:
@night I've fixed the order of new chat messages appearing and the automatic scrolling. 

I ran into a small issue with the scroll syncing where it would not properly stay synced to the top which I fixed by manually scrolling to the top of the container on every scroll event, but maybe you know a better way of handling this.

PR #5653, opened 2022-10-02
Add option to video seek using alt+scroll

### LINKED ISSUE #4827: Option to video seek using the Mousewheel Up/Down
I have multiple monitors running twitch and other things such as work and games.
I have a Chrome extension in Youtube (Youtube(TM) Mouse Controls) where I can simply move the mouse to the monitor playing youtube and use the mouse wheel to seek the video forward and back 5 seconds.

I have not found an extension for twitch that does the same thing, so I'm having to click and pause the twitch video to get focus, then un-pause it, and then reach out to the keyboard and press the arrow keys to seek... and at the same time the other app I was using on the other monitor has lost focus so I have to click that app to gain focus again. It's something that takes me 1 second to do in youtube, but 5-10 seconds in twitch.

It would be fantastic if I could simply use the mouse wheel Up/Down to seek the video when the mouse is hovering over the twitch video area, and scroll down the page when the mouse is hovering outside of the main video area .

MartijnWelker:
Fixes #4827


src/modules/settings/components/settings/twitch/ScrollVideoSeek.jsx
```diff
@@ -0,0 +1,27 @@
...
+import {registerComponent} from '../../Store.jsx';
+import {CategoryTypes, SettingIds} from '../../../../../constants.js';
+import styles from '../../../styles/header.module.css';
+import useStorageState from '../../../../../common/hooks/StorageState.jsx';
+
+function ScrollVideoSeek() {
+  const [value, setValue] = useStorageState(SettingIds.SCROLL_VIDEO_SEEK);
+
+  return (
+    <Panel header="Scroll Video Seek">
```
MartijnWelker, line 13: Not entirely sure how you would describe this feature. `Seek by scrolling` maybe? Currently not happy with the name.

src/modules/video_player/index.js
```diff
@@ -57,7 +57,32 @@ function handlePlayerClick() {
...
 
 function handlePlayerScroll(event) {
+  maybeSeek(event);
+  maybeControlVolume(event);
+}
+
+function maybeSeek (event) {
+  if (!settings.get(SettingIds.SCROLL_VIDEO_SEEK)) return;
+  // Non-Alt scrolling controls volume
+  if (!event.altKey) return;
```
MartijnWelker, line 67: Should we include the `alt` modifier to toggle between volume control and seeking? We could also not do this, and if a user has both enabled just let it both happen.

src/modules/settings/components/settings/twitch/ScrollVideoSeek.jsx
MartijnWelker: Updated it

src/modules/video_player/index.js
night: i would imagine if you are pressing alt it does not affect volume, and if you are not it only affects volume

src/modules/video_player/index.js
```diff
@@ -57,7 +57,31 @@ function handlePlayerClick() {
 }
 
 function handlePlayerScroll(event) {
+  maybeSeek(event);
+  maybeControlVolume(event);
+}
+
+function maybeSeek (event) {
```
night, line 64: these changes need to be formatted with prettier

src/modules/settings/components/settings/twitch/ScrollVideoSeek.jsx
```diff
@@ -0,0 +1,27 @@
...
+import styles from '../../../styles/header.module.css';
+import useStorageState from '../../../../../common/hooks/StorageState.jsx';
+
+function ScrollVideoSeek() {
+  const [value, setValue] = useStorageState(SettingIds.SCROLL_VIDEO_SEEK);
+
+  return (
+    <Panel header="Seek by scrolling">
+      <div className={styles.toggle}>
+        <p className={styles.description}>Enables alt+scrolling on the twitch player to seek through VODs</p>
```
night, line 15: this + title copy needs better casing + formatting

src/modules/video_player/index.js
```diff
@@ -57,7 +57,31 @@ function handlePlayerClick() {
...
+  // Non-Alt scrolling controls volume
+  if (!event.altKey) return;
+
+  // Default seek time is 2 seconds for VODs
+  const delta = event.originalEvent.deltaY > 0 ? -2 : 2;
+
+  const currentPlayer = twitch.getCurrentPlayer();
+  if (!currentPlayer) return;
+
+  currentPlayer.seekTo(currentPlayer.getPosition() + delta);
```
night, line 75: how does this code function for live content?

src/modules/video_player/index.js
MartijnWelker: image
I'm getting a ton of linting errors when running prettier. Looks like maybe something with line endings on my end? I'm running on Windows.

src/modules/video_player/index.js
MartijnWelker: I've removed the possibility to do this in live content. It somewhat worked but was very unintuitive.

src/modules/settings/components/settings/twitch/ScrollVideoSeek.jsx
MartijnWelker: I've checked other settings and I think the casing should be correct now 👍

src/modules/video_player/index.js
MartijnWelker: Allright, that's how I made it now 👍

src/modules/video_player/index.js
MartijnWelker: Read into this a bit more. Looks like all files are formatted with unix line endings and prettier is configured to enforce that. However,  running `npm run eslint --fix` immediately converts it back to dos line endings for me. Does it make sense to keep this rule or should we disable it?

PR #5800, opened 2022-12-16
Emote Menu API

dclstn:
Allows the registration of new emote categories

PR will allow third-party extensions to register their emotes on BTTV's emote menu, e.g. #4840


src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
+/* eslint-disable import/prefer-default-export */
+// import {EmoteCategories, EmoteProviders} from '../../constants.js';
```
night, line 2: rm

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
...
+        id: emote.id,
+        category: {
+          id,
+          provider,
+          displayName,
+        },
+        code: emote.code,
+        images: {
+          '1x': emote.images['1x'],
+          '2x': emote.images['2x'],
```
night, line 71: how are other missing sizes handled today? what if these are not valid image urls?

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
...
+      if (emote.code == null || emote.code.length === 0) {
+        throw new Error('Emote code must be a non-empty string');
+      }
+
+      if (emote.images['1x'] == null || emote.images['1x'].length === 0) {
+        throw new Error('Emote must have a 1x image');
+      }
+
+      return new Emote({
+        id: emote.id,
```
night, line 62: isn't this missing `channel`?

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
...
+      });
+    });
+
+    emoteMenuViewStore.upsertRegisteredProviderCategory({
+      id,
+      provider,
+      displayName,
+      icon,
+      emotes: parsedEmotes,
+      channel,
```
night, line 83: perhaps this should be `channelId` to imply it is an identifier?

src/common/stores/emote-menu-view-store.js
```diff
@@ -169,6 +181,23 @@ class EmoteMenuViewStore extends SafeEventEmitter {
...
+    for (const provider of Object.values(registeredProviderCategories)) {
+      if (provider?.channel != null && provider?.channel !== currentChannel?.id) {
+        continue;
+      }
+      const category = createCategory(
+        provider.id,
+        provider.provider,
+        provider.displayName,
+        provider.channel == null
+          ? Icons.IMAGE(provider.icon.src, provider.icon.alt)
```
night, line 194: why do we need alt text? can't we just have `iconUrl` and generate an alt text?

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
...
+
+    if (displayName == null || displayName.length === 0) {
+      throw new Error('Display name must be a non-empty string');
+    }
+
+    if (channel != null && channel.length === 0) {
+      throw new Error('Channel must be a non-empty string');
+    }
+
+    if (RESERVED_EMOTE_CATEGORY_IDS.includes(id)) {
```
night, line 28: why do we permit providers to generate their own category identifiers? would it not be more wise to assign those and return them on creation? we could generate a uuid and return that. future changes would require the uuid be specified to update/delete.

src/common/api/emote-menu.js
night: side note: it looks like we're mis-setting `channel` on `Emote` in the emote menu's usage. we set it to a string instead of an object.

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
...
+// import {EmoteCategories, EmoteProviders} from '../../constants.js';
+import {EmoteCategories, EmoteProviders} from '../../constants.js';
+import Emote from '../../modules/emotes/emote.js';
+import emoteMenuViewStore from '../stores/emote-menu-view-store.js';
+
+const RESERVED_EMOTE_PROVIDER_IDS = Object.values(EmoteProviders);
+const RESERVED_EMOTE_CATEGORY_IDS = Object.values(EmoteCategories);
+
+export default {
+  upsertProvider({id, provider, displayName, icon, emotes: providerEmotes = [], channel = null}) {
```
night, line 11: they aren't adding a provider, they are adding an emote menu category, no?

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
...
+    if (id == null || id.length === 0) {
+      throw new Error('Category ID must be a non-empty string');
+    }
+
+    if (provider == null || provider.length === 0) {
```
night, line 16: if i were to specify a number, wouldn't this pass validation?

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,98 @@
...
+
+    if (channel != null && channel.length === 0) {
+      throw new Error('Channel must be a non-empty string');
+    }
+
+    if (RESERVED_EMOTE_CATEGORY_IDS.includes(id)) {
+      throw new Error(`Category ID ${id} is reserved`);
+    }
+
+    if (RESERVED_EMOTE_PROVIDER_IDS.includes(provider)) {
```
night, line 32: similarly for providers, would it be more wise to have a provider register their provider, and then upsert emote categories for themselves using that id?

src/common/api/emote-menu.js
dclstn: I will address this issue in future PR, once the current is merged.

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+  const {displayName} = provider;
+  return {
+    id: (global ? `${displayName}-channel` : `${displayName}-global`).toLowerCase(),
+    displayName: global ? `${displayName} Channel Emotes` : `${displayName} Global Emotes`,
+  };
+}
+
+function upsertCategory({id, displayName, providerId, emotes: providerEmotes = [], channelId = null}) {
+  if (providerEmotes.length == null || providerEmotes.length === 0) {
```
night, line 17: what if this is not an array? `providerEmotes='test'` would pass this check

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,93 @@
...
+    if (displayName == null || displayName.length === 0 || typeof displayName !== 'string') {
+      throw new Error('Display name must be a non-empty string');
+    }
+    if (iconSrc == null || iconSrc.length === 0 || typeof iconSrc !== 'string') {
+      throw new Error('Icon source must be a non-empty string');
+    }
+    return emoteMenuViewStore.registerProvider({displayName, iconSrc: new URL(iconSrc).toString()});
+  },
+
+  upsertGlobalCategory({providerId, emotes}) {
```
night, line 73: curious: why not have the value of `channelId` dictate if it's global or not?

src/common/api/emote-menu.js
```diff
@@ -0,0 +1,93 @@
...
+    upsertCategory({providerId, emotes, ...serializeProvider({providerId, global: true})});
+  },
+
+  upsertChannelCategory({providerId, emotes, channelId}) {
+    if (channelId == null || channelId.length === 0 || typeof channelId !== 'string') {
+      throw new Error('Channel ID must be a non-empty string');
```
night, line 79: it is a code smell that we do validation in two places in this path. it would be great to have validation in 1 place. combining these duplicate methods may help too

src/common/api/emote-menu.js
dclstn: Improves clarity, and also means we don't need to implement loose boolean logic and gives the possibility of adding more variants like personal emotes (if we choose to do so)

src/common/stores/emote-menu-view-store.js
```diff
@@ -112,6 +116,26 @@ class EmoteMenuViewStore extends SafeEventEmitter {
...
+  getRegisteredProvider(providerId) {
+    return registeredProviders[providerId];
+  }
+
+  upsertRegisteredProviderCategory(categoryId, category) {
+    registeredProviderCategories[categoryId] = category;
+    this.updateProviders();
+  }
+
+  deleteRegisteredProviderCategory({id}) {
```
night, line 134: why is this accepting an object?

src/common/stores/emote-menu-view-store.js
```diff
@@ -112,6 +116,26 @@ class EmoteMenuViewStore extends SafeEventEmitter {
...
+    const registeredProviderId = uuidv4();
+    registeredProviders[registeredProviderId] = provider;
+    return registeredProviderId;
+  }
+
+  getRegisteredProvider(providerId) {
+    return registeredProviders[providerId];
+  }
+
+  upsertRegisteredProviderCategory(categoryId, category) {
```
night, line 129: should `id` be part of the category?

src/common/stores/emote-menu-view-store.js
```diff
@@ -169,6 +193,27 @@ class EmoteMenuViewStore extends SafeEventEmitter {
...
+      const provider = this.getRegisteredProvider(registeredCategory.providerId);
+      if (provider == null) {
+        continue;
+      }
+      const category = createCategory(
+        registeredCategoryId,
+        registeredCategory.providerId,
+        `${provider.displayName} ${registeredCategory.channelId == null ? 'Global' : 'Channel'}`,
+        registeredCategory.channelId == null
+          ? Icons.IMAGE(provider.iconSrc, registeredCategoryId)
```
night, line 210: instead of using an id as the alt text, why not use its name?

```
const categoryName = `${provider.displayName} ${registeredCategory.channelId == null ? 'Global' : 'Channel'}`;
```

this also doesn't support i18n though, so that's a bit of a pickle.

src/common/stores/emote-menu-view-store.js
```diff
@@ -169,6 +193,27 @@ class EmoteMenuViewStore extends SafeEventEmitter {
...
+        continue;
+      }
+      const provider = this.getRegisteredProvider(registeredCategory.providerId);
+      if (provider == null) {
+        continue;
+      }
+      const category = createCategory(
+        registeredCategoryId,
+        registeredCategory.providerId,
+        `${provider.displayName} ${registeredCategory.channelId == null ? 'Global' : 'Channel'}`,
```
night, line 208: this doesn't support i18n, so that's a bit of a pickle. we prob need two strings

```
{providerName} Global
```

```
{providerName} Channel
```

src/common/stores/emote-menu-view-store.js
night, line 210: rather than have the category id as alt text, could we use its display name? `provider.displayName`

src/common/stores/emote-menu-view-store.js
night: this hasn't been addressed, so marking unresolved.

src/common/stores/emote-menu-view-store.js
night: this hasn't been addressed, so marking unresolved.

`upsertRegisteredProviderCategory(category)` should be the function declaration

night (changes requested):
review the two comments left on previously closed comments

night (approved):

PR #5874, opened 2023-01-19
Fix highlighted keywords not showing 3-digit hex colors

### LINKED ISSUE #5871: Highlighted keywords don't show 3-digit hex colors
**Describe the bug**
If a keyword is given a 3-digit hex as the highlight color, the message is instead highlighted in black (with the forced 30% transparency, `rgba(0, 0, 0, 0.3)`). The color appears correctly in the color picker preview window (and the highlight sound still plays if enabled).

**To Reproduce**

1. Add a new highlighted keyword.
2. In the keyword's Advanced Settings, change the Highlight Color to a 3-digit hex color (e.g. `f00`).
3. When the keyword is used, the message is highlighted in black (with the forced 30% transparency, `rgba(0, 0, 0, 0.3)`)

**Expected behavior**
When the keyword is used, the highlight color should appear (with the forced 30% transparency).

**Screenshots**
image
The colors of the messages, in order:
1. Alternate background color (`#1f1925`)
2. Default background color (`#18181b`)
3. `#f00` (incorrectly appears as `rgba(0, 0, 0, 0.3)`)
4. `#ff0000` (appears as `rgba(255, 0, 0, 0.3)`)

image
The color `#f00` being shown correctly in the color picker.

**Desktop:**
- OS: Microsoft Windows 10 Pro, Version 10.0.19045 Build 19045
- Browser: Chrome, Version 109.0.5414.75 (Official Build) (64-bit)
- Version: BetterTTV 7.5.0

fowwo:
Fixes #5871.

I've just added a check for 3-digit hex colors in `getRgb(color)` similar to the check for 6-digit hex colors.

### DIFF
--- a/src/utils/colors.js
+++ b/src/utils/colors.js
@@ -159,6 +159,15 @@ function getRgb(color) {
     };
   }
 
+  const hex3Regex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i.exec(color);
+  if (hex3Regex != null) {
+    return {
+      r: parseInt(hex3Regex[1], 16) * 17,
+      g: parseInt(hex3Regex[2], 16) * 17,
+      b: parseInt(hex3Regex[3], 16) * 17,
+    };
+  }
+
   const rgbRegex = /^rgb\((\d+),\s*(\d+),\s*(\d+)\)$/i.exec(color);
   if (rgbRegex != null) {
     return {
--- a/src/utils/colors.js
+++ b/src/utils/colors.js
@@ -162,9 +162,9 @@ function getRgb(color) {
   const hex3Regex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i.exec(color);
   if (hex3Regex != null) {
     return {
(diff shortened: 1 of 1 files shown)


src/utils/colors.js
```diff
@@ -159,6 +159,15 @@ function getRgb(color) {
     };
   }
 
+  const hex3Regex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i.exec(color);
+  if (hex3Regex != null) {
+    return {
+      r: parseInt(hex3Regex[1], 16) * 17,
```
night, line 165: this logic feels a bit unclear as a reader. it might be cleaner to simply `parseInt(hex3Regex[1] + hex3Regex[1], 16)` and so on

fowwo: Sure, let's do that. I agree that it's a bit unclear. The `* 17` seems wrong even though I know it's right. haha

night (approved):

PR #5882, opened 2023-01-21
Add logo to README

Void48:
also changed the Discord gh badge color to the new blurple: https://discord.com/branding


night:
Thanks for the PR! The logo doesn't appear to be transparent though, can that be updated? We probably also want to store it in the repo versus using a GitHub hosted URL that the repo does not control.

Void48:
> Thanks for the PR! The logo doesn't appear to be transparent though, can that be updated? We probably also want to store it in the repo versus using a GitHub hosted URL that the repo does not control.

I will look into it tomorrow 👍

Void48:
Should be good now @night
when you merge it it should be displayed in the README using the path to the file

README.md
```diff
@@ -1,6 +1,6 @@
-# BetterTTV
+# <img src="https://raw.githubusercontent.com/night/betterttv/master/.github/bttvlogotransparent.gif" height="50"> BetterTTV
```
Felanbird, line 1:
```suggestion
# <img src=".github/bttvlogotransparent.gif" height="50"> BetterTTV
```
This works as a cleaner approach, though I'm unsure if `.github` is the best place to drop this

night:
Thanks for the update. Is it possible to get the GIF's edges to be smoother? Maybe due to its size the edges look pretty jagged, and making it smaller may help. If not, I might propose using the logo without animation so we can achieve higher quality.

Void48:
> Thanks for the update. Is it possible to get the GIF's edges to be smoother? Maybe due to its size the edges look pretty jagged, and making it smaller may help. If not, I might propose using the logo without animation so we can achieve higher quality.

Dealing with the .gif format is pretty annoying. It's definitely possible but I don't think I want to spend any more time on this😕
I tried enabling light mode on GitHub to see if the edges are that noticeable but I think it's honestly fine if it's just for the README

README.md
Void48: thanks by the way :bowtie:

PR #5885, opened 2023-01-22
Add missing 7TV global emotes to autocomplete

NBruderman:


NBruderman:
Nevermind, I see global 7TV emotes are not supported yet

PR #5957, opened 2023-02-18
Support Emote Overlays

dclstn:
Addresses #5925

### DIFF
--- a/src/modules/emote_menu/components/EmoteButton.jsx
+++ b/src/modules/emote_menu/components/EmoteButton.jsx
@@ -4,7 +4,7 @@ import Emote from '../../../common/components/Emote.jsx';
 import styles from './EmoteButton.module.css';
 
 export default function EmoteButton({emote, onClick, onMouseOver, active}) {
-  const locked = emote?.metadata?.isLocked != null ? emote.metadata.isLocked() : false;
+  const locked = emote.metadata?.isLocked?.() ?? false;
 
   return (
     <button
--- a/src/modules/emote_menu/components/EmoteButton.jsx
+++ b/src/modules/emote_menu/components/EmoteButton.jsx
@@ -4,7 +4,7 @@ import Emote from '../../../common/components/Emote.jsx';
 import styles from './EmoteButton.module.css';
 
 export default function EmoteButton({emote, onClick, onMouseOver, active}) {
-  const locked = emote.metadata != null ? emote.metadata.isLocked() : false;
+  const locked = emote?.metadata?.isLocked != null ? emote.metadata.isLocked() : false;
 
   return (
     <button
--- a/src/modules/emote_menu/components/EmoteMenu.jsx
+++ b/src/modules/emote_menu/components/EmoteMenu.jsx
@@ -35,7 +35,7 @@ export default function EmoteMenu({toggleWhisper, appendToChat, onSetTip}) {
         return;
       }
 
-      if (emote.metadata && emote.metadata.isLocked()) {
+      if (emote?.metadata?.isLocked != null && emote.metadata.isLocked()) {
         return;
       }
 
--- a/src/modules/emote_menu/components/EmoteMenu.jsx
+++ b/src/modules/emote_menu/components/EmoteMenu.jsx
@@ -35,7 +35,7 @@ export default function EmoteMenu({toggleWhisper, appendToChat, onSetTip}) {
         return;
       }
 
-      if (emote?.metadata?.isLocked != null && emote.metadata.isLocked()) {
+      if (emote.metadata?.isLocked?.() ?? false) {
         return;
       }
 
--- a/src/modules/emote_menu/components/Preview.jsx
+++ b/src/modules/emote_menu/components/Preview.jsx
@@ -9,7 +9,7 @@ export default function Preview({emote}) {
   if (emote == null) return null;
 
   let icon = null;
-  if (emote.metadata != null && emote.metadata.isLocked()) {
+  if (emote?.metadata?.isLocked != null && emote.metadata.isLocked()) {
     icon = Icons.LOCK;
   } else if (emoteMenuViewStore.hasFavorite(emote)) {
     icon = Icons.STAR;
--- a/src/modules/emote_menu/components/Preview.jsx
+++ b/src/modules/emote_menu/components/Preview.jsx
@@ -9,7 +9,7 @@ export default function Preview({emote}) {
   if (emote == null) return null;
 
   let icon = null;
-  if (emote?.metadata?.isLocked != null && emote.metadata.isLocked()) {
+  if (emote.metadata?.isLocked?.() ?? false) {
     icon = Icons.LOCK;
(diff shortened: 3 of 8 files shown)


src/modules/seventv/utils.js
```diff
@@ -24,6 +24,9 @@ export function createEmote(id, code, animated, owner, category) {
       '2x_static': animated ? emoteUrl(id, '2x', true) : undefined,
       '4x_static': animated ? emoteUrl(id, '4x', true) : undefined,
     },
+    metadata: {
+      isOverlay: hasFlag(flags, 1 << 0),
```
ScrubN, line 28: I believe the overlay flag should be `1 << 8` (256), not `1 << 0` (1)

src/modules/emote_menu/components/EmoteButton.jsx
```diff
@@ -4,7 +4,7 @@ import Emote from '../../../common/components/Emote.jsx';
 import styles from './EmoteButton.module.css';
 
 export default function EmoteButton({emote, onClick, onMouseOver, active}) {
-  const locked = emote.metadata != null ? emote.metadata.isLocked() : false;
+  const locked = emote?.metadata?.isLocked != null ? emote.metadata.isLocked() : false;
```
night, line 7: you could combine these:

```
emote.metadata?.isLocked?.() ?? false
```

src/modules/emote_menu/components/EmoteMenu.jsx
```diff
@@ -35,7 +35,7 @@ export default function EmoteMenu({toggleWhisper, appendToChat, onSetTip}) {
         return;
       }
 
-      if (emote.metadata && emote.metadata.isLocked()) {
+      if (emote?.metadata?.isLocked != null && emote.metadata.isLocked()) {
```
night, line 38: you could combine these:

```
emote.metadata?.isLocked?.() ?? false
```

src/modules/emotes/emote.js
```diff
@@ -48,6 +48,9 @@ export default class Emote {
     if (shouldRenderStatic) {
       container.classList.add('bttv-animated-static-emote');
     }
+    if (this?.metadata?.isOverlay === true) {
```
night, line 51: `this` cannot be undefined/null, so you could change this to

```
metadata?.isOverlay === true
```

src/modules/emote_menu/components/Preview.jsx
```diff
@@ -9,7 +9,7 @@ export default function Preview({emote}) {
   if (emote == null) return null;
 
   let icon = null;
-  if (emote.metadata != null && emote.metadata.isLocked()) {
+  if (emote?.metadata?.isLocked != null && emote.metadata.isLocked()) {
```
night, line 12: you could combine these:

```
emote.metadata?.isLocked?.() ?? false
```

src/modules/emotes/style.css
```diff
@@ -98,6 +98,12 @@ img.emoji.yt-live-chat-text-message-renderer + .bttv-global-emo-5e76d399d6581c37
   margin: -0.5rem 0;
 }
 
+div[data-test-selector='emote-button'] + span[data-a-target='chat-message-text'] span .bttv-emote-overlay,
+div.bttv-emote + .bttv-emote-overlay,
+img.emoji.yt-live-chat-text-message-renderer + .bttv-emote-overlay {
+  margin-left: -32px;
```
night, line 104: how does this logic interplay today with wide emotes?

src/modules/seventv/channel-emotes.js
```diff
@@ -55,13 +55,14 @@ class SevenTVChannelEmotes extends AbstractEmotes {
...
           name: code,
+          flags,
           data: {listed, animated, owner},
         } of emotes) {
           if (!listed) {
             continue;
           }
 
-          this.emotes.set(code, createEmote(id, code, animated, owner, category));
+          this.emotes.set(code, createEmote(id, code, animated, owner, category, flags));
```
night, line 65: we also need to support the events below. i believe for those `visibility` has flag `1 << 7`.

we may want to consider passing a boolean instead of parsing flags in the `createEmote` utility

src/modules/seventv/utils.js
night: this comment is correct, the overlay flag appears to be `1 << 8`.

night (changes requested):

src/modules/emotes/style.css
ScrubN: In the 7TV V3 extension, it appears as though the emote token object has a field dedicated for overlay emotes and the tokenizer always keeps track of the last emote token, which makes interplay with wide emotes almost native since it's not relying on hard coded css values. https://github.com/SevenTV/ExtensionV3/blob/ea8f1d9e5488e14995b7bd7eaaa0d063dcd5f311/src/common/chat/Tokenizer.ts#L24-L35

I'm not sure how emotes are handled by the Bttv extension and I know both extensions use different frameworks, but maybe this information still helps?

NBruderman:
any news about this one?

src/modules/emotes/style.css
night: hard coded css values can work fine here for wide emotes too, we just need to account for them properly in the stylesheet

src/modules/seventv/channel-emotes.js
dclstn:
```
event: ready
data: 7tv-event-sub.v1

event: update
data: {"channel":"vasp","emote_id":"60ae65b29627f9aff4fd8bef","name":"NOOOO","action":"ADD","actor":"Vasp","emote":{"name":"NOOOO","visibility":0,"mime":"image/webp","tags":[],"width":[34,68,102,136],"height":[32,64,96,128],"urls":[["1","https://cdn.7tv.app/emote/60ae65b29627f9aff4fd8bef/1x.webp"],["2","https://cdn.7tv.app/emote/60ae65b29627f9aff4fd8bef/2x.webp"],["3","https://cdn.7tv.app/emote/60ae65b29627f9aff4fd8bef/3x.webp"],["4","https://cdn.7tv.app/emote/60ae65b29627f9aff4fd8bef/4x.webp"]],"owner":{"id":"60ae49350e35477634486602","twitch_id":"238682742","display_name":"JustRogan","login":"justrogan"}}}

event: heartbeat
data: {}
```

Flags aren't provided it seems

src/modules/emotes/style.css
dclstn: Screenshot 2023-04-01 at 02 08 40

Seems like a niche use-case, worth supporting wide overlay emotes?

NBruderman: I think so, because suddenly it can be used on a wide emote and become meta. I just can't know with twitch chat what will happen next

ScrubN: I am personally fine with this result; it is identical to how overlays are handled in another project I help develop.

Another question to ask though is how do wide overlays behave? Example: https://7tv.app/emotes/6360b863593eb316d898eb78

night (approved):

PR #6072, opened 2023-04-06
Emote Menu Button Placement

dclstn:
PR Changes:
- Migrates legacy setting to allow replacing the native emote menu button
- Reworks some default setting value logic to migrate past settings
- Future installations will have the emote menu defaultly enabled
- Adds tip to give users option to replace default emote picker button

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -204,16 +204,7 @@ export const SettingDefaultValues = {
   [SettingIds.WHISPERS]: true,
   [SettingIds.SHOW_DIRECTORY_LIVE_TAB]: false,
   [SettingIds.CHANNEL_POINTS_MESSAGE_HIGHLIGHTS]: true,
-  [SettingIds.EMOTE_MENU]: (settings) => {
-    if (settings == null) {
-      return EmoteMenuTypes.NONE;
-    }
-    const emoteMenu = settings[SettingIds.LEGACY_EMOTE_MENU];
-    if (emoteMenu != null && emoteMenu === true) {
-      return EmoteMenuTypes.LEGACY;
-    }
-    return EmoteMenuTypes.NONE;
-  },
+  [SettingIds.EMOTE_MENU]: EmoteMenuTypes.NONE,
   [SettingIds.AUTO_THEME_MODE]: false,
   [SettingIds.DARKENED_MODE]: false,
   [SettingIds.PRIME_PROMOTIONS]: true,
--- a/src/constants.js
+++ b/src/constants.js
@@ -61,7 +61,7 @@ export const DeletedMessageTypes = {
 
 export const EmoteMenuTypes = {
   NONE: 0,
-  LEGACY: 1,
+  LEGACY_ENABLED: 1,
   ENABLED: 2,
 };
 
--- a/src/constants.js
+++ b/src/constants.js
@@ -19,7 +19,8 @@ export const SettingIds = {
   WHISPERS: 'whispers',
   SHOW_DIRECTORY_LIVE_TAB: 'showDirectoryLiveTab',
   CHANNEL_POINTS_MESSAGE_HIGHLIGHTS: 'channelPointsMessageHighlights',
-  EMOTE_MENU: 'clickTwitchEmotes',
+  LEGACY_EMOTE_MENU: 'clickTwitchEmotes', // legacy
+  EMOTE_MENU: 'emoteMenu',
   AUTO_THEME_MODE: 'autoThemeMode',
   DARKENED_MODE: 'darkenedMode',
   PRIME_PROMOTIONS: 'primePromotions',
@@ -56,6 +57,12 @@ export const DeletedMessageTypes = {
   SHOW: 1,
   HIDE: 2,
   HIGHLIGHT: 3,
+};
+
+export const EmoteMenuTypes = {
+  NONE: 0,
+  DEFAULT: 1,
+  REPLACE_NATIVE: 2,
 };
 
 export const SidebarFlags = {
@@ -196,7 +203,16 @@ export const SettingDefaultValues = {
   [SettingIds.WHISPERS]: true,
   [SettingIds.SHOW_DIRECTORY_LIVE_TAB]: false,
   [SettingIds.CHANNEL_POINTS_MESSAGE_HIGHLIGHTS]: true,
-  [SettingIds.EMOTE_MENU]: false,
+  [SettingIds.EMOTE_MENU]: (settings) => {
+    if (settings == null) {
+      return EmoteMenuTypes.DEFAULT;
+    }
+    const emoteMenu = settings[SettingIds.LEGACY_EMOTE_MENU];
+    if (emoteMenu != null && emoteMenu === false) {
+      return EmoteMenuTypes.NONE;
+    }
+    return EmoteMenuTypes.DEFAULT;
+  },
   [SettingIds.AUTO_THEME_MODE]: false,
   [SettingIds.DARKENED_MODE]: false,
   [SettingIds.PRIME_PROMOTIONS]: true,
--- a/src/constants.js
+++ b/src/constants.js
@@ -61,8 +61,8 @@ export const DeletedMessageTypes = {
 
 export const EmoteMenuTypes = {
   NONE: 0,
-  DEFAULT: 1,
-  REPLACE_NATIVE: 2,
+  LEGACY: 1,
+  ENABLED: 2,
 };
 
 export const SidebarFlags = {
@@ -206,13 +206,13 @@ export const SettingDefaultValues = {
   [SettingIds.CHANNEL_POINTS_MESSAGE_HIGHLIGHTS]: true,
   [SettingIds.EMOTE_MENU]: (settings) => {
     if (settings == null) {
-      return EmoteMenuTypes.DEFAULT;
-    }
-    const emoteMenu = settings[SettingIds.LEGACY_EMOTE_MENU];
-    if (emoteMenu != null && emoteMenu === false) {
       return EmoteMenuTypes.NONE;
     }
-    return EmoteMenuTypes.DEFAULT;
+    const emoteMenu = settings[SettingIds.LEGACY_EMOTE_MENU];
+    if (emoteMenu != null && emoteMenu === true) {
+      return EmoteMenuTypes.LEGACY;
+    }
+    return EmoteMenuTypes.NONE;
   },
   [SettingIds.AUTO_THEME_MODE]: false,
   [SettingIds.DARKENED_MODE]: false,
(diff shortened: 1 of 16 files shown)


src/modules/settings/components/settings/twitch/EmoteMenu.jsx
```diff
@@ -0,0 +1,61 @@
...
+                <p className={styles.settingDescription}>
+                  {formatMessage({defaultMessage: 'Emote picker button appears below chat input box'})}
+                </p>
+              </div>
+            </Radio>
+            <Radio key="none" value={EmoteMenuTypes.NONE}>
+              <div>
+                <p className={styles.heading}>{formatMessage({defaultMessage: 'None'})}</p>
+                <p className={styles.settingDescription}>
+                  {formatMessage({defaultMessage: 'Hide the emote picker button entirely'})}
```
night, line 45: as a user, this copy is harder to understand than the prior. It would be great if we can reduce that cognitive load by keeping the primary on/off toggle and having a secondary toggle which is disabled unless the emote menu is enabled, which explicitly controls the placement

night (changes requested):

src/modules/settings/components/settings/twitch/EmoteMenu.jsx
```diff
@@ -13,42 +11,26 @@ const SETTING_NAME = formatMessage({defaultMessage: 'Emote Menu'});
...
-              </div>
-            </Radio>
-          </RadioGroup>
-        </FormGroup>
+        <Toggle checked={toggled} onChange={(state) => setValue(state ? EmoteMenuTypes.LEGACY : EmoteMenuTypes.NONE)} />
       </div>
+      {toggled ? (
+        <div className={styles.settingRow}>
+          <p className={styles.settingDescription}>
+            {formatMessage({defaultMessage: "Replace twitch's native emote menu."})}
```
night, line 26: nit: capitalize T

src/modules/emote_menu/components/Tip.jsx
```diff
@@ -40,6 +51,24 @@ function getTipToDisplay() {
     ];
   }
 
+  if (!tips[EmoteMenuTips.EMOTE_MENU_REPLACE_DEFAULT] && emoteMenuValue !== EmoteMenuTypes.ENABLED) {
+    return [
+      EmoteMenuTips.EMOTE_MENU_REPLACE_DEFAULT,
+      formatMessage(
+        {defaultMessage: '<button>Replace</button> the default emote picker button'},
```
night, line 58: is this confusing UX? clicking on the link causes a potential destructive behavior. we may want to present the setting to a user at most.

src/settings.js
```diff
@@ -22,6 +22,13 @@ class Settings extends SafeEventEmitter {
     const oldSettings = storage.get(SETTINGS_STORAGE_KEY);
     settings = {...defaultSettings, ...oldSettings};
 
+    if (oldSettings != null && oldSettings.version !== process.env.EXT_VER) {
```
night, line 25: this should be below the following code block. additionally, it would be great if there was a way to combine this with upgrading the flags?

src/settings.js
```diff
@@ -22,6 +22,13 @@ class Settings extends SafeEventEmitter {
...
+      const oldEmoteMenuValue = oldSettings[SettingIds.LEGACY_EMOTE_MENU];
+      const emoteMenuValue = oldEmoteMenuValue ? EmoteMenuTypes.LEGACY : EmoteMenuTypes.NONE;
+      settings = {...settings, [SettingIds.EMOTE_MENU]: emoteMenuValue};
```
night, line 28: replacing the settings object for every migration will be slower over time. maybe there's a way to do all mutations and then change the reference?

src/modules/settings/components/settings/twitch/EmoteMenu.jsx
```diff
@@ -0,0 +1,43 @@
...
+function EmoteMenu() {
+  const [value, setValue] = useStorageState(SettingIds.EMOTE_MENU);
+  const toggled = value !== EmoteMenuTypes.NONE;
+  return (
+    <Panel header={SETTING_NAME}>
+      <div className={styles.settingRow}>
+        <p className={styles.settingDescription}>
+          {formatMessage({defaultMessage: 'Enables a more advanced emote menu for chat'})}
+        </p>
+        <Toggle checked={toggled} onChange={(state) => setValue(state ? EmoteMenuTypes.LEGACY : EmoteMenuTypes.NONE)} />
```
night, line 21: if we do want to call the old type legacy, perhaps toggling this fresh on sets `ENABLED` and and toggling the following setting off is `LEGACY_ENABLED`

src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -58,29 +59,46 @@ export default class EmoteMenuModule {
...
+    const emoteMenuValue = settings.get(SettingIds.EMOTE_MENU);
 
-    if (emoteMenuEnabled && legacyContainer == null) {
-      const container = document.querySelector(CHAT_SETTINGS_BUTTON_CONTAINER_SELECTOR);
-      if (container == null) {
-        return;
-      }
-      const rightContainer = container.lastChild;
+    if (emoteMenuValue !== EmoteMenuTypes.NONE) {
       const buttonContainer = document.createElement('div');
```
night, line 65: this new logic will create a button container even if the button is already rendered. we should keep the prior optimizations in tact

src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -91,18 +109,25 @@ export default class EmoteMenuModule {
...
-      this.show(emoteMenuEnabled);
+      this.show(emoteMenuValue);
     }
   }
 
-  show(visible) {
+  show(emoteMenuValue) {
     const legacyContainer = document.querySelector(LEGACY_BTTV_EMOTE_PICKER_BUTTON_CONTAINER_SELECTOR);
-
     if (legacyContainer == null) {
```
night, line 118: this logic doesn't make sense. how is the legacy container not null when we want to to show the new button?

src/constants.js
```diff
@@ -58,6 +59,12 @@ export const DeletedMessageTypes = {
   HIGHLIGHT: 3,
 };
 
+export const EmoteMenuTypes = {
+  NONE: 0,
+  LEGACY: 1,
+  ENABLED: 2,
```
night, line 65: nit: `ENABLED_LEGACY`

night (changes requested):

src/modules/emote_menu/twitch/EmoteMenu.jsx
dclstn: Prior logic doesn't take into account the placement I believe, so even if the button is already rendered we should try replace its location on the DOM? I may be wrong however?

night: could we check for either button?

src/modules/settings/components/settings/youtube/EmoteMenu.jsx
```diff
@@ -28,5 +32,5 @@ registerComponent(EmoteMenu, {
   settingId: SettingIds.EMOTE_MENU,
   name: SETTING_NAME,
   category: CategoryTypes.CHAT,
-  keywords: ['twitch', 'emotes', 'popup'],
+  keywords: ['youtube', 'emotes', 'popup'],
```
night, line 35: did we duplicate just for a keyword? why not just remove twitch/youtube?

src/settings.js
```diff
@@ -77,7 +77,17 @@ class Settings extends SafeEventEmitter {
...
 
-  upgradeFlags(version) {
+  upgrade(version) {
+    if (version !== process.env.EXT_VER) {
+      const oldEmoteMenuValue = this.get(SettingIds.LEGACY_EMOTE_MENU);
+      if (oldEmoteMenuValue != null) {
+        const emoteMenuValue = oldEmoteMenuValue ? EmoteMenuTypes.LEGACY_ENABLED : EmoteMenuTypes.NONE;
+        // now storing emote menu as an enum rather than a boolean
+        this.set(SettingIds.EMOTE_MENU, emoteMenuValue);
+        this.set(SettingIds.LEGACY_EMOTE_MENU, undefined);
```
night, line 87: did we verify `undefined` removes from storage

src/modules/settings/components/settings/youtube/EmoteMenu.jsx
dclstn: The component differs, twitch has 3 configuration however youtube has 2.

src/modules/emote_menu/twitch/EmoteMenu.jsx
night: this doesn't appear resolved. to fix this, we could split the two button references up between `legacyButtonContainer` and `buttonContainer`, which would allow you to check existence of either when toggling between them.

src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -46,41 +52,70 @@ export default class EmoteMenuModule {
...
-      rightContainer.insertBefore(buttonContainer, rightContainer.lastChild);
-
-      if (mountedRoot != null) {
-        mountedRoot.unmount();
-        isMounted = false;
+
+      const currentContainer = document.querySelector(BTTV_BUTTON_CONTAINER_SELECTOR);
+      if (currentContainer != null) {
+        currentContainer.remove();
+        if (mountedRoot != null) {
```
night, line 106: we've been iterating on this specific function for a bit now, but it feels like we're going in circles. 

the goals of this function were to:
1. load the button into the DOM
2. unmount the prior react root + mount the new react root, if we loaded the button into the DOM

due to the complexities added by your changes, the function is trying to do too much at once, as it is now responsible for toggling the different buttons.

a cleaner approach may be to fully unmount the prior button, if needed and if there is one. and then mount the new button, if it is needed. we can split up this function to reduce the edge cases being ran into with this.

as written, there is a new bug introduced here where we are leaking root references, since if container is not found in the dom, we never cleanup the root

src/modules/emote_menu/twitch/EmoteMenu.module.css
```diff
@@ -3,16 +3,18 @@
...
-  }
-}
-
 .hideEmoteMenuButton {
   display: none;
 }
 
+.chatInputIcon {
+  display: inline-flex;
+  vertical-align: top;
```
night, line 12: we should avoid changing container styles, especially because we're injecting blindly into a parent in the chain. if the button isn't rendering properly, we should change the button itself.

src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -37,82 +35,142 @@ class SafeEmoteMenuButton extends React.Component {
...
-    legacyContainer.classList.toggle(styles.hideEmoteMenuButton, !visible);
+  if (mountedRoot != null) {
+    mountedRoot.unmount();
   }
 
-  appendToChat({code: text}, shouldFocus = true) {
-    let prefixText = twitch.getChatInputValue();
+  mountedRoot = createRoot(buttonContainer);
+  mountedRoot.render(
+    <SafeEmoteMenuButton
```
night, line 134: should the button be our logo?

night (changes requested):

night (approved):

PR #6112, opened 2023-04-29
Kick.com base, Global Emotes Support

abhishekg999:
Hi! This PR adds base kick.com support for bttv. It currently only supports BTTV Global Emotes.
image

The additions follow the same styles as the addition of youtube. It makes minimal adjustments and relies on existing emote replacement functionality. 

Adds the following:
- Adds a kick.com watcher and module
- Adds kick message emitter
- Processes message node usuing chat.messageReplacer
- Configures css for kick messages

Im unsure on whether or not BTTV has plans for kick, but if so, this would be a start.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -253,6 +253,7 @@ export const PlatformTypes = {
   TWITCH: 1,
   TWITCH_CLIPS: 2,
   YOUTUBE: 3,
+  KICK: 4,
 };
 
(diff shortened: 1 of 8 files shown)


night:
Thanks for the PR. As this feature implementation requires more changes than just global emotes to be effective we are going to close this for right now. Feel free to voice your support for this over at https://github.com/night/betterttv/discussions/6146 so we can help prioritize it.

PR #6126, opened 2023-05-02
Add right label for font settings

dclstn:
This PR aims to make it more clear when a font size / family is set, to overall prevent unintended configurations.

### DIFF
--- a/src/modules/chat_settings/index.js
+++ b/src/modules/chat_settings/index.js
@@ -1,10 +1,11 @@
 import watcher from '../../watcher.js';
 import settings from '../settings/index.js';
-import chatFontSettings from '../chat_font_settings/index.js';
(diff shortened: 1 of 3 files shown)


src/modules/chat_settings/index.js
```diff
@@ -94,6 +121,7 @@ class ChatSettingsModule {
     domObserver.on(CHAT_SETTINGS_MOD_TOOLS_SELECTOR, () => {
       this.renderSettings();
     });
+    setFontUpdateCallback(this.renderSettings);
```
night, line 124: we can rerender after the set functions complete, since they are synchronous

night (changes requested):

night (approved):

PR #6161, opened 2023-05-22
Display uptime duration aside stream preview

dclstn:
Addresses feature request #6038


dclstn:
Note: current implementation won't work due to too many requests*

dclstn:
Closing PR due to lack of feasibility and strain it will cause on twitch's infrastructure.

PR #6188, opened 2023-05-30
Emote Modifier Images

dclstn:


night (approved):
this change looks ok. can you double check that implementation of modifier emotes in the chat module?

dclstn:
> this change looks ok. can you double check that implementation of modifier emotes in the chat module?

Screenshot 2023-05-30 at 16 28 32

(wrong image is being used for rotate right)

Screenshot 2023-05-30 at 16 28 40

As far as I can tell, it works as intended. Unless there's a new test case you'd like me to try?

PR #6243, opened 2023-06-30
[Feature] Load frequently used emotes from twitch

dclstn:


src/modules/emote_menu/stores/emote-menu-store.js
```diff
@@ -33,13 +34,38 @@ function computeScore({totalUses, recentUses}) {
   return Math.floor((totalUses * frecency) / recentUses.length);
 }
 
+const TWITCH_EMOTE_USAGE_HISTORY_KEY = 'twilight.emote_picker_history';
+
+function loadDefaultFrequentEmotes() {
+  if (getPlatform() !== PlatformTypes.TWITCH || !storage._localStorageSupport) {
```
night, line 40: keys prefixed with underscore are generally considered private and should not be accessed out of the class.

src/modules/emote_menu/stores/emote-menu-store.js
```diff
@@ -33,13 +34,38 @@ function computeScore({totalUses, recentUses}) {
...
+    return DEFAULT_FREQUENT_EMOTES;
+  }
+
+  const history = storage.get(TWITCH_EMOTE_USAGE_HISTORY_KEY, null);
+  if (history == null) {
+    return DEFAULT_FREQUENT_EMOTES;
+  }
+
+  const emoteHistory = {};
+  for (const [emoteId, {uses, lastUpdatedAt}] of Object.entries(history)) {
```
night, line 50: if the format of this storage value changes, it could break the emote menu store load. should this be in a try/catch?

src/modules/emote_menu/stores/emote-menu-store.js
```diff
@@ -33,13 +34,38 @@ function computeScore({totalUses, recentUses}) {
...
+const TWITCH_EMOTE_USAGE_HISTORY_KEY = 'twilight.emote_picker_history';
+
+function loadDefaultFrequentEmotes() {
+  if (getPlatform() !== PlatformTypes.TWITCH || !storage._localStorageSupport) {
+    return DEFAULT_FREQUENT_EMOTES;
+  }
+
+  const history = storage.get(TWITCH_EMOTE_USAGE_HISTORY_KEY, null);
+  if (history == null) {
+    return DEFAULT_FREQUENT_EMOTES;
```
night, line 46: what happens if the value is empty rather than null?

src/modules/emote_menu/stores/emote-menu-store.js
```diff
@@ -33,13 +34,45 @@ function computeScore({totalUses, recentUses}) {
...
+  }
+
+  try {
+    const historyEntries = Object.entries(history);
+    if (historyEntries.length === 0) {
+      return DEFAULT_FREQUENT_EMOTES;
+    }
+    const emoteHistory = {};
+    for (const [emoteId, {uses, lastUpdatedAt}] of historyEntries) {
+      const serializedEmoteId = `${EmoteProviders.TWITCH}-${emoteId}`;
```
night, line 56: nit: it would be great if this were shared with how we already generate other emote canonical ids. if we change how canonical ids are generated, this will break.

night (approved):
lgtm otherwise

PR #6255, opened 2023-07-05
Add Setting to hide Hype Chat

dclstn:
PR uses feature that isn't yet available on firefox browsers, but is roadmapped to be added soon.

https://developer.mozilla.org/en-US/docs/Web/CSS/:has

### DIFF
--- a/src/modules/hype_chat/index.js
+++ b/src/modules/hype_chat/index.js
@@ -9,19 +9,29 @@ import styles from './styles.module.css';
 const HYPE_CHAT_WRAPPER_SELECTOR = '.paid-pinned-chat-message-content-wrapper';
 const HYPE_CHAT_MESSAGE_SELECTOR = 'span[data-a-target="chat-message-text"]';
 
+const CHAT_TEXT_AREA_SELECTOR = '.chat-input__textarea';
+const STREAM_CHAT_SELECTOR = '.stream-chat';
+
 let removeHypeChatListener = null;
 class HypeChatModule {
   constructor() {
-    this.loadHideHypeChat();
-    this.loadHypeChatMessageReplacer();
-    settings.on(`changed.${SettingIds.HYPE_CHAT}`, () => {
+    this.load();
+    settings.on(`changed.${SettingIds.HYPE_CHAT}`, () => this.load());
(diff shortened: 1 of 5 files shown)


src/modules/hype_chat/styles.module.css
```diff
@@ -0,0 +1,13 @@
+.hideHypeChatMessages {
+  div[class$='paid-pinned-chat-message-content-wrapper'] {
+    display: none !important;
+  }
+}
+
+.hideHypeChatButton {
+  button:has(
```
night, line 8: can we scope this down further so we're not filtering every button on the page? alternatively, we could leverage the dom watcher to check this within a scope. this would then work on firefox too..

src/modules/hype_chat/styles.module.css
```diff
@@ -0,0 +1,13 @@
+.hideHypeChatMessages {
+  div[class$='paid-pinned-chat-message-content-wrapper'] {
```
night, line 2: can we scope this down further or match against the fully qualified class names?

dclstn: The styles are not applied to the body anymore, `.hideHypeChatMessages` is applied to the `.streamChat` and `.hideHypeChatButton` is applied to the chat box.

src/modules/hype_chat/index.js
```diff
@@ -1,26 +1,67 @@
...
+const CHAT_TEXT_AREA_SELECTOR = '.chat-input__textarea';
+const STREAM_CHAT_SELECTOR = '.stream-chat';
+
+let removeHypeChatListener = null;
 class HypeChatModule {
   constructor() {
-    dom.on(HYPE_CHAT_WRAPPER_SELECTOR, (node, isConnected) => {
+    this.load();
+    settings.on(`changed.${SettingIds.HYPE_CHAT}`, () => this.load());
+    dom.on(CHAT_TEXT_AREA_SELECTOR, (_, isConnected) => {
```
night, line 20: let's refactor this a bit.. dom watcher doesn't need to be loaded all the time, only if the setting is disabled. it would be great to pull this out into a method like we do for the other watcher too.

night (approved):

PR #6317, opened 2023-08-08
Create chat_fullscreen_popup module

nath:
Hey, not sure if you're open to contributions, but this is something I've found myself wanting a couple times so here it is.

This creates a module to toggle showing the chatbox over top of the fullscreen stream. The use case for this is when watching in fullscreen and something happens on stream that makes you want to see chat for a few moments without having to exit fullscreen.

It's fairly clunky how it just overlays the chatbox and blocks part of the stream, but the use case is really just for where you want to check chat for a second, compared to watching in theater mode or something.

Please let me know if there's anything that needs to be adjusted, this is my first time working with this codebase.

Screenshots:
Fullscreen with no chat
Fullscreen with chat window overlaid

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -56,11 +56,6 @@ export const ChatLayoutTypes = {
   RIGHT: 2,
 };
 
-export const ChatFullscreenPopupTypes = {
-  DISABLED: 0,
-  ENABLED: 1,
-};
-
 export const DeletedMessageTypes = {
   DEFAULT: 0,
   SHOW: 1,
@@ -212,6 +207,7 @@ export const SettingDefaultValues = {
   [SettingIds.TIMEOUT_HIGHLIGHTS]: true,
   [SettingIds.HIGHLIGHT_FEEDBACK]: false,
   [SettingIds.CHAT_LAYOUT]: ChatLayoutTypes.RIGHT,
+  [SettingIds.CHAT_FULLSCREEN_POPUP]: true,
   [SettingIds.TAB_COMPLETION_TOOLTIP]: false,
   [SettingIds.TAB_COMPLETION_EMOTE_PRIORITY]: false,
   [SettingIds.WHISPERS]: true,
--- a/src/constants.js
+++ b/src/constants.js
@@ -15,6 +15,7 @@ export const SettingIds = {
   TIMEOUT_HIGHLIGHTS: 'timeoutHighlights',
   HIGHLIGHT_FEEDBACK: 'highlightFeedback',
   CHAT_LAYOUT: 'chatLayout',
+  CHAT_FULLSCREEN_POPUP: 'chatFullscreenPopup',
   TAB_COMPLETION_TOOLTIP: 'tabCompletionTooltip',
   TAB_COMPLETION_EMOTE_PRIORITY: 'tabCompletionEmotePriority',
   WHISPERS: 'whispers',
@@ -53,6 +54,11 @@ export const CategoryTypes = {
 export const ChatLayoutTypes = {
   LEFT: 1,
   RIGHT: 2,
+};
+
+export const ChatFullscreenPopupTypes = {
+  DISABLED: 0,
+  ENABLED: 1,
 };
 
 export const DeletedMessageTypes = {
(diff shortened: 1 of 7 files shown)


src/constants.js
```diff
@@ -55,6 +56,11 @@ export const ChatLayoutTypes = {
   RIGHT: 2,
 };
 
+export const ChatFullscreenPopupTypes = {
```
dclstn, line 59: Why isn't this a boolean?

src/modules/chat_fullscreen_popup/style.css
```diff
@@ -0,0 +1,13 @@
+.bttv-fullscreen-chat {
```
dclstn, line 1: Can this be converted to a CSS module?

src/constants.js
```diff
@@ -15,6 +15,7 @@ export const SettingIds = {
   TIMEOUT_HIGHLIGHTS: 'timeoutHighlights',
   HIGHLIGHT_FEEDBACK: 'highlightFeedback',
   CHAT_LAYOUT: 'chatLayout',
+  CHAT_FULLSCREEN_POPUP: 'chatFullscreenPopup',
```
dclstn, line 18: There's no default value for this, meaning it's both disabled and enabled?

src/modules/chat_fullscreen_popup/index.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  updateElements() {
+    this.chatElement = document.querySelector('.stream-chat');
+    this.originalChatParentElement = this.chatElement.parentNode;
+  }
+
+  fullscreenChanged() {
+    if (
+      document.fullscreenElement &&
```
dclstn, line 35: nit prefer: `document.fullscreenElement != null`

src/modules/chat_fullscreen_popup/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+    if (this.fullscreenChatPopupShown) {
+      this.hideChatPopup();
+    } else {
+      this.showChatPopup();
+    }
+  }
+
+  showChatPopup() {
+    if (this.chatElement.parentNode === this.originalChatParentElement) {
```
dclstn, line 84: what if `this.chatElement` doesn't exist anymore?

dclstn: what happens when the chat isn't available, i.e fullscreening the frontpage then pressing 'c'?

src/modules/chat_fullscreen_popup/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+  updateElements() {
+    this.chatElement = document.querySelector('.stream-chat');
+    this.originalChatParentElement = this.chatElement.parentNode;
+  }
+
+  fullscreenChanged() {
+    if (
+      document.fullscreenElement &&
+      settings.get(SettingIds.CHAT_FULLSCREEN_POPUP) !== ChatFullscreenPopupTypes.DISABLED
```
dclstn, line 36: nit: `settings.get(SettingIds.CHAT_FULLSCREEN_POPUP) === ChatFullscreenPopupTypes.ENABLED`

but can be considered redundant if swapped to boolean

src/modules/chat_fullscreen_popup/style.css
```diff
@@ -0,0 +1,13 @@
...
+  position: absolute !important;
+  width: 400px !important;
+  top: 0;
+  right: 0;
+}
+
+/* Respect chat_left_side */
+.bttv-swap-chat {
```
dclstn, line 9: I like this, however this logic should pertain to the left side chat module

dclstn (commented):
Collaboration is always welcome and this is a good first pull requeust, just a few nits + requests but overall I like the feature.

nath (commented):
Thanks for the feedback, I did everything except for the CSS module.

Also, how do I access the settings page when running in dev mode with the tampermonkey script? That's the only thing I haven't tried out yet. Thanks!

src/modules/chat_fullscreen_popup/index.js
```diff
@@ -79,25 +82,23 @@ class ChatFullscreenPopup {
...
     if (this.chatElement == null || this.originalChatParentElement == null) {
-      // Something has gone wrong in initialization or the page doesn't have chat
       return;
     }
 
     if (this.chatElement.parentNode === this.originalChatParentElement) {
       document.fullscreenElement.appendChild(this.chatElement);
     }
-    this.chatElement.classList.add('bttv-fullscreen-chat');
+    this.chatElement.classList.add(styles.bttvFullscreenChat);
```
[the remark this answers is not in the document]
dclstn: Thanks for the PR!

PR #6590, opened 2024-01-25
Overlay Emote Modifier

Jake2500:
Adds a `o!` modifier that makes the modified emote an "overlay".

I'm not very creative, but an example: `Waiting o! Clap `
Screenshot 2024-01-25 230905

Screenshot 2024-01-25 230614

I have no idea whether `o!` is a good choice.
Suggestions and review comments welcome.


night:
Thanks for the PR. To prevent abuse of this, you'll need to add a restriction preventing multiple emotes from stacking on top of a single emote. I think a sane limit is only 1 emote can stack on top of the prior emote.

Jake2500:
> I think a sane limit is only 1 emote can stack on top of the prior emote.
> 
> Additionally, what happens if there isn't a prior emote to stack on top of? Doesn't that move the emote out of view?

This is a good point. I've covered both these cases by requiring there to be a prior emote and that it does **not** have the overlay modifier.

| message | before | after |
|-----------|---------|------|
| `peepoHappy o! Clap` |  |  |
| `o! Clap` |  |  |
| `peepoHappy foo o! Clap` |  |  |
| `peepoHappy o! Clap o! Clap2` |  |  |

src/modules/chat/index.js
```diff
@@ -341,11 +344,35 @@ class ChatModule {
...
+            predecessor != null &&
+            partMetadata[predecessor]?.emote &&
+            !isOverlay[predecessor] &&
+            modifiers.some(overlayModifierPredicate)
+          ) {
+            // Keep track of this part being an overlay.
+            isOverlay[j] = true;
+          } else {
+            // Strip the overlay modifier from the part.
+            modifiers = modifiers.filter(negate(overlayModifierPredicate));
```
night, line 373: stripping it works, but it also results in the modifier going missing from the output. for other modifiers, we do not show the modified emote but we show the modifier was sent when used incorrectly. is there a way to replicate this behavior here?

image

night: as an aside, do you have an image asset for this modifier? if not, we can probably have one made

night (changes requested):

src/modules/chat/index.js
```diff
@@ -341,11 +344,35 @@ class ChatModule {
...
             ) {
+              predecessor = k;
               break;
             }
             modifiers.push(partMetadataItem);
             parts[k] = null;
           }
+
+          const overlayModifierPredicate = ({modifier}) => modifier === 'o!';
+          const negate = (pred) => (x) => !pred(x);
```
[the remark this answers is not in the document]
Jake2500: I've moved these out and near the definitions of the modifiers.

src/modules/chat/index.js
Jake2500: > is there a way to replicate this behavior here?

I've modified it so that if an overlay modifier is removed, it replaces the corresponding parts in the part array.
Screenshot 2024-02-17 094835

> do you have an image asset for this modifier?

I'm no artist, so feel free to use or not use this.
overlay2

night: @dclstn any ideas on the icon for this?

dclstn: Something like so perhaps:

DallasHoff: When the image is small in emote form, it kind of just looks like 1 square. Maybe make it so that the top square doesn't overlap the bottom one quite so completely?

night: agreed with Dallas' feedback. you can't tell what this is when smaller. it also resembles a copy button, which is not be the intention.

dclstn: How's this?


downsized for reference:

night:
It doesn't look like overlay emotes seem to work when you post them. but they do appear when others post theme. might be related to how twitch is rendering messages you post.

src/modules/chat/index.js
night: this looks better

Jake2500:
> It doesn't look like overlay emotes seem to work when you post them. but they do appear when others post them. might be related to how twitch is rendering messages you post.

Screenshot 2024-03-16 173012
That's a screenshot of a message posted by me in twitch chat.

What are you seeing and how can I reproduce it?

night:
> What are you seeing and how can I reproduce it?

Make sure you disable other browser extensions, as perhaps another extension is affecting things. But essentially, posting the overlay modifier results in 

image

It seems when other users post it, or you reload, it renders properly:

image

My guess here is Twitch is rendering what is posted differently when you post.

Jake2500:
That's very odd. I can't reproduce this in Chrome 122.0.6261.113,

PR #6591, opened 2024-01-27
Party Emote Modifier

DallasHoff:
Implements a modifier `p!` which adds a "party" effect to emotes. The modifier gives the emote a rotating color change effect using a CSS `filter` that increases saturation and animates `hue-rotate()`. It also uses a bit of `sepia()` which "averages" the base emote's colors a little while also making the effect work for emotes that have a lot of white in them.

I also made the global emote for `p!` if that is helpful:
bttv-party

### DIFF
--- a/src/modules/chat/index.js
+++ b/src/modules/chat/index.js
@@ -30,6 +30,7 @@ const EMOTE_MODIFIERS = {
   'c!': 'bttv-emote-modifier-cursed',
   'l!': 'bttv-emote-modifier-rotate-left',
   'r!': 'bttv-emote-modifier-rotate-right',
+  'p!': 'bttv-emote-modifier-party',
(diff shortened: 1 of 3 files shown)


dclstn:
This is awesome, how is this performance wise? What happens when the entire chat is filled with these modified emotes?

DallasHoff:
Animating the `filter` property is something that the browser will GPU-accelerate, so the animation will be optimized and moved off the main thread. It will not block or be blocked by any work happening on the main thread. 

I tested it a bit with a chat full of emotes using the effect. I confirmed that the browser composites the animation, and profiling with the effect enabled versus disabled showed no significant performance difference.

night (approved):

night:
Thanks for your PR!

PR #6593, opened 2024-01-28
Add Clip Command

dclstn:


src/modules/chat_commands/commands/clip.js
```diff
@@ -0,0 +1,63 @@
...
+          broadcasterID: channel.id,
+          offsetSeconds: startOffset,
+        },
+      });
+
+      if (title != null && title.length > 0) {
+        await twitch.graphqlMutation(renameClipMutation, {input: {slug: data.createClip.clip.slug, title}});
+      }
+
+      twitch.sendChatMessage(data.createClip.clip.url);
```
night, line 57: should this send to chat, or send to you?

if we want it to send to chat, should it have the title prefixed?

night (changes requested):

src/modules/chat_commands/commands/clip.js
dclstn: Won't the title be visible through the embed anyway?

night (approved):

PR #6621, opened 2024-02-06
Case sensitive doubleclick mention

Tester798:
The nickname of the mentioned user can be pasted with the correct case if possible.


night (approved):

night:
Thanks for the PR!

PR #6657, opened 2024-02-27
move list to modal

dclstn:
Screenshot 2024-02-27 at 18 52 54

### DIFF
--- a/src/modules/settings/components/settings/global/Emotes.jsx
+++ b/src/modules/settings/components/settings/global/Emotes.jsx
@@ -4,12 +4,15 @@ import Checkbox from 'rsuite/Checkbox';
 import CheckboxGroup from 'rsuite/CheckboxGroup';
 import Modal from 'rsuite/Modal';
 import Panel from 'rsuite/Panel';
+import Tag from 'rsuite/Tag';
 import useStorageState from '../../../../../common/hooks/StorageState.jsx';
 import {CategoryTypes, SettingIds, EmoteTypeFlags} from '../../../../../constants.js';
 import formatMessage from '../../../../../i18n/index.js';
 import {hasFlag, setFlag} from '../../../../../utils/flags.js';
 import styles from '../../../styles/header.module.css';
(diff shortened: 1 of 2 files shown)


src/modules/settings/components/settings/global/Emotes.jsx
```diff
@@ -35,9 +38,54 @@ function SafetyWarningModal({open, onClose}) {
...
+  return (
+    <Modal backdrop="static" open={open} onClose={() => onClose()} size="xs">
+      <Modal.Header>
+        <strong>{formatMessage({defaultMessage: 'Emote Modifiers'})}</strong>
+      </Modal.Header>
+      <Modal.Body>
+        <p className={emotesStyles.modifiersModalDescription}>
+          {formatMessage({
+            defaultMessage:
+              'Emote modifiers allow you to transform emotes in realtime. To use them, simply add the modifier to the start of the emote code. For example, w! Kappa will display the emote in a wide format.',
```
DallasHoff, line 62: The example should probably use a global BTTV emote, not Kappa, since the modifier would not actually work on Kappa. Might I recommend `w! monkaS`?

I might also rephrase "add the modifier to the start of the emote code" to "add the modifier directly before the emote".

src/modules/settings/components/settings/global/Emotes.jsx
```diff
@@ -35,9 +38,54 @@ function SafetyWarningModal({open, onClose}) {
   );
 }
 
+const EMOTE_MODIFIERS_DESCRIPTION = {
+  'w!': formatMessage({defaultMessage: 'Will display the emote in a wide format'}),
+  'h!': formatMessage({defaultMessage: 'Will flip the emote horizontally'}),
+  'v!': formatMessage({defaultMessage: 'Will flip the emote vertically'}),
+  'z!': formatMessage({defaultMessage: 'Will display zero-width between prior emote'}),
```
DallasHoff, line 45: I'd probably describe z! as "Will remove the gap after the prior emote".

Issue #6658, opened 2024-02-28
Chat rerender causes native emote menu to show


dclstn:
### Describe the bug

Whenever twitch chat rerenders, the native emote menu button will reshow

### Steps to reproduce

1. Have emote menu enabled
2. Fullscreen then exit, or any other form of rerendering chat

### Expected behavior

Native emote menu button should remain hidden

### Screenshots

<img width="139" alt="Screenshot 2024-02-28 at 12 09 17" src="https://github.com/night/betterttv/assets/43322006/8bba9144-0a79-4660-ad69-8bc057eb2771">

### Device information

-

PR #6728, opened 2024-04-08
Prevent native emote menu from rerendering

dclstn:
Fix for #6658

### DIFF
--- a/src/modules/emote_menu/twitch/EmoteMenu.jsx
+++ b/src/modules/emote_menu/twitch/EmoteMenu.jsx
@@ -100,7 +100,7 @@ function unloadButton(container, nativeEmotePickerButton) {
     nativeEmotePickerButton = document.querySelector(EMOTE_PICKER_BUTTON_SELECTOR);
   }
   if (nativeEmotePickerButton != null) {
-    nativeEmotePickerButton.classList.remove(styles.hideEmoteMenuButton);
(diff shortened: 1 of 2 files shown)


src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -100,7 +100,7 @@ function unloadButton(container, nativeEmotePickerButton) {
     nativeEmotePickerButton = document.querySelector(EMOTE_PICKER_BUTTON_SELECTOR);
   }
   if (nativeEmotePickerButton != null) {
-    nativeEmotePickerButton.classList.remove(styles.hideEmoteMenuButton);
+    nativeEmotePickerButton.parentElement.classList.remove(styles.hideEmoteMenuButtonContainer);
```
night, line 103: we should make some sort of assertion that the parentElement has specific properties to ensure we don't accidentally hide a container with multiple other things.

dclstn: The css queries for the button, so hopefully this won't happen.

dclstn: Maybe it would be wiser to apply the style to the chat input button container?

night: i dont know if there's an easy way to target that container. we could target the text area by class name, and then this emote button by target name.

night (approved):

PR #6766, opened 2024-04-23
Fix search for twitch command store

dclstn:


UniversalGeneve1:
`/join` is still broken

dclstn:
> `/join` is still broken

Expect this change to be available in the next patch

PR #6859, opened 2024-06-05
[Feature] Enlarge Emotes on Hover

jamalalhneidi:
Hello,
I believe this is a nice feature to have and I had some free time so I implemented it. This is what it looks like in action:

Screenshot 2024-06-05 231943

This is my first PR, so kindly bear with me if I have done something in an unconventional way.
Thanks.


src/modules/emotes/emote.js
```diff
@@ -79,11 +79,23 @@ export default class Emote {
     }
     container.appendChild(image);
 
-    const tooltip = document.createElement('div');
-    tooltip.classList.add('bttv-tooltip', 'bttv-tooltip--up', 'bttv-tooltip--align-center');
-    tooltip.textContent = `${this.code}\n${
+    const enlargedImage = new Image();
+    enlargedImage.classList.add('bttv-emote-image-enlarged');
+    enlargedImage.src = createSrc(this.images, shouldRenderStatic, '4x');
+    enlargedImage.alt = `${prefixModifiers != null && prefixModifiers.length > 0 ? `${prefixModifiers.join(' ')} ` : ''}${
```
night, line 85: i'd remove the alt text since it isn't really going to add value to a screen reader when they won't be able to open this tooltip

src/modules/emotes/style.css
```diff
@@ -98,6 +98,10 @@ img.emoji.yt-live-chat-text-message-renderer + .bttv-global-emo-5e76d399d6581c37
   margin: -0.5rem 0;
 }
 
+.bttv-emote-image-enlarged {
+  max-height: none !important;
```
night, line 102: 1. why is this set to important?
2. setting none as max height may actually cause undesirable behavior in some conditions. there's really no guarantee that image providers provide reasonably sized images, meaning we should always maintain an upper bound.
3. we may want to add some bottom margin to make it look a bit more organized
4. squares images may look funky, so it makes sense to radius them a tad.

```
max-height: 112px;
width: auto;
max-width: 336px;
border-radius: 0.2rem;
overflow: hidden;
```

night (changes requested):

night (approved):

jamalalhneidi:
@night Thank you for your time and work. I'll make sure to keep the things you pointed out in mind for future PRs.

PR #6961, opened 2024-07-23
[Fix] Replaced hardcoded 7tv cdn

FrantaBOT:
Hello,
I've replaced the hardcoded 7TV emote url with the url from the api because it's going to cause problems in the future.

### DIFF
--- a/src/modules/seventv/channel-emotes.js
+++ b/src/modules/seventv/channel-emotes.js
@@ -55,13 +55,13 @@ class SevenTVChannelEmotes extends AbstractEmotes {
         for (const {
           id,
           name: code,
-          data: {listed, animated, owner, flags},
+          data: {listed, animated, owner, flags, host: {url}},
         } of emotes) {
           if (!listed && !hasFlag(settings.get(SettingIds.EMOTES), EmoteTypeFlags.SEVENTV_UNLISTED_EMOTES)) {
             continue;
           }
 
-          this.emotes.set(code, createEmote(id, code, animated, owner, category, isOverlay(flags)));
+          this.emotes.set(code, createEmote(id, code, url, animated, owner, category, isOverlay(flags)));
         }
 
         eventSource = new ReconnectingEventSource(
--- a/src/modules/seventv/channel-emotes.js
+++ b/src/modules/seventv/channel-emotes.js
@@ -92,14 +92,14 @@ class SevenTVChannelEmotes extends AbstractEmotes {
       const {
         id,
         name: code,
-        data: {listed, animated, owner, flags},
+        data: {listed, animated, owner, flags, host: {url}},
       } = value;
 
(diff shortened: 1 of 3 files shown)


night:
appreciate the PR, but is there any link to a discussion or explanation of upstream changes?

src/modules/seventv/channel-emotes.js
```diff
@@ -55,13 +55,13 @@ class SevenTVChannelEmotes extends AbstractEmotes {
...
           name: code,
-          data: {listed, animated, owner, flags},
+          data: {listed, animated, owner, flags, host: {url}},
         } of emotes) {
           if (!listed && !hasFlag(settings.get(SettingIds.EMOTES), EmoteTypeFlags.SEVENTV_UNLISTED_EMOTES)) {
             continue;
           }
 
-          this.emotes.set(code, createEmote(id, code, animated, owner, category, isOverlay(flags)));
+          this.emotes.set(code, createEmote(id, code, url, animated, owner, category, isOverlay(flags)));
```
night, line 64: there are multiple places where `createEmote` is called that were missed

src/modules/seventv/utils.js
```diff
@@ -1,11 +1,11 @@
...
 import Emote from '../emotes/emote.js';
 
-function emoteUrl(id, version, static_ = false) {
-  return `https://cdn.7tv.app/emote/${encodeURIComponent(id)}/${version}${static_ ? '_static' : ''}.webp`;
+function emoteUrl(url, version, static_ = false) {
+  return `https:${url}/${version}${static_ ? '_static' : ''}.webp`;
 }
 
-export function createEmote(id, code, animated, owner, category, overlay) {
+export function createEmote(id, code, url, animated, owner, category, overlay) {
```
night, line 8: rather than adding an arg to the middle of the list, we prefer to add new args onto the end of the list

src/modules/seventv/utils.js
```diff
@@ -1,11 +1,11 @@
 import {hasFlag} from '../../utils/flags.js';
 import Emote from '../emotes/emote.js';
 
-function emoteUrl(id, version, static_ = false) {
-  return `https://cdn.7tv.app/emote/${encodeURIComponent(id)}/${version}${static_ ? '_static' : ''}.webp`;
+function emoteUrl(url, version, static_ = false) {
+  return `https:${url}/${version}${static_ ? '_static' : ''}.webp`;
```
night, line 5: is there any guarantee they won't add a scheme in front of the url? trailing slash?

night (changes requested):

src/modules/seventv/utils.js
FrantaBOT: No, there isn't. It makes sense to remove the scheme

dclstn:
Lgtm, just a few formatting issues and bug fix that are now fixed

Issue #6998, opened 2024-08-15
Following + Chat on the left when going to stories and going back one page


WillLBegin:
### Describe the bug

On firefox, if I open the new stories feature and then click back on the browser, the chat moves to the left of the page with the option chat on right enabled. It seems like it forgets the option is enabled. Refreshing the page fixes it.

![image](https://github.com/user-attachments/assets/a1bc5985-35d1-4e79-a0af-eac5f8743dc7)

### Steps to reproduce

1. Go to stories
2. Go back 
![image](https://github.com/user-attachments/assets/db2e432f-f0ba-469c-a3a6-8cd075d7cc81)
3. Chat is now left.

### Expected behavior

Chat does not move.

### Screenshots

![image](https://github.com/user-attachments/assets/a1bc5985-35d1-4e79-a0af-eac5f8743dc7)

### Device information

PC, Firefox latest version

### BetterTTV Version

7.6.5

### Version Check

- [X] I have checked that I'm using the newest version of BetterTTV available.

night:
fixed via 2cbffc2dce24972a9486993c6e4b20e41bcf1734

PR #7010, opened 2024-08-27
Fix left-side chat on stories close

dclstn:
Removed comment as it's rendered null due to dom observer

### DIFF
--- a/src/modules/chat_left_side/index.js
+++ b/src/modules/chat_left_side/index.js
@@ -1,25 +1,15 @@
 import {ChatLayoutTypes, PlatformTypes, SettingIds} from '../../constants.js';
-import dom from '../../observers/dom.js';
 import settings from '../../settings.js';
 import {loadModuleForPlatforms} from '../../utils/modules.js';
-
-const SIDE_NAV_SELECTOR = '.side-nav';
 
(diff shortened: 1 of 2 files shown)


dclstn:
Fix for #6998

src/modules/chat_left_side/index.js
```diff
@@ -1,18 +1,21 @@
...
   constructor() {
     settings.on(`changed.${SettingIds.CHAT_LAYOUT}`, () => this.toggleLeftSideChat());
+    dom.on(SIDE_NAV_SELECTOR, () => this.toggleLeftSideChat());
     this.toggleLeftSideChat();
   }
 
   toggleLeftSideChat() {
     document.body.classList.toggle('bttv-swap-chat', settings.get(SettingIds.CHAT_LAYOUT) === ChatLayoutTypes.LEFT);
 
-    // TODO: replace this eventually when :has exists in CSS
```
night, line 14: https://caniuse.com/css-has is it time?

dclstn: `:has` will immediately grab the highest element in the tree that meets this condition. so the root is returned

What we want is `div > .side-nav` I believe

PR #7012, opened 2024-08-27
Add toggle to hide stories

dclstn:
The setting for sidebar being flags is problematic as ideally this setting would be default on, as in "Show Stories" by default


dclstn:
Addresses #6997

src/modules/hide_sidebar_elements/index.js
```diff
@@ -80,6 +82,25 @@ class HideSidebarElementsModule {
...
   }
 
+  loadHideOpenStories() {
+    const setting = settings.get(SettingIds.SIDEBAR);
+    const enabled = hasFlag(setting, SidebarFlags.OPEN_STORIES);
+    const [element] = document.getElementsByClassName(styles.hideOpenStories);
+
+    if (enabled && element == null) {
+      const sidenav = document.querySelector('.side-bar-contents');
+      const node = twitch.getOpenStoriesDOMNode(sidenav);
```
night, line 92: do we need to have this be under a dom watcher like the other sidebar elements?

dclstn: You're right, we do

night (approved):

PR #7076, opened 2024-10-05
Smaller icons on chat lines

1898182:
In mod view, sometimes we might want to copy/paste a message but are annoyed by the icons getting in the way.

I wanted to contribute and saw this discussion post: https://github.com/night/betterttv/discussions/6962 which convinced me to go ahead.

Maybe there's a better file in which I can add the styling. And perhaps should this be a feature under Chat settings or a default behavior?

## Screenshots below are before / after

Before

After

This last screenshot is when we preview the messages outside of mod view.

viewer


night:
Thanks for the PR. However, because the click targets appear too small to be usable, we'd recommend turning these icons off rather than making them smaller. You can already turn these Mod Icons off within the Twitch chat settings. You can still perform moderation actions like these from the moderator card, which can be accessed by clicking on the user's name in chat.

PR #7096, opened 2024-10-11
Fixes #7095 Auto Leave Raids

### LINKED ISSUE #7095: Auto join raids
### Describe the bug

I have auto join raids turned off and for some reason it has still been sending me recently

### Steps to reproduce

go to settings>channel> auto join raids option off 

### Expected behavior

I expect not to be sent to another channel when the streamer raids someone else

### Screenshots

image

### Device information

PC Edge 

### BetterTTV Version

7.6.7

### Additional information

_No response_

### Version Check

- [X] I have checked that I'm using the newest version of BetterTTV available.

Dendraspis:
According to the code and this source:
image

...the condition should be changed because there are no attributes anymore.
Then the feature should work again...

### DIFF
--- a/src/modules/auto_join_raids/index.js
+++ b/src/modules/auto_join_raids/index.js
@@ -27,10 +27,7 @@ class AutoJoinRaidsModule {
 
   leaveRaid() {
     const leaveButton = document.querySelector(RAID_LEAVE_BUTTON_SELECTOR);
-    if (
-      leaveButton == null ||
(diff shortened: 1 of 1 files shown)


night:
did you test this? if there are no attributes anymore, then this code change would not do anything.

the logic is an OR, meaning that if there is no leave button OR we see a button that controls the raid for the broadcaster, we early return.

night:
this has been properly fixed via e710064d704275106c8fa9bf22eaafafeff8f549

PR #7168, opened 2024-11-12
Prevent freq used order change during shift hold

Invertex:
Addresses https://github.com/night/betterttv/discussions/6932

When wanting to click a "Frequently Used" emote multiple times while holding shift, the order will often change between each click, leading to misclicks. Instead I've cached the clicks while Shift is held and then process them after it's no longer held, resulting in order changes only happening after the user lets go of Shift or clicks an emote while not holding shift.

(cache needs to be outside EmoteMenu otherwise we end up losing it during other processing)

### DIFF
--- a/src/modules/emote_menu/components/EmoteMenu.jsx
+++ b/src/modules/emote_menu/components/EmoteMenu.jsx
@@ -16,6 +16,18 @@ function setKeyPressCallback(newKeyPressCallback) {
   keyPressCallback = newKeyPressCallback;
 }
 
+const cachedEmoteClicks = new Map();
+
+function cacheOrProcessEmoteClicks(emote, shiftPressed) {
+  if (emote && !cachedEmoteClicks.has(emote.canonicalId)) {
+    cachedEmoteClicks.set(emote.canonicalId, emote);
+  }
+  if (!shiftPressed && cachedEmoteClicks.size > 0) {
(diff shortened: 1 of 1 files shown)


src/modules/emote_menu/components/EmoteMenu.jsx
```diff
@@ -16,6 +16,18 @@ function setKeyPressCallback(newKeyPressCallback) {
   keyPressCallback = newKeyPressCallback;
 }
 
+const cachedEmoteClicks = new Map();
```
night, line 19: nit: `pendingEmoteHistory`

src/modules/emote_menu/components/EmoteMenu.jsx
```diff
@@ -56,6 +68,9 @@ export default function EmoteMenu({toggleWhisper, appendToChat, onSetTip}) {
     function buttonPressCallback(event) {
       setAltPressed(event.altKey);
       setShiftPressed(event.shiftKey);
+      if (!event.shiftKey) {
+        cacheOrProcessEmoteClicks(null, false);
```
night, line 72: if the popout is closed via other means (clicking out of the popout or closing it via the X in the upper right), these track events are lost. this operation may need to be in the parent. since `appendToChat` is a parent function, perhaps the track could be moved there instead.

night:
closing this due to inactivity. we can re-open this in the future if the review comments are addressed

Issue #7394, opened 2025-04-02
Twitch is Showing Live Unfollowed Channels in the Sidebar


mcavery333:
### Describe the bug

I have the Recommended Channels and Similar Channels options unchecked but Live Channels are still showing in my sidebar.

![Image](https://github.com/user-attachments/assets/158c8ec8-e1ad-4dc9-bf93-2ff6a557d023)

![Image](https://github.com/user-attachments/assets/4c56ff35-19b1-423a-a8c7-d19c19cf8128)

### Steps to reproduce

1. Login to Twitch
2. See Live Channels in the Sidebar

### Expected behavior

I expect the Live Channels to be hidden since I have the Recommended Channels and Similar Channels options unchecked.

### Screenshots

![Image](https://github.com/user-attachments/assets/970b957a-f9fa-4f8f-9a71-14d505bfa0bc)

![Image](https://github.com/user-attachments/assets/00efdd64-deef-43c0-9ab3-5996c49881ac)

### Device information

PC, Windows 10

### BetterTTV Version

7.6.11

### Version Check

- [x] I have checked that I'm using the newest version of BetterTTV available.

Ibrax4:
Its annoying man and I'm here for this reason Twitch  and I think its one of Twitch sh!t
Their goal is forcing you to watch those sh!t streams.
thinhking to leave that sh!t Twitch

App/browser suck

Lilykoi:
Please fix, Twitch is already cluttered and I like that I only saw people who I follow are live.

Ibrax4:
And when you watch a stream youre gonna see Viewers Also Watch

al11588:
Same issue. Glad I am not the only person with the issue.

M4DMAXX:
Got the same yes, and if you have a stream open, it also shows "Viewers also watch" Too much Clutter for me....
Hope they can fix this :)

M4DMAXX:
Where do we put this in?

M4DMAXX:
> This should hide the recommended channels:
> 
> ```
> twitch.tv##.side-nav-section:nth-of-type(2)
> twitch.tv##.side-nav-section:nth-of-type(3)
> ```

Are you sure about the filter, and which elements they block?
I tried both, without the filter I see "Followed Channels, "Live Channel" and "Viewers also Watch"
With the filter on, I still see "Live Channel" and "Viewers also Watch" it just disabled for me the "Followed Channel", which is the Important for me.

[...]

PR #7397, opened 2025-04-04
Fix: Twitch is Showing Live Unfollowed Channels in the Sidebar

evaccari:
The goal of this PR is to fix the following issue: https://github.com/night/betterttv/issues/7394

The section "Recently watched channels" seems to no longer exist: I did not retrieve the mapping value, and decided to remove the option.


night:
fixed via e7c9803d0ae5218ae4b767c1c97505d9fd3f587a, which is preferred to removing code potentially still in use via experimentation

Issue #7431, opened 2025-04-27
7TV emote preview max height is capped to 28 pixels


ptkr0:
### Describe the bug

While other emotes show bigger preview image on hover (max height 112px) 7TV emotes are mostly capped at 28px which makes previewing them impossible.

### Steps to reproduce

Send BetterTTV, FrankerFaceZ and 7TV emote. hover mouse on each one and notice how 7TV will show smaller preview than other

### Expected behavior

7TV emote preview should be as big as BetterTTV and FrankerFaceZ

### Screenshots

![Image](https://github.com/user-attachments/assets/e7238d5e-1f59-4142-a4a9-b6dc363c5577)

![Image](https://github.com/user-attachments/assets/3cf46ca3-e312-483c-9d80-1ca8991f65fe)

![Image](https://github.com/user-attachments/assets/408c1915-be3e-48e9-9bb5-a5d60159d248)

### Device information

Tested on:
Firefox 137.0.2 (64 bits) 
Microsoft Edge Version 135.0.3179.98

### BetterTTV Version

7.6.13

### Version Check

- [x] I have checked that I'm using the newest version of BetterTTV available.

PR #7432, opened 2025-04-27
Increase max-height of seventv-channel emote images to 112px

ptkr0:
Fixes issue #7431 by changing the max-height property for CSS rule

(before and after)
obraz

obraz

There probably was or maybe still is a reason for this rule, but I'm not sure what it actually is.
_Deleting whole rule would also fix the problem._


night:
this rule existed because there were some large images being served, and we do not generally restrict the height of images in chat. in this case the height should still be max 28px but we can add an override for these emotes displayed within the `bttv-tooltip-emote-image` class.

ptkr0:
As you suggested adding !important flag to the max-height property inside `bttv-tooltip-emote-image` also fixes the issue, while leaving max height set inside `.bttv-emote.seventv-channel img` class to 28px.

night (approved):

Issue #7460, opened 2025-05-07
Drops not auto claiming


HiddenKnife526:
### Describe the bug

Despite the Drops Auto Claim option being enabled, earned drops are not being auto claimed as of today for me. A minor issue, but one that should be fixed eventually.

### Steps to reproduce

1. Earn a drop from an active drop campaign
2. Check the Inventory page

### Expected behavior

With that option enabled, the drops should be auto claimed as they were before.

### Screenshots

![Image](https://github.com/user-attachments/assets/85de5a5b-5ba2-4433-8f88-793ada619911)

![Image](https://github.com/user-attachments/assets/f93ade4b-f5aa-453d-8c0a-327d7519d19c)

### Device information

- Laptop
- Windows 10 (OS Build 19045.5737)
- Firefox 138.0.1 (64-bit)

### BetterTTV Version

7.6.13

### Version Check

- [x] I have checked that I'm using the newest version of BetterTTV available.

SigmaTheDJ:
I'm having the same issue. I'm also using Firefox 138.0.1 on Windows 10, and my BetterTTV version is 7.6.13.

It was definitely working on April 22nd as that was the start of a new Overwatch season and the drops auto-claimed then. I didn't try and get any drops again until today, and now the auto-claim is not working. I've tried removing and reinstalling the app, and my settings are definitely correct.

DuploMinh:
I'm having the same issue on FF Dev Edition 139.0b6 on Windows 11, BetterTTV 7.6.13.

Angry-Maid:
Same issue, win11, Firefox 139.0, BTTV 7.6.13

Deejay-Monsta:
> ### Describe the bug
> Despite the Drops Auto Claim option being enabled, earned drops are not being auto claimed as of today for me. A minor issue, but one that should be fixed eventually.
> 
> ### Steps to reproduce
> 1. Earn a drop from an active drop campaign
> 2. Check the Inventory page
> 
> ### Expected behavior
> With that option enabled, the drops should be auto claimed as they were before.
> 
> ### Screenshots
> ![Image](https://github.com/user-attachments/assets/85de5a5b-5ba2-4433-8f88-793ada619911)
> 
> ![Image](https://github.com/user-attachments/assets/f93ade4b-f5aa-453d-8c0a-327d7519d19c)
> 
> ### Device information
> * Laptop
> * Windows 10 (OS Build 19045.5737)
> * Firefox 138.0.1 (64-bit)
[...]

PR #7469, opened 2025-05-12
Fix Auto-Claim

dclstn:
Adresses #7460

### DIFF
--- a/src/modules/auto_claim/index.js
+++ b/src/modules/auto_claim/index.js
@@ -38,7 +38,7 @@ const claimDropMutation = gql`
   }
 `;
 
-async function handleMessage(message) {
+async function handleOnsiteNotificationMessage(message) {
   const messageData = JSON.parse(message);
 
   if (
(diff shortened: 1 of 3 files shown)


src/modules/auto_claim/index.js
```diff
@@ -81,19 +81,30 @@ class AutoClaimModule {
...
     const shouldAutoClaim = hasFlag(autoClaim, AutoClaimFlags.DROPS);
-    const client = window.__twitch_pubsub_client;
-    const currentUser = getCurrentUser();
-    if (client == null || currentUser == null) {
-      return;
-    }
-    const topicId = `onsite-notifications.${currentUser.id}`;
-    if (!shouldAutoClaim && listenerAttached) {
-      listenerAttached = false;
-      client.topicListeners.removeListener(topicId, handleMessage);
```
night, line 92: the old system was built around listening to the already-established topic listener.

it looks like Twitch is moving away from pubsub, thus why this is no longer being listened to here. we likely should move to the new hermes system instead.

Issue #7479, opened 2025-05-20
Video Size broken


maxxxwinder:
### Describe the bug

Hello, the video player size on twitch was about half size today when I logged on. I checked for updates in both Firefox and BTTV. Once I disabled BTTV the video size went back to normal after a refresh of page. 

### Steps to reproduce

Go to Twitch with BTTV running, 
watch stream and notice the video about half the normal size .
disable bttv and refresh the  video goes to the full size of the twitch browser video window. 

### Expected behavior

A video feed that fills the normal window on twitch

### Screenshots

![Image](https://github.com/user-attachments/assets/0bb23033-fd7c-4da4-825a-6735f1fbd385)

### Device information

Win 10
Firefox 138.0.4
twitch 

### BetterTTV Version

7.6.13

### Version Check

- [x] I have checked that I'm using the newest version of BetterTTV available.

maxxxwinder:
After looking at the bug above Status: Open. #7478 it seems that Left Side chat is to blame for the reduced video size. If I move the option back to right side chat the video size fixes itself

neverim2:
I confirm the problem, and I checked it not only in Firefox, but also in Chrome, Vivaldi and Opera. The problem occurs when moving the chat to the left.

dclstn:
Closing in favour of #7478

Issue #7478, opened 2025-05-19
Left-sided chat breaking


TheYxel:
### Describe the bug

Today I encountered an error on the Twitch website. At first I didn't know what was causing it. Then I accidentally disabled addons. And after a while I discovered it was BTTV. Then I tried changing the chat position to the default (right) and after reloading the video it was fixed.
I tried testing BTTV with other browsers (Opera, Firefox), the result is the same.

### Steps to reproduce

1. Turn on chat on the left side of screen
2. Watch any stream

### Expected behavior

Video plays in a small window

### Screenshots

![Image](https://github.com/user-attachments/assets/8dec2a76-457c-4dab-9965-27db7089c44e)

### Device information

- PC
- Windows 10
- Google Chrome
- 136.0.7103.114 

### BetterTTV Version

7.6.13

### Version Check

- [x] I have checked that I'm using the newest version of BetterTTV available.

starslyr:
Even when I switch back to chat on right I have bars on the left and right side which I don't have when I use incognito mode (no plugins) and I didn't have them on streams earlier in the day either.

Ewig01:
Same issue with chat on left side causing the video to be smaller. Windows 11, Brave Browser, BTTV 7.6.13.

divector8:
I'm having the same issue as well with chat set to left side. Switching it to right side makes the bug go away. I'm on Windows 11, Firefox 138.0.4, BTTV 7.6.13

DEGO0527:
Same issue with chat on left side causing the video to be smaller. Windows 11, EDGE Browser 136.0.3240.76 (x64), BTTV 7.6.13.

MoDog0:
Same issue. Almost feels like the stream is playing at native resolution instead of stretched. 
Left side chat enabled appears to be causing it. 

BTTV Version 7.6.13, 
Doesn't sound like it is browser dependent from reading the other comments but 
Chrome Version 136.0.7103.114 (Official Build) (64-bit)

jt415gz:
same issue as above comments

Benjier:
Same problem. Windows 11. Issue on Firefox and Chrome

MisterWoody:
If you know css there's a workaround. 
1. right click and inspect the video area
2. find the enclosing dive with the InjectLayout-sc-1i43xsx-0 class and select it
3. look for the css that refers to @media screen and (min-width:920px) which should have an associated width: calc(100% - 34rem) and disable it (vidfeo should then fill the area

[...]

PR #7480, opened 2025-05-20
Fix left side chat

### LINKED ISSUE #7477: left side chat option bug stream
### Describe the bug

left side chat option bug stream 

### Steps to reproduce

turn chat to left side

### Expected behavior

i think that not have that black screen

### Screenshots

Image

### Device information

w11
opera
lastest version both

### BetterTTV Version

7.6.13

### Additional information

_No response_

### Version Check

xgerhard:
fixes #7477 #7478 #7479

This uses the selector of  `.persistent-player` used in the twitch css.

### DIFF
--- a/src/modules/chat_left_side/style.css
+++ b/src/modules/chat_left_side/style.css
@@ -27,7 +27,7 @@
     display: none !important;
   }
 
-  .persistent-player:has(+.channel-root) {
+  div:has(> .channel-root) > .persistent-player {
     width: 100%;
   }
 
--- a/src/modules/chat_left_side/style.css
+++ b/src/modules/chat_left_side/style.css
@@ -27,6 +27,10 @@
     display: none !important;
   }
 
+  .persistent-player:has(+.channel-root) {
+    width: 100%;
+  }
+
   .channel-root__upper-watch--with-chat {
     width: 100% !important;
   }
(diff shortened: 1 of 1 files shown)


xgerhard:
There are a few other selectors in the stylesheet that I’m unsure about, they might no longer be relevant, but I didn’t want to remove them without confirmation.

One I couldn’t find in use:
```css
  .channel-root__upper-watch--with-chat {
    width: 100% !important;
  }
```

Also, this selector might have matched a previous DOM structure, but based on the current order:
```
<div class="persistent-player"></div>
<div class="channel-root--watch-chat"></div>
``` 
the following selector no longer works:
```css
  .channel-root--watch-chat + .persistent-player {
    width: 100%;
  }
```

dclstn:
>There are a few other selectors in the stylesheet that I’m unsure about, they might no longer be relevant, but I didn’t want to remove them without confirmation.

You should be safe to remove this, only usage I can find is from us:

Screenshot 2025-05-20 at 10 57 41

xgerhard:
@dclstn Thanks for checking, other one as well right, with the current order of divs?

dclstn:
Yeah, ` .channel-root--watch-chat` doesn't look used either

xgerhard:
That's smart, updated!

neverim2:
Ok, and how do I update the extension in Firefox and Chrome?

night (approved):

PR #7596, opened 2025-08-01
Fix pinned chat messages

RedSteelMoss:
Update container position to sticky instead of absolute.
Will position relative to nearest scrolling ancestor.

fixes night#7571


night:
fixed via 7644eecd08f41c66d7dec061ee32b391dab669af instead, since fixed doesn't handle `top`

PR #7893, opened 2026-04-17
Fix temp values being improperly synced

dclstn:


src/settings.js
```diff
@@ -23,6 +23,20 @@ class TempValue {
...
+
+  for (const [key, value] of Object.entries(currentSettings)) {
+    if (value instanceof TempValue) {
+      result[key] = value.storedValue;
+    } else {
+      result[key] = value;
+    }
+  }
+
+  return {...result, version: currentSettings.version};
```
night: can you explain the need for `version`? isn't this covered already by line 33?

night (approved):
lgtm otherwise

PR #7898, opened 2026-04-19
Fix theme variable selector

dclstn:


src/modules/shadow_dom/index.js
```diff
@@ -58,9 +70,9 @@ class ShadowDOM {
 
     const {variables, dark, light} = mantineVariablesResolver(theme, primaryColor);
 
-    const baseCssVariables = variablesToCSS('.bttv-mantine-scope', variables);
-    const darkCssVariables = variablesToCSS('.bttv-mantine-scope:has([data-mantine-color-scheme="dark"])', dark);
-    const lightCssVariables = variablesToCSS('.bttv-mantine-scope:has([data-mantine-color-scheme="light"])', light);
+    const baseCssVariables = variablesToCSS(`.${SCOPE_CLASS}`, variables);
+    const darkCssVariables = variablesToCSS(`.${DARK_MODE_CLASS}`, dark);
+    const lightCssVariables = variablesToCSS(`.${LIGHT_MODE_CLASS}`, light);
```
night: since there are only two themes, we could throw these in the base css and leverage only a `theme dark` class in the mount node.

an added benefit is things look less broken if the dark theme scoping breaks

night (approved):
lgtm otherwise

PR #7908, opened 2026-04-23
Fix emote menu button rendering

dclstn:


src/utils/svg.js
```diff
@@ -0,0 +1,22 @@
+export function createInlineSvg(svgRaw, options = {}) {
+  const {className, size, decorative = true} = options;
+
+  const container = document.createElement('span');
+  if (className) {
+    container.className = className;
+  }
+  container.innerHTML = svgRaw;
+
+  const svg = container.querySelector('svg');
```
night: can we leverage jsx instead?

```
function Svg() {
  return <svg><path .../>...</svg>;
}
```

night:
i think the root cast of mask image breaking was a CORS error. should be fixed now?

PR #7910, opened 2026-04-24
Add emote aliases for tab completion and chat

deboralbarros:
## Summary

- Users can create personal aliases (e.g. `lol` → `KEKW`) that map any typed shortcut to any emote code
- Aliases take priority over regular emotes in tab completion (pressing Tab)
- Aliases are also substituted on message send (pressing Enter), so typing `lol` directly and hitting Enter sends `KEKW`
- Alias management UI added under **Settings → Tab Completion → Emote Aliases**, following the same table pattern as Highlight/Blacklist Keywords
- All new strings translated across all 17 supported locales

## How it works

**Storage:** aliases are persisted in `localStorage` as `{ uuid: { id, alias, targetCode } }` under `SettingIds.EMOTE_ALIASES`, following the same pattern as highlight keywords.

**Tab completion** (`chat_tab_completion`): `getAliasMatches()` checks user aliases before the regular emote list, so matching aliases always surface first.

**Message send** (`send_message`): `replaceEmoteAliases()` is registered in the `methodList` pipeline and substitutes each word in the message against the alias map before dispatch — the same mechanism used by the emoji substitution module.

## Test plan

- [ ] Create an alias (`lol` → `KEKW`) in Settings → Tab Completion → Emote Aliases
- [ ] Type `lol` in chat and press Tab — verify `KEKW` appears as the first suggestion
- [ ] Type `lol` in chat and press Enter directly — verify the message is sent as `KEKW`
- [ ] Verify aliases with no matching emote still insert the target code string
- [ ] Verify deleting an alias removes it from both tab completion and send behavior
- [ ] Verify the Settings page back button returns to the main Settings page

## Preview
Demo of the feature in action.

https://github.com/user-attachments/assets/089b5958-fe67-4bf7-ad5a-7554f848c4d1


night (commented):
Thanks for putting this together. Really clean implementation, and following the existing `SettingKeywords` pattern made it easy to read.

A few things before we can merge:

1. **Please rebase onto `master`.** We just landed a Mantine 9 + React 19 upgrade (#7947), so this branch is now behind. The change should still merge cleanly given how closely it tracks the existing settings patterns.

2. **Drop the non-English locale changes.** Only the English strings (the `defaultMessage` values in source) need to be edited. Every other locale is managed via Crowdin and gets imported automatically, so hand-edited translations there will get overwritten on the next sync. Removing the `cs-CZ.json` through `uk-UA.json` edits will shrink the diff substantially.

3. **Naming/framing.** What you've built isn't really an emote alias system, it's a general text replacement feature (the "target" can be any string, with no validation that it's an emote, and the substitution runs unconditionally on send). That's actually fine, and probably more useful than a strict emote-only version, but the naming ("Emote Aliases", `EMOTE_ALIASES`, `targetCode`) overpromises. Could you rename around "Text Replacements" or "Substitutions" or similar? `SettingIds.TEXT_REPLACEMENTS`, page name, settings copy, etc. Happy to bikeshed the exact term if you have a preference.

src/modules/send_message/index.js
```diff
@@ -1,15 +1,33 @@
...
+  for (const {alias, targetCode} of Object.values(aliases)) {
+    if (alias && targetCode) aliasMap[alias.toLowerCase()] = targetCode;
+  }
+
+  if (Object.keys(aliasMap).length === 0) return;
+
+  sendState.message = sendState.message
+    .split(' ')
+    .map((word) => aliasMap[word.toLowerCase()] ?? word)
+    .join(' ');
```
night: Splitting on a single space and rejoining collapses runs of whitespace, so `"hi  there"` becomes `"hi there"`. A regex-based replace would preserve the original spacing.

src/modules/settings/components/SettingTextReplacements.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+  return (
+    <TableTr>
+      <TableTd className={styles.inputDataCell}>
+        <TextInput
+          variant="unstyled"
+          classNames={{input: styles.cellInput, root: styles.inputRoot, wrapper: styles.inputWrapper}}
+          ref={triggerInputRef}
+          defaultValue={data.alias}
+          onBlur={({target: {value}}) => onUpdate({alias: value.trim()})}
```
night: Aliases containing whitespace are silently unreachable. `applyTextReplacements` matches with `/\S+/g` (single non-whitespace tokens), so a trigger like `good morning` is stored and shown in the table but will never fire on send, and tab completion has the same blind spot. Consider stripping internal whitespace on blur, or validating and showing an inline error.

src/modules/chat_tab_completion/index.js
```diff
@@ -147,12 +147,29 @@ class ChatTabcompletionModule {
     }
   };
 
+  getTextReplacementMatches(prefix) {
+    const replacements = settings.get(SettingIds.TEXT_REPLACEMENTS);
+    if (replacements == null) return [];
+
+    const seen = new Set();
+    const matches = [];
+    for (const {alias, replacement} of Object.values(replacements)) {
```
night, line 156: The `seen` set only dedupes replacements against each other, not against the emote/user lists built below. If a user adds `lol -> KEKW` and `KEKW` is already a known emote, `KEKW` shows up twice in tab suggestions. Easiest fix is to filter the emote and user lists against the alias output before concatenating (or pass the alias set into the emote loop).

src/modules/settings/components/SettingsModal.jsx
```diff
@@ -230,7 +234,9 @@ function SettingsModal({setHandleOpen}) {
             <Title order={1}>
               {page === PageTypes.HIGHLIGHT_KEYWORDS
```
night: Now that three page types share this branch, the nested ternary for the title is getting hard to scan, and adding a fourth sub-page will make it worse. A small lookup map keyed by `page` (e.g. `PAGE_TITLES[page]`) or a switch would read more cleanly.

src/modules/send_message/index.js
```diff
@@ -30,6 +45,7 @@ class SendState {
 
 let twitchSendMessage;
 const methodList = [
+  (msgObj) => applyTextReplacements(msgObj),
```
night, line 49: Worth a one-line comment noting that `applyTextReplacements` must stay first in the pipeline so triggers are substituted before emoji shortcuts run and before `chatTabCompletion.onSendMessage` records the final message into history. Otherwise the ordering looks arbitrary and is easy to accidentally shuffle.

src/modules/chat_tab_completion/index.js
```diff
@@ -147,12 +147,29 @@ class ChatTabcompletionModule {
...
+    return matches;
+  }
+
   getSuggestions(prefix, includeUsers = true, includeEmotes = true) {
     let userList = [];
     let emoteList = [];
 
     prefix = prefix.toLowerCase();
 
+    const aliasList = includeEmotes ? this.getTextReplacementMatches(prefix) : [];
```
night, line 171: it looks like we'd want this on for includeUsers or includeEmotes, right?

src/modules/settings/components/SettingTextReplacements.module.css
```diff
@@ -0,0 +1,72 @@
+.panelContent {
+  padding: 0;
+  overflow: hidden;
+}
+
+.table {
```
night, line 5: these are duplicate styles from the other keywords table. is there a way to reuse those?

src/modules/settings/settings/twitch/TabCompletion.jsx
```diff
@@ -1,18 +1,32 @@
...
 const SETTING_NAME = formatMessage({defaultMessage: 'Tab Completion'});
 
 function TabCompletion(props, ref) {
+  const {setPage} = useContext(PageContext);
   const [value, setValue] = useStorageState(SettingIds.TAB_COMPLETION_EMOTE_PRIORITY);
 
   return (
     <SettingGroup ref={ref} {...props} name={SETTING_NAME}>
+      <SettingWrapper
+        name={formatMessage({defaultMessage: 'Text Replacements'})}
```
night: these replacements seem to work outside of tab completion. does this placement make sense here? is the intention for them to only work within tab completion?

PR #7974, opened 2026-05-13
Fix emote menu in left side chat

dclstn:


src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -172,6 +175,7 @@ function loadEmoteMenu(onMount, onError) {
       boundingQuerySelector={CHAT_TEXT_AREA}
       emoteMenuToggleButtonSelector={EMOTE_MENU_TOGGLE_BUTTON_SELECTOR}
       offsetOptions={{mainAxis: 8}}
+      placement={placement}
```
night, line 183: what about popout/embed

night (approved):

PR #7981, opened 2026-05-18
Fix cloud backup guards

dclstn:


src/modules/cloud_backup/index.js
```diff
@@ -126,17 +126,17 @@ class CloudBackup extends SafeEventEmitter {
     try {
       const currentSettings = settings.getSettings();
 
-      if (currentSettings.version !== serverSettings.version) {
```
dclstn, line 129: this guard was blocking syncing between older versions of betterttv settings

PR #7996, opened 2026-05-22
Fix legacy emote menu button placement

dclstn:


src/modules/emote_menu/twitch/EmoteMenu.jsx
```diff
@@ -86,10 +87,9 @@ function loadLegacyButton() {
     chatInput.classList.add(styles.hideShieldModeButton);
   }
 
-  const rightContainer = container.lastChild;
```
dclstn, line 89: This is what was causing the bug, twitch moved the "Chat" button to this container

night (approved):

PR #7998, opened 2026-05-25
Add search to blacklist/highlight keywords pages

dclstn:


src/modules/settings/components/SettingKeywords.jsx
```diff
@@ -27,9 +27,17 @@ import Icon from '../../../common/components/Icon.jsx';
...
+  data,
+  updateHandler,
+  deleteHandler,
+  colorColumn,
+  keywordInputRefCallback,
+  currentChannel,
+  ...props
+}) {
   const portalRef = usePortalRef();
-  const currentChannel = useCurrentChannel();
```
dclstn, line 32: pulling this up a level, as we're creating a new channel listeners for every row otherwise

src/modules/settings/components/SettingKeywords.jsx
```diff
@@ -136,13 +144,32 @@ function createNewEntry(defaultKeyword = '') {
...
+    default:
+      return 'message';
+  }
+}
+
+function keywordMatchesSearch(row, query) {
+  const keyword = row.keyword?.toLowerCase() ?? '';
+  const channels = (row.channels ?? []).join(' ').toLowerCase();
+  const target = getTargetLabel(row.type);
+  return keyword.includes(query) || channels.includes(query) || target.includes(query);
```
night: these all have equal weight. consider either matching only keywords or ranking results based on match type (keyword > target > channels)

night (approved):

PR #8018, opened 2026-05-31
fix: resolve potential null reference in AnimatedCanvas effect

MatrixNeoKozak:
The useEffect in AnimatedCanvas.jsx does not check that `ref.current` is still valid when `requestAnimationFrame` fires the callback. If the component unmounts after the interval sets up the next frame but before the frame callback runs, `ref.current` will be null, causing a null reference error when accessing `ref.current.getContext('2d')`. This is a real runtime bug that can cause console errors. The fix adds a guard in `drawAnimatedImage` to exit early if `ref.current` is null. This is a low-risk, high-value change.

### DIFF
--- a/src/common/components/AnimatedCanvas.jsx
+++ b/src/common/components/AnimatedCanvas.jsx
@@ -62,8 +62,17 @@ export default function AnimatedCanvas({width = CANVAS_WIDTH, height = CANVAS_HE
       timeRef.current = t + 0.01;
     }
 
-    const requestDrawAnimatedImage = () => window.requestAnimationFrame(drawAnimatedImage);
+    let animationFrameId = null;
+    const requestDrawAnimatedImage = () => {
+      animationFrameId = window.requestAnimationFrame(drawAnimatedImage);
+    };
     const interval = setInterval(() => requestDrawAnimatedImage(), FRAME_INTERVAL);
+    return () => {
+      clearInterval(interval);
+      if (animationFrameId != null) {
+        cancelAnimationFrame(animationFrameId);
+      }
+    };
     return () => clearInterval(interval);
   }, [documentState, isVisible, ref]);
 


src/common/components/AnimatedCanvas.jsx
```diff
@@ -62,8 +62,17 @@ export default function AnimatedCanvas({width = CANVAS_WIDTH, height = CANVAS_HE
...
+      animationFrameId = window.requestAnimationFrame(drawAnimatedImage);
+    };
     const interval = setInterval(() => requestDrawAnimatedImage(), FRAME_INTERVAL);
+    return () => {
+      clearInterval(interval);
+      if (animationFrameId != null) {
+        cancelAnimationFrame(animationFrameId);
+      }
+    };
     return () => clearInterval(interval);
```
dclstn, line 76: What's the purpose of this, if we return on line 70?

src/common/components/AnimatedCanvas.jsx
```diff
@@ -62,8 +62,17 @@ export default function AnimatedCanvas({width = CANVAS_WIDTH, height = CANVAS_HE
...
-    const requestDrawAnimatedImage = () => window.requestAnimationFrame(drawAnimatedImage);
+    let animationFrameId = null;
+    const requestDrawAnimatedImage = () => {
+      animationFrameId = window.requestAnimationFrame(drawAnimatedImage);
+    };
     const interval = setInterval(() => requestDrawAnimatedImage(), FRAME_INTERVAL);
+    return () => {
+      clearInterval(interval);
+      if (animationFrameId != null) {
+        cancelAnimationFrame(animationFrameId);
```
dclstn, line 73: Can we do window.cancelAnimationFrame, to be consistent with window.requestAnimationFrame

Also this can be just:

```
clearInterval(interval);
window.cancelAnimationFrame(animationFrameId);
```

The only time animationFrameId will be null is when initialized, but even then window.cancelAnimationFrame does not throw or reject null as an argument

src/common/components/AnimatedCanvas.jsx
```diff
@@ -62,8 +62,17 @@ export default function AnimatedCanvas({width = CANVAS_WIDTH, height = CANVAS_HE
       timeRef.current = t + 0.01;
     }
 
-    const requestDrawAnimatedImage = () => window.requestAnimationFrame(drawAnimatedImage);
+    let animationFrameId = null;
+    const requestDrawAnimatedImage = () => {
+      animationFrameId = window.requestAnimationFrame(drawAnimatedImage);
+    };
     const interval = setInterval(() => requestDrawAnimatedImage(), FRAME_INTERVAL);
```
dclstn, line 69: Let's inline this new anon function,

```
const interval = setInterval(() => {
  animationFrameId = window.requestAnimationFrame(drawAnimatedImage);
}, FRAME_INTERVAL);
```

dclstn (changes requested):
Good catch, happy to merge once these are fixed.

dclstn:
Closing in favour of 97142498004f46fd763a49e24047bd533580718d

Issue #5939, opened 2023-02-11
Threads and Pins are overlapping emote information


EdVraz:
**Describe the bug**
if there is an emote in a pin or thread and you hover over it the information is not readable

**To Reproduce**
Have a pinned message with an emote and hover over it.

**Expected behavior**

**Screenshots**
![grafik](https://user-images.githubusercontent.com/88881326/218277862-44aeb6da-9d79-43af-a650-b311ce632a81.png)
![grafik](https://user-images.githubusercontent.com/88881326/218278178-1774fcea-9a1e-4c08-b7a0-b6acb8e0b230.png)

**Desktop (please complete the following information):**

- OS: [e.g. iOS] windows
- Browser [e.g. chrome, safari] Firefox
- Version [e.g. 22] latest at time of this post

night:
This is a pretty minor issue resulting from Twitch preventing overflow on its chat box elements. This used to not be the case in the past, but changed after Twitch moved its tooltips to use portals.  It will likely be eventually addressed by changing how we render tooltips.

dclstn:
Fixed by https://github.com/night/betterttv/pull/8031

PR #8031, opened 2026-06-02
Improve tooltip rendering

dclstn:
before:
Screenshot 2026-06-04 at 11 13 51
after:
Screenshot 2026-06-04 at 11 14 29

Today tooltips are not reactive, meaning they're easily hidden when clipping the edge of the screen. Or won't auto-repoisition when too close to the top of the screen.

I propose we render them in the shadow dom, as a react component with floating ui to handle these edge cases.


dclstn:
Addresses https://github.com/night/betterttv/issues/5939

With being rendered in the shadow dom, tooltips should always appear ontop.

src/modules/tooltip/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+export function bindTooltip(element, {elementId, content, className = null}) {
+  const id = elementId ?? crypto.randomUUID();
+
+  if (!tooltipMap.has(id)) {
+    tooltipMap.set(id, {content, className});
+  }
+
+  if (!element.hasAttribute(TOOLTIP_TARGET_ATTRIBUTE)) {
+    element.setAttribute(TOOLTIP_TARGET_ATTRIBUTE, id);
```
night: rather than keep a memory map up to date, have we considered storing the tooltip as accessory data on the dom node? something like `node.__$bttvTooltip`?

this way we don't need to manage the state of tooltips in memory, the renderer will pull the desired render from the dom node.

night (approved):

PR #8046, opened 2026-06-09
feat: select first command argument on autocomplete completion

dclstn:
## What

When completing a chatbot command that takes arguments via the command autocomplete, the first argument placeholder is now inserted **and selected**, so the next keystroke replaces it.

- `!shoutout` → completes to `!shoutout [user]` with `[user]` selected → typing immediately overwrites it.
- Commands with **no** arguments are unchanged: `!command ` with the caret at the end.

Only the first argument is inserted (commands can have several — `[user]`/`[word]`/`[phrase]`); the placeholder text is reused from `ArgumentDisplayTextByArgumentType` so it stays consistent with the autocomplete row subtitle. Twitch only for now.

## How

`twitch.setChatInputValue(text, shouldFocus, selection?)` gains an optional `selection` (`{start, end}` character offsets into `text`). When a range is requested, it's applied through Slate rather than the controlled value setters.

This matters because Twitch's chat input is a controlled Slate editor whose value prop reaches `editor.children` **asynchronously**. Setting the value and then selecting `[start, end)` immediately lands the range on stale (usually empty) content — an invalid selection that corrupts the next keystrokes (e.g. inserting newlines). A raw DOM selection doesn't work either: the controlled re-render collapses it back to a caret.

So for the range case we drive everything through Slate synchronously — `select-all → insertText(text) → select(range)` — which populates the content before selecting and keeps Twitch's value in sync via `onChange`. Collapsed-caret callers (emote menu, tab completion, mod cards, no-arg commands) are unchanged.

## Testing

Verified live on `twitch.tv/popout/<channel>/chat`: after completion the input shows `!shoutout [user]` as a single block with a valid selection over `[user]`, and a keystroke replaces it cleanly (`!shoutout x`) with no extra blocks/newlines.

### DIFF
--- a/src/utils/twitch.js
+++ b/src/utils/twitch.js
@@ -112,6 +112,18 @@ export const SelectionTypes = {
   END: 3,
 };
 
+// Sets the chat input to `text` and selects its [start, end) character range.
+// The text is inserted through Slate rather than the controlled value setters
+// (which apply asynchronously) so the range lands on the real content instead of
+// stale/empty content, which would otherwise corrupt the next keystrokes. Slate's
+// onChange keeps Twitch's value in sync; the value is one text node (path [0, 0]).
+function setSlateEditorValueWithSelection(editor, text, start, end) {
+  editor.select({anchor: editor.start([]), focus: editor.end([])});
+  editor.insertText(text);
+  editor.select({anchor: {path: [0, 0], offset: start}, focus: {path: [0, 0], offset: end}});
+  editor.onChange?.();
+}
+
 export default {
   async getUserProfilePicture(userId = null) {
     const currentUser = getCurrentUser();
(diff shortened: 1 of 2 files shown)


src/utils/twitch.js
```diff
@@ -745,24 +756,33 @@ export default {
...
+      chatInput.memoizedProps.value = text;
+      chatInput.memoizedProps.setInputValue(text);
+      chatInput.memoizedProps.onValueUpdate(text);
+    }
+
+    if (!shouldFocus) {
+      return;
+    }
+
+    // TODO: remove setSelectionRange branch after legacy slate is gone
```
dclstn, line 776: Am I safe to remove this legacy selection range?

night (changes requested):
maybe instead of this tricky input logic, we take a simpler approach that will support multiple arguments.. we could instead have the UI for autocomplete remain open while the user is interfacing with the command. if the user presses esc, clicks out of the dialog, or presses enter, we can close the autocomplete ui

PR #8055, opened 2026-06-15
qol: settings modal side navigation

dclstn:
Builds on the Hugeicons Pro migration branch to refresh the settings modal UX.

## Highlights
- **Bigger modal** with a redesigned side navigation: text-labelled, per-setting icons, scrollable list with fixed logo and user-settings sections, and a scroll-spy that highlights the entry you're viewing.
- **Per-page sticky headers** with a bottom border, the close button moved into the header, and descendant pages (keyword editors) using a Mantine `Breadcrumbs` header.
- **Theme-aware surfaces** — sidebar/page backgrounds swap by color scheme, with new `--mantine-color-body-raised` / `-raised-border` tokens for the elevated user-settings card (instead of inline `color-mix`).
- **User-settings button** shows the account avatar (or Mantine placeholder when signed out) and a `Pro` badge when applicable, with proper label truncation.
- Settings list scroll-spy uses the project's `lodash` throttling convention.

### DIFF
--- a/src/common/hooks/ScrollSpy.jsx
+++ b/src/common/hooks/ScrollSpy.jsx
@@ -1,0 +1,61 @@
+import {useEffect, useState} from 'react';
+
+// Scroll spy scoped to a scroll container, so it works inside a shadow root where a document-level
+// query (e.g. Mantine's useScrollSpy) can't see any elements. Options are captured once on mount:
+// spied elements are discovered with `scrollHost.querySelectorAll(selector)`, and the hook returns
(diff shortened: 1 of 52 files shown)


night (changes requested):
overall i like the direction here, but maybe we can iterate a bit further by placing items into groupings. this will at least shorten how many "pages" exist.

one other area of improvement would be presenting options from the dashboard here too:
- manage emotes
- bots
- editors
- channel points reward
- connections
- sub badge

PR #8076, opened 2026-06-19
feat: link regex usage guide in keyword settings

dclstn:
## Problem

Highlight and Blacklist keywords already support regular expressions via the `~/pattern/flags` syntax (validated by `safe-regex2`), but this is completely undiscoverable in the UI. It's a recurring question in discussions ("Are regexes allowed in highlighting?").

## Change

Adds a small dimmed tip below the keyword table in `SettingKeywords.jsx` — the shared component rendered by **both** the Highlight Keywords and Blacklist Keywords pages — so a single change surfaces it in both places:

> Tip: Match with a regular expression by wrapping the pattern in `~/ /`. [Learn more](https://gist.github.com/dclstn/33207501beb97cad8a3aaf845218cb43)

The "Learn more" link points to a usage guide gist covering the syntax, flags, supported target types (Message / Username / Badge), and the safe-regex behavior.

## Notes

- No behavior change — regex matching already worked; this only documents it in the UI.
- Strings are wrapped in `formatMessage` for i18n.
- The guide is hosted as a gist for now; happy to move it into the repo/wiki if preferred.

## Requested in

- [#8030 — Are regexes allowed in highlighting?](https://github.com/night/betterttv/discussions/8030) (Q&A) — user asks whether regex works and suggests surfacing it in the UI to avoid confusing less-technical users. This PR closes that discoverability gap.
- [#267 — Add Regular Expressions to Highlight/Blacklist](https://github.com/night/betterttv/issues/267) (Issue, closed) — the original regex request referenced by #8030; the syntax already shipped, this documents it.

### DIFF
--- a/src/modules/settings/components/SettingKeywords.jsx
+++ b/src/modules/settings/components/SettingKeywords.jsx
@@ -2,6 +2,7 @@ import {
   ActionIcon,
   Avatar,
   Button,
+  Kbd,
   NativeSelect,
   Pill,
   Table,
@@ -15,7 +16,7 @@ import {
   TextInput,
 } from '@mantine/core';
 import React, {useCallback, useMemo, useRef, useState} from 'react';
-import {faTrash} from '@fortawesome/free-solid-svg-icons';
+import {faCircleInfo, faTrash} from '@fortawesome/free-solid-svg-icons';
 import styles from './SettingKeywords.module.css';
 import {KeywordTypes} from '@/utils/keywords';
 import formatMessage from '@/i18n/index';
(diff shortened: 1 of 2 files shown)


dclstn:
Adding some context on motivation: regex support for keywords is one of the most frequently asked-about features. It comes up repeatedly in the Discussions/Q&A (e.g. "Are regexes allowed in highlighting?") and in the Discord, yet the capability has existed for a while with no in-UI hint. This is purely a discoverability fix — surfacing an existing feature where people actually look for it, which should cut down on the recurring questions.

src/modules/settings/components/SettingKeywords.jsx
```diff
@@ -27,6 +29,16 @@ import Icon from '@/common/components/Icon';
 import usePortalRef from '@/common/hooks/PortalRef';
 import useCurrentChannel from '@/common/hooks/CurrentChannel';
 
+const REGEX_GUIDE_URL = 'https://gist.github.com/dclstn/33207501beb97cad8a3aaf845218cb43';
```
night: linking to a gist is a bit rough. is there a way to make this better?

src/modules/settings/components/SettingKeywords.jsx
```diff
@@ -320,6 +332,18 @@ function SettingKeywords({value, setValue, colorColumn = null}) {
           {formatMessage({defaultMessage: 'No keywords found, start by adding a new entry.'})}
         </Text>
       )}
+      <Text className={styles.regexTip} c="dimmed" size="md">
+        {formatMessage(
+          {
+            defaultMessage:
+              'Tip: Match with a regular expression by wrapping the pattern in {syntax}. <link>Learn more</link>',
```
night: normal users do not know what a regular expression is. it may make more sense to have this behind a ? that describes what input values are possible when clicked

dclstn:
image

src/modules/settings/components/SettingKeywords.jsx
```diff
@@ -295,9 +357,19 @@ function SettingKeywords({value, setValue, colorColumn = null}) {
...
+          <Button size="lg" className={styles.newEntryButton} onClick={newEntryHandler}>
+            {formatMessage({defaultMessage: 'New Entry'})}
+          </Button>
+          <ActionIcon
+            size="lg"
+            variant="transparent"
+            color="gray"
+            aria-label={formatMessage({defaultMessage: 'Advanced keywords'})}
+            onClick={openRegexGuideModal}>
+            <Icon icon={faCircleInfo} />
```
night: maybe we add this here: 

Image

dclstn: image

done.

PR #8078, opened 2026-06-19
feat: add setting to disable on-screen celebration effects

dclstn:
## Summary

Adds a **Celebrations** toggle to the Chat settings panel that lets users disable on-screen celebration and cheer effects — confetti animations, channel "Celebrations", and Power-up screen effects. This is a recurring request (e.g. the "disable cheer confetti effect on chat" discussion).

The setting is **enabled by default**, so existing behavior is unchanged until the user opts out.

## Implementation

Mirrors the existing `hide_bits` module pattern:

- New `ChatFlags.CELEBRATIONS` flag in `constants.js`, added to the `CHAT` default value (on by default).
- New `hide_celebrations` module that toggles a `bttv-hide-celebrations` body class based on the flag.
- `style.css` hides the relevant celebration/Power-up overlay and effect containers.
- New `Celebrations` checkbox + search keywords (`celebrations`, `confetti`, `power-ups`) on the Chat settings page.

## Notes

The CSS targets known Twitch celebration/Power-up overlay selectors. Since these effects only render transiently (when triggered), they could not be exercised live in this change; selectors use resilient attribute matching consistent with the existing `hide_bits` styles. Happy to refine selectors if maintainers know the current canonical ones.

## Requested in

- [#6628 — disable cheer confetti effect on chat](https://github.com/night/betterttv/discussions/6628) (💡 Ideas) — the request this implements ("way too obnoxious, and redundant on multiple scopes").
- [#8019 — "Twitch has a native Reactions toggle that disables this confetti effect" ????](https://github.com/night/betterttv/discussions/8019) (Q&A) — related confusion about how to disable the effect.


src/constants.js
```diff
@@ -115,6 +115,7 @@ export const ChatFlags = {
   SUB_NOTICE: 1 << 4,
   COMMUNITY_HIGHLIGHTS: 1 << 5,
   CHAT_MESSAGE_HISTORY: 1 << 6,
+  CELEBRATIONS: 1 << 7,
```
dclstn: Does this need migrated or updated somehow? since we're mutating the default

src/modules/hide_celebrations/style.module.css
```diff
@@ -0,0 +1,8 @@
+.hideCelebrations {
+  /* Channel "Celebrations" overlay container rendered over the player */
+  :global(.celebration__overlay),
+  /* Confetti / particle effect canvas used by cheer and celebration effects */
+  :global(.particle-canvas) {
```
night: what if someone wants cheers but not celebrations, or any other ui that uses this?

src/constants.js
night: not unless you broke the default-value-change bit upgrade logic when extensions update. you may wanna check that

PR #8079, opened 2026-06-19
feat: add reply hint tooltip on chat usernames

dclstn:
## Problem

BetterTTV lets you double-click a chat username to insert an `@mention`, but the feature is completely undiscoverable — there's no visual hint that double-clicking does anything.

## Change

Bind a tooltip on each chat username (via the existing `chat.message` watcher event) that reads **"Double-click to reply"** on hover. Reuses the existing `bindTooltip` utility and `formatMessage` for i18n — no new dependencies or settings.

## Notes

- The action inserts an `@mention` into the chat input (not Twitch's threaded reply); wording can be adjusted to "Double-click to mention" if preferred.

### DIFF
--- a/src/modules/doubleclick_mention/index.js
+++ b/src/modules/doubleclick_mention/index.js
@@ -2,6 +2,7 @@ import {off, on} from 'delegated-events';
 import {PlatformTypes} from '@/constants';
 import formatMessage from '@/i18n/index';
 import {bindTooltip} from '@/modules/tooltip/index';
+import storage from '@/storage';
 import {loadModuleForPlatforms} from '@/utils/modules';
 import twitch from '@/utils/twitch';
 import watcher from '@/watcher';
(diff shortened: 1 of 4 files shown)


dclstn:
This may be too distracting on a scrolling chat?

dclstn:
> This may be too distracting on a scrolling chat?

I updated this to hide the hint after 3 uses

src/modules/doubleclick_mention/index.js
```diff
@@ -37,11 +47,30 @@ function handleDoubleClick(e) {
...
+    if (mentionUsageCount >= MENTION_USAGE_HINT_LIMIT) {
+      return;
+    }
+
+    const usernameElement = element.querySelector(CHAT_LINE_USERNAME_SELECTOR);
+    if (usernameElement == null) {
+      return;
+    }
+
+    bindTooltip(usernameElement, {content: formatMessage({defaultMessage: 'Double-click to mention'})});
```
night: when we do educational tooltips like this, we should show somehow that it's a betterttv feature. add our logo?

src/modules/doubleclick_mention/index.js
```diff
@@ -8,6 +11,13 @@ const CHAT_ROOM_SELECTOR = '.chat-list,.chat-list--default,.chat-list--other';
 const CHAT_LINE_SELECTOR = '.chat-line__message';
 const USERNAME_SELECTORS =
   '.chat-line__message span.chat-author__display-name, .chat-line__message span[data-a-target="chat-message-mention"]';
+const CHAT_LINE_USERNAME_SELECTOR = '.chat-author__display-name';
+
+// Stop hinting once the user has clearly learned the gesture.
+const MENTION_USAGE_STORAGE_KEY = 'doubleClickMentionUsage';
+const MENTION_USAGE_HINT_LIMIT = 3;
```
night: are we sure seeing this 3 times isn't annoying? maybe it's like 3 times, but 1 time per day? per session?

PR #8080, opened 2026-06-19
feat: add /screenshot chat command

dclstn:
## Summary

Adds a `/screenshot` chat command that saves the current stream frame as a PNG.

It grabs the player's `<video>` element, draws the current frame to a canvas, and downloads it as `<channel>_<timestamp>.png`. No external API or request — everything happens client-side.

## Behavior

- `/screenshot` → downloads a PNG of the current frame at the video's native resolution
- If the stream isn't playing (no video / zero dimensions), shows a BTTV admin error
- If the frame is cross-origin "tainted" (e.g. during some ads), `drawImage`/`toBlob` is caught and a BTTV admin error is shown instead of throwing

## Implementation

Follows the existing command conventions in `src/modules/chat_commands/commands/` (mirrors `clip.js`). Auto-registered via the existing `import.meta.glob('./commands/*.js')` loader, so no wiring changes are needed.

## Testing

- `npm run lint` (ESLint + Prettier): clean
- Dev build compiles with the new command bundled
- Verified live on a 2560×1440 stream: the draw + PNG encode path produces a valid `image/png` with no cross-origin taint error

### DIFF
--- a/src/modules/chat_commands/commands/screenshot.js
+++ b/src/modules/chat_commands/commands/screenshot.js
@@ -1,0 +1,66 @@
+import {DateTime} from 'luxon';
+import formatMessage from '@/i18n/index';
+import commandStore, {PermissionLevels} from '@/modules/chat_commands/store';
+import {getCurrentChannel} from '@/utils/channel';
+import twitch from '@/utils/twitch';
+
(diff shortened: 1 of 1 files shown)


night (approved):
this is probably fine for now. i imagine users may wanna share these tho

PR #8092, opened 2026-06-25
feat: add setting to hide Twitch AI stream summaries

dclstn:
## What

Twitch is testing a **"Mid-Stream Summary"** feature — an AI-generated recap shown to viewers who join a stream that's already in progress. This adds a setting to hide it.

It's delivered as a chat **private callout** (callout type `mid-stream-summary`) and reuses the generic private-callout chrome (`.chat-private-callout__header-segment`, `.inline-private-callout-line`, …) shared with sub upsells, hype-train rewards, etc., so there's no feature-specific class. The only hooks unique to the summary are the hard-coded element ids the viewer component (`MidStreamSummaryCallout`) renders:

- `#summary-title` — the "Stream Summary" `<h5>` heading
- `#summary-survey-question` — the "Is this summary accurate?" feedback prompt
- the callout shell's `[role="region"][aria-labelledby="summary-title"]`, wrapped by an outer `[role="status"]`

## Changes

- New `hide_ai_summary` module — toggles a `bttv-hide-ai-summary` body class (mirrors `hide_chat_clips`).
- CSS hides `[role="status"]:has([aria-labelledby="summary-title"])`, with the region itself as a fallback if Twitch changes the outer wrapper.
- New `ChatFlags.AI_STREAM_SUMMARY` flag, added to the Chat settings panel as **"AI Stream Summaries"** (default on, so it only hides when unchecked). Included in the CHAT default mask so existing users get it via the flag-upgrade system.

## Notes

The feature is still a server-gated experiment (channel must be eligible + the streamer opts in), so it can't be reproduced on demand yet. Tested against a faithful DOM mock of the callout; the selectors are taken from Twitch's shipping client code rather than guessed, so they should hold up — though they may shift while the feature is experimental.


ara303:
Sorry if this isn't within the repo's etiquette, but I'd like to add a "yes, please" comment to this! :+1:

One thing of note, in the "Notes" section of this PR's body:

> server-gated experiment (channel must be eligible + **the streamer opts in**)

This is incorrect. It's opt-out on the streamer's end. Twitch may be manually selecting streams to have it "opted-in" for them. Minor note, I admit, but it's worth being accurate. 

Several high-profile, four/five viewer digit streamers that I watch expressed surprise that they suddenly had it.

PR #8101, opened 2026-06-29
fix: add missing @emnapi entries to package lock for npm ci

dclstn:
## Problem

`npm ci` fails on a fresh checkout of `master`:

```
npm error Missing: @emnapi/core@1.11.1 from lock file
npm error Missing: @emnapi/runtime@1.11.1 from lock file
```

`@rolldown/binding-wasm32-wasi@1.1.2` (pulled in via the recent vite 8 / `postcss-preset-env` change) declares dependencies on `@emnapi/core@1.11.1` and `@emnapi/runtime@1.11.1`, but those package entries were never written into `package-lock.json`. The lockfile is therefore internally inconsistent, which `npm ci` rejects.

## Fix

Regenerated the lockfile (`npm install --package-lock-only`), which adds the two missing `@emnapi/*@1.11.1` entries (and reconciles some stale `peer` metadata). No actual dependency versions changed.

Verified with a clean `npm ci` — installs successfully with 0 vulnerabilities.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -106,7 +106,6 @@
       "integrity": "sha512-RgHBCvtjbOK2gXSNBNIkNoEc9qoVEtau3hj8gEqKQuL3HZAibKarWFEI3Lfm6EYKkLalOh8eSrj9b+ch9H/VBA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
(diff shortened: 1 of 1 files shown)


night:
fixes via https://github.com/night/betterttv/commit/b1d04de39d7e33fd246516e8d2a96d3d2b6e90e0

PR #8105, opened 2026-06-29
fix: stop tooltips sticking by scoping listeners to the bound element

### LINKED ISSUE #8098: Some Icons stay on screen no matter what
### Describe the bug

I outlined the big parts on discord.

Image 

https://github.com/user-attachments/assets/3ec6029c-760f-449e-a3c5-fa33b97829d7

the icons for some things stick around and wont stop showing the big versions when hovering over. 
this happened around the first of june..  
I am using Firefox ( auto updated over the month a few times ) 
I am using the violent monkey version script along with frankerface ( i tested with this off and it still happens ) 

major extentions i use: 
Bitwarden
Privacy badger
Ublock origine
disable twitch extentions
mute current tab
mute sites by default
tab session manager
unwanted twitch
violent monkey as well

### Steps to reproduce

highlight over a emote

### Expected behavior

emote stops being being when not hovering over

### Screenshots

( linked in first part) 

### Device information

firefox in first box of here

### BetterTTV Version

auto updated from script

### Additional information

_No response_

### Version Check

dclstn:
## Problem

Fixes #8098 — emote (and other) tooltips can get stuck on screen and won't dismiss no matter where the mouse moves.

Since #8031 the tooltip is driven by a global controller that listens for `mouseenter`/`mouseleave` on `document` and **only** closes on a `mouseleave` landing on a tooltip element. But a node removed from the DOM while hovered never fires `mouseleave` (true in every browser). In chat this happens constantly — buffer pruning of old lines, moderation/timeout deletions, emote-menu filtering — so the tooltip stays open, anchored to a now-detached element. The old CSS `:hover` implementation couldn't get stuck because the browser clears `:hover` automatically.

## Fix

Move the open/close triggers off `document` and onto each bound element:

- `bindTooltip` attaches `mouseenter`/`mouseleave` **directly to the element**. The handlers are module-level constants, so `addEventListener` dedups on re-bind, and they're garbage-collected with the element when it scrolls out — no manual cleanup, no document-level delegation that other handlers could interfere with. The element's own `__bttvTooltip` drives a shared singleton tooltip via `openTooltip`/`closeTooltip`.
- For the one case a per-element `mouseleave` still can't catch — the element being removed **while** hovered — the floating-ui `autoUpdate` loop (which only runs while a tooltip is open) is guarded on `referenceElement.isConnected` and closes the tooltip when it detaches.
- Drops the now-unused UUID marker attribute and `crypto.randomUUID()`.

Net −93/+62 lines across 3 files.

## Testing

Verified live on twitch.tv with a dev build, against real BetterTTV emotes carrying their actual `bindTooltip` listeners:

- ✅ hover opens the tooltip
- ✅ **the bug:** removing the hovered element from the DOM (no `mouseleave`) closes the tooltip via the `isConnected` guard — immediately, on the same `autoUpdate` tick the removal triggers
- ✅ ordinary `mouseleave` still closes it
- ✅ `eslint`/`prettier` clean


src/modules/tooltip/TooltipController.jsx
```diff
@@ -1,91 +1,43 @@
...
     return () => {
-      document.removeEventListener('mouseenter', handleMouseEnter, true);
-      document.removeEventListener('mouseleave', handleMouseLeave, true);
+      setActive = null;
     };
-  }, [handleMouseEnter, handleMouseLeave]);
+  }, []);
 
-  return <Tooltip state={tooltipState} />;
+  return <Tooltip state={state} onAutoClose={() => setState((current) => ({...current, open: false}))} />;
```
night: leverage usecallback

src/modules/tooltip/Tooltip.jsx
```diff
@@ -1,22 +1,32 @@
...
   const {refs, floatingStyles, context} = useFloating({
     open: state.open,
     placement: `top-${state.alignment ?? 'center'}`,
     strategy: 'fixed',
     elements: {reference: state.referenceElement},
-    whileElementsMounted: autoUpdate,
+    // mouseleave never fires for a node removed while hovered (common in chat). The scroll that
+    // removes a chat line also drives this autoUpdate loop, so detect the detached reference and
+    // close instead of leaving the tooltip stuck open.
+    whileElementsMounted: (reference, floating, update) =>
```
night: ever render will change this function reference. this may need to be a useCallback

src/modules/tooltip/TooltipController.jsx
```diff
@@ -1,91 +1,43 @@
...
+  open: false,
+  content: null,
+  className: null,
+  alignment: 'center',
+  referenceElement: null,
+};
+
+// The mounted controller registers its setState here so the per-element listeners in
+// index.jsx can drive the single shared tooltip without document-level delegation.
+let setActive = null;
```
night: it may be safer to leverage a store rather than hack on useState. tooltip state is considered a singleton and out of render function updates will propagate properly

night (approved):

PR #8107, opened 2026-06-29
chore: resolve @eslint-react/exhaustive-deps warnings

night:
Follow-up to #8104. Triages all **24** `@eslint-react/exhaustive-deps` warnings.

### Fixed (safe, behavior-preserving)
| File | Change |
|------|--------|
| `EmoteMenu.jsx` | remove unnecessary `focusRef` dep |
| `VirtualizedList.jsx` | remove unnecessary `onHeaderChange` dep |
| `EmoteList.jsx` | add used `ref` dep |
| `VirtualizedList.jsx` | add used `rowHeight` dep |

### Suppressed with reason (20)
The rest are **intentional dependency lists in working code**: mount-only setup effects, stable prop setters/handlers, and deliberately-scoped arrays. Adding the "missing" deps would re-run setup or re-register callbacks and change runtime behavior, so each gets an `// eslint-disable-next-line @eslint-react/exhaustive-deps -- <reason>` rather than a risky edit.

Verified: `eslint .` 0 errors / 0 remaining exhaustive-deps warnings, `prettier --check` clean, production build succeeds.

> Note: overlaps a few files with #8106; whichever merges first, the other may need a trivial rebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8108, opened 2026-06-29
chore: migrate forwardRef to ref-as-prop (React 19)

night:
Follow-up to #8104. Migrates `React.forwardRef` to React 19's ref-as-prop for the **65 of 70** `no-forward-ref` warnings that are mechanical.

### Fixed
- **34 settings panels / group components** — `function X(props, ref)` + `React.forwardRef(X)` → `function X({ref, ...props})` + `X` (drops the wrapper at both the `registerSetting` and `export default` sites).
- **PageHeader, Preview** — inline `forwardRef` arrows → `ref` in the props destructure.

Verified: `eslint .` 0 errors, `prettier --check` clean, production build succeeds.

### Left for a follow-up decision (5 warnings, 3 files)
These merge the forwarded ref with internal refs via `useMergedRef` (and `EmoteList` is `memo`-wrapped with 3 components), so they're called out separately rather than converted blindly:
- `src/common/components/Scrollbar.jsx`
- `src/modules/emote_menu/components/VirtualizedList.jsx`
- `src/modules/emote_menu/components/EmoteList.jsx`

> Note: overlaps a few files with #8106 / #8107; whichever merges first, the others may need a trivial rebase.

### DIFF
--- a/src/modules/emote_menu/components/Preview.jsx
+++ b/src/modules/emote_menu/components/Preview.jsx
@@ -7,7 +7,7 @@ import formatMessage from '@/i18n/index';
 import Icons from './Icons';
 import styles from './Preview.module.css';
 
-const Preview = React.forwardRef(({className, emote}, ref) => {
+const Preview = ({className, emote, ref}) => {
   if (emote == null) return null;
 
   let icon = null;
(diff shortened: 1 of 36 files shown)


KainosSLS:
Ok

PR #8109, opened 2026-06-29
chore: resolve remaining @eslint-react warnings (0 warnings)

night:
Final follow-up to the eslint 10 / @eslint-react migration. Clears the **last 21 warnings** so `eslint .` is completely clean (0 errors, 0 warnings).

### Fixed
- **forwardRef → ref-as-prop** for the 3 components that merge the forwarded ref with internal refs via `useMergedRef` (`Scrollbar`, `VirtualizedList`, `EmoteList`) — completes the React 19 migration started in #8108.
- **lazy `useState` init** where the initial value is a function call (`EmoteMenu`, `VirtualizedList`).
- **purity**: `Footer` computes the copyright year once at module load instead of during render.
- **no-array-index-key**: `Icon` keys svg `<path>`s by their `d` string.

### Suppressed with an inline reason (intentional / stylistic)
- **set-state-in-effect** (7) — effects that legitimately sync external state (DOM nodes, storage subscription, measured layout, image load).
- **use-state** destructure style (2), **no-array-index-key** on a static word list that never reorders (1), **web-api** fire-and-forget timeout (1), **purity** random animation seed on a mutable ref (1).

Verified: `eslint .` 0 errors / 0 warnings, `prettier --check` clean, production build succeeds.

### DIFF
--- a/src/common/components/AnimatedCanvas.jsx
+++ b/src/common/components/AnimatedCanvas.jsx
@@ -10,6 +10,7 @@ const FRAME_INTERVAL = 1000 / FRAME_RATE;
 
 export default function AnimatedCanvas({width = CANVAS_WIDTH, height = CANVAS_HEIGHT, className, ...props}) {
   const ref = useRef(null);
+  // eslint-disable-next-line @eslint-react/purity -- random seed for a mutable animation-time ref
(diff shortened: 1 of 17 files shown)


KainosSLS:
Ok

PR #8115, opened 2026-06-30
fix: dedupe session-lock acquires so a held lock isn't dropped

dclstn:
## Problem

`acquireSessionLock` sends an `acquire_session_lock` frame on **every** call, but the same lock gets nudged repeatedly — `self_bot`'s `updateSessionLock` fires from several triggers (the `SELF_BOT` setting change, the auth-user subscription, and `load.chat`), so one socket can emit two acquires for the same key in quick succession.

When two acquires race, the responses can come back as a contradictory `acquired: true` followed by `acquired: false`. `handleSessionLockUpdate` applies whichever arrives last, so the `false` clears a lock the session actually holds. After that, `hasSessionLock('self_bot')` is `false` and the bot silently stops responding even though it owns the lock, until the page reloads.

## Fix

Track **in-flight** acquires and skip a duplicate while one is already pending, rather than firing an acquire on every call. All three acquire paths (`acquireSessionLock`, the reconnect re-claim in `handleAuthenticationUpdate`, and the retry in `handleSessionLockReleased`) go through one `sendAcquireSessionLock` helper; `pendingSessionLockAcquires` is cleared on every `session_lock_update` response and on reconnect/logout.

Gating on *pending* rather than *desired* matters: a failed or expired grant leaves the key desired but not held, and the server's heartbeat-expiry sends `acquired: false` with **no** `session_lock_released` — so recovery has to come from a fresh acquire. A `desired`-based guard would suppress that and leave the bot stuck; the `pending`-based guard allows it while still collapsing the duplicate burst.

## Validation

Local socket harness against the real lock protocol:
- duplicate acquires no longer desync — the bot keeps the lock and replies
- after a server-injected heartbeat `expired`, the next `updateSessionLock` re-acquires and the bot recovers (a `desired`-based guard fails this; the `pending`-based guard passes)
- reconnect re-acquire and release-retry still work

Prettier passes; the eslint config references a plugin that isn't installed locally, so that's left to CI.

### DIFF
--- a/src/socket-client.js
+++ b/src/socket-client.js
@@ -37,6 +37,13 @@ const joinedChannels = [];
 // session locks this client wants to hold, and the subset the server has granted us
 const desiredSessionLocks = new Set();
 const heldSessionLocks = new Set();
+const pendingSessionLockAcquires = new Set();
+
+// desired intent survives a reconnect; what we hold and what we're mid-acquire does not
+function resetSessionLockConnectionState() {
+  heldSessionLocks.clear();
+  pendingSessionLockAcquires.clear();
+}
 
 function makeSocketChannel(provider, providerId) {
   return `${provider}:${providerId}`;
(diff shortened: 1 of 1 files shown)


src/socket-client.js
```diff
@@ -201,6 +201,10 @@ class SocketClient extends SafeEventEmitter {
   }
 
   acquireSessionLock(key) {
+    if (desiredSessionLocks.has(key)) {
```
night: if the acquire fails, is this removed?

night (approved):

PR #8116, opened 2026-07-01
Complete missing pt-BR translations

Duskorr:
## Summary

This PR completes the `pt-BR.json` translation file, which was missing 218 translation keys that exist in the source code but were never added to the translation file. As a result, many settings descriptions and UI strings were falling back to English instead of displaying in Portuguese.

## Problem

While using BetterTTV with the browser set to Brazilian Portuguese, I noticed several strings in the Settings menu (e.g. under "Sidebar") were still showing in English, even though the setting names themselves were translated.

## What I did

1. Ran the project's own extraction script (`formatjs extract`) to pull every translatable string directly from the source code (`src/**/*.{js,jsx}`), which produces the authoritative list of message keys.
2. Compared the extracted keys against `src/i18n/messages/pt-BR.json` and found 218 keys present in the code but missing from the translation file.
3. Translated all 218 missing strings into natural Brazilian Portuguese.
4. Preserved all ICU placeholders (e.g. `{year}`, `{count}`, `{channelName}`) and inline tags (e.g. `<button>`, `<kbd>`, `<bold>`) exactly as they appear in the source strings, to avoid breaking formatting or interpolation.
5. Validated that the resulting JSON is syntactically valid and that every placeholder/tag in the translated strings matches the original English source 1:1.

## Notes

- I did not modify any existing translations, only added the missing keys.
- I noticed the project's Crowdin integration (linked in the README) appears to be out of sync with the current codebase — several languages show "100%" translated with "0 words" fetched, and pt-BR itself shows 4,158 words translated/approved on Crowdin despite these 218 keys being completely absent there. This PR was created by working directly from the source code rather than relying on Crowdin, since it seems to reflect an outdated snapshot.
- Happy to do the same for `pt-PT.json` if useful, and happy to also update the Crowdin source once the sync issue is addressed by maintainers.

## Testing

Ran `npm start` locally and verified via the userscript workflow described in the README that the previously-English strings (e.g. "Recommended Categories" description, "Similar Channels" description, "Offline Followed Channels" description, "Auto-Expand Channels" description) now render correctly in Portuguese.


night:
we cannot accept translations strings submitted via github PRs. you'll need to submit translations via Crowdin at https://crowdin.com/project/betterttv

Duskorr:
Accept my join request so I can upload the translation

PR #8123, opened 2026-07-03
fix: preserve flag defaults when syncing settings from cloud backup

dclstn:
## Problem

Cloud backup pulls strip flag settings down to the flags number (`[value] = value`) and re-apply them through `settings.set()`, which recomputes `changedBits` from the diff against the local value. When a new default-on flag ships, the sequence on a cloud-backup user's device is:

1. `upgrade()` adds the new bit locally (it's not in `changedBits`)
2. the pull applies the server copy — written by a pre-update client, so the bit is off
3. `settings.set()` records that diff into `changedBits`, permanently marking the new flag as a deliberate user change

The upgrade pass then never re-adds it, so every Pro user with cloud backup enabled silently gets the old default forever. This currently affects any newly added default-on flag (e.g. #8091's watch streaks flag, which is how this was found — the reverted value also gets pushed back to the server, poisoning other devices).

## Fix

- Sync the full `[flags, changedBits]` tuple on pull instead of flags-only. The push side already uploads tuples wholesale, so the server has always stored `changedBits` — the pull just discarded them. Deliberate opt-outs now transfer across devices without being misattributed.
- Re-run the flag defaults pass (extracted from `upgrade()` into `settings.applyFlagDefaults()`) after applying server settings, so default-on bits the user never touched are inferred over a stale server copy and pushed back once.
- The defaults pass now only writes when the flags actually change. It previously called `settings.set()` on every run even with no change; harmless in `upgrade()` (before listeners attach), but after a pull it would fire a settings push containing no changes, and two online devices would ping-pong version bumps indefinitely.

## Verification

Tested in-browser against a stubbed settings endpoint:

- Stale server copy `chat: [61, 2]` (Bits off, predates the Message History bit): local converges to `[125, 66]` — the Bits opt-out syncs, Message History is inferred on, and the merged tuple is pushed back once.
- Deliberate opt-out from an up-to-date device `chat: [63, 64]`: survives the pull untouched, is not re-added by the defaults pass, and triggers no push.

Ideally this lands before #8091 so no production profiles get poisoned `changedBits` (which is indistinguishable from a real opt-out after the fact).

### DIFF
--- a/src/modules/cloud_backup/index.js
+++ b/src/modules/cloud_backup/index.js
@@ -2,7 +2,7 @@ import debounce from 'lodash.debounce';
 import isEqual from 'lodash.isequal';
 import {getExtensionSettings, updateExtensionSettings} from '@/actions/extension';
 import {openConfirmModal} from '@/common/utils/modal';
-import {CLOUD_BACKUP_SETTINGS_STORAGE_KEY, FlagSettings} from '@/constants';
+import {CLOUD_BACKUP_SETTINGS_STORAGE_KEY} from '@/constants';
 import formatMessage from '@/i18n/index';
 import settings from '@/settings';
 import socketClient, {EventNames} from '@/socket-client';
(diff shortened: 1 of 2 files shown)


src/modules/cloud_backup/index.js
night, line 139: value here is now an array. does that work?

src/settings.js
```diff
@@ -151,8 +155,9 @@ class Settings extends SafeEventEmitter {
       // upgrade flags where default bits changed
       const [oldFlags, oldChangedBits] = settings[flagSettingId];
       const flagsToAdd = setFlag(defaultFlags, oldChangedBits, false);
-      if (flagsToAdd > 0) {
-        this.set(flagSettingId, setFlag(oldFlags, flagsToAdd, true));
+      const newFlags = setFlag(oldFlags, flagsToAdd, true);
+      if (newFlags !== oldFlags) {
```
night, line 163: this looks like it may change the flag default value in a breaking way.

we really only want to add a flag when it is a previously unchanged bit. i.e. upgrades only.

the logic we may want is something like:

1. settings sync from cloud storage
2. flags values copy into settings value for value, including changed bits
2. upgrade applies, and if there is an unchanged bit we change that one flag per usual

PR #8132, opened 2026-07-10
feat: add badge autocomplete and improve keyword settings comboboxes

dclstn:
image

## What

Improves autocomplete across the keyword settings, and adds badge suggestions:

- **Badge autocomplete** — setting a keyword row's target to Badge now suggests Twitch's global badges (~437, fetched once via GQL through Twitch's own Apollo client and cached), rendered with the badge image and title and filtered by title as you type. Suggestions match on the badge title, which is what keyword matching runs against (the chat badge's alt text).
- **Exact-match completion** — picking a badge inserts an anchored regex (`~/^…$/`) so titles like `Subscriber` don't also match `6-Month Subscriber`. Manually typed keywords are stored as-is.
- **Shared combobox styling** — dropdown styles for `Autocomplete`/`TagsInput` now live in the theme: symmetric 4px insets where the top/bottom breathing room scrolls with the options (dropdown clips to its rounded shape), option border radius that complements the dropdown radius, and a drop shadow.
- **Channels field alignment** — the channels input is now inset with an outer margin (matching how the keyword input is inset by its cell), so its dropdown naturally aligns with the field's visible bounds instead of spanning the full column.

## Why

Badge keywords previously required knowing a badge's exact title, and a partial title could unintentionally match other badges. The settings comboboxes also had inconsistent dropdown geometry (full-bleed options clipping against rounded corners, mismatched insets between fields).

## Testing

- Verified the GQL query shape against gql.twitch.tv (437 badges, 407 unique titles).
- Exercised via the dev server build; lint, prettier, and build pass. No new i18n strings.


src/modules/settings/components/SettingKeywords.jsx
```diff
@@ -90,6 +96,83 @@ function openRegexGuideModal() {
...
+  } catch (e) {
+    debug.log('failed to fetch twitch global badges', e);
+    badgeOptionsPromise = null;
+    return [];
+  }
+
+  return buildBadgeOptions(globalBadges);
+}
+
+function useBadgeOptions(enabled) {
```
night: pull this logic out into a different helper hook file. is it not needed here.

PR #8136, opened 2026-07-12
qol: dim the default color of setting radio cards

dclstn:


dclstn:
Folded into #8135 so the badge-picker changes ship together.

PR #8160, opened 2026-07-24
feat: add pro upgrade page

dclstn:
Adds a Pro upgrade page to the settings modal: a fixed dark surface with a sticker emote hero, a Pro feature checklist, and an upgrade button. The sidenav bottom button links to it, Pro badges click through to it, and the changelog moved to the footer version number.


dclstn:
https://github.com/user-attachments/assets/eb02a320-2b5f-49e0-b203-970bdf524e96

src/modules/settings/components/Footer.jsx
```diff
@@ -47,6 +49,13 @@ function LinkColumn({title, links}) {
 }
 
 function Footer() {
+  const {setPage} = use(PageContext);
+  const portalRef = usePortalRef();
+
+  function handleVersionClick() {
```
dclstn, line 55: Is this too hidden away now? Maybe we do what discord does with a "What's New" modal?

src/modules/settings/pages/ProHome.jsx
```diff
@@ -0,0 +1,81 @@
...
+            color="contrast"
+            rightSection={<Icon icon={faArrowUpRightFromSquare} size={14} />}
+            onClick={handleUpgradeClick}>
+            {isPro
+              ? formatMessage({defaultMessage: 'Manage Subscription'})
+              : formatMessage({defaultMessage: 'Upgrade to Pro'})}
+          </Button>
+        </div>
+      }>
+      <div className={styles.featureColumns}>
```
dclstn, line 67: How should these be ordered?

src/modules/settings/components/pro/UsernameEffectShowcaseCard.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+const MARQUEE_ROWS = [
+  {id: 'row-1', direction: -1},
+  {id: 'row-2', direction: 1},
+  {id: 'row-3', direction: -1},
+  {id: 'row-4', direction: 1},
+  {id: 'row-5', direction: -1},
+];
+
+// js-driven on purpose: a css transform animation of these rows measured 23fps in chrome
```
night, line 30: was the css version tested with `will-change: transform` and the sweep frozen like it is now? i'd expect a composited css animation to beat 5 rAF loops on the main thread. do you have the trace?

src/modules/settings/components/pro/UsernameEffectShowcaseCard.jsx
```diff
@@ -0,0 +1,104 @@
...
+  const x = useMotionValue(0);
+  const groupRef = useRef(null);
+
+  useAnimationFrame((time, delta) => {
+    const group = groupRef.current;
+    if (group == null) {
+      return;
+    }
+
+    const width = group.offsetWidth;
```
night, line 43: this reads `offsetWidth` every frame in all 5 rows. could we measure once (or on resize) instead? also does this keep running when the card is scrolled out of view?

src/modules/settings/components/pro/CloudBackupShowcaseCard.module.css
```diff
@@ -0,0 +1,122 @@
...
+
+.cloudGlyph {
+  position: absolute;
+  font-family: monospace;
+  font-size: 11px;
+  font-weight: 700;
+  line-height: 1;
+  color: rgba(219, 226, 255, 0.6);
+  offset-rotate: 0deg;
+  animation-name: glyph-rise;
```
night, line 75: that's 60 infinite `offset-distance` animations under a `mask-image`, and i don't think chrome composites motion path. what does this card cost per frame? none of the new animations (this, `cloud-bob`, the marquee) seem to respect reduced motion either, should they?

src/modules/settings/components/pro/CloudBackupShowcaseCard.jsx
```diff
@@ -0,0 +1,60 @@
...
+      <img className={styles.cloudBeamSvg} src={cdn.url('assets/pro/cloud_beams.webp')} alt="" loading="lazy" />
+      <div className={styles.cloudStreams} aria-hidden="true">
+        {DATA_STREAMS.map((stream, streamIndex) => (
+          <React.Fragment key={stream.id}>
+            {DATA_STREAM_SLOTS.map((slotKey, slotIndex) => (
+              <span
+                key={`${stream.id}-${slotKey}`}
+                className={stream.glyphClassName}
+                style={{
+                  animationDuration: `${stream.duration}s`,
```
night, line 40: not a fan of generating styles in js. could duration live on the per-stream classes and the delay come from an `--i` custom property or `nth-child` in the stylesheet?

src/modules/settings/components/pro/UsernameEffectShowcaseCard.jsx
```diff
@@ -0,0 +1,104 @@
+import {Title} from '@mantine/core';
+import {motion, useAnimationFrame, useMotionValue} from 'framer-motion';
+import shuffle from 'lodash.shuffle';
```
night, line 3: do we need a new dep to shuffle 5 items? could each row just be a fixed rotation of the list?

src/common/components/ProBadge.jsx
```diff
@@ -1,11 +1,38 @@
...
+import React, {use} from 'react';
+import {PageTypes} from '@/constants';
 import formatMessage from '@/i18n/index';
+import {PageContext} from '@/modules/settings/contexts/PageContext';
+import styles from './ProBadge.module.css';
 
 // Props (including ref) pass through to Badge so a wrapping Tooltip can anchor to it.
-export default function ProBadge(props) {
+export default function ProBadge({clickable = true, ...props}) {
+  const pageContext = use(PageContext);
```
night, line 11: a `common` component importing settings `PageContext` feels backwards. could `SettingWrapper` pass an `onClick` instead? it's the only clickable caller.

src/common/components/ProBadge.jsx
```diff
@@ -1,11 +1,38 @@
...
+
   return (
-    <Badge color="indigo" variant="elevated" size="lg" {...props}>
+    <Badge
+      color="indigo"
+      variant="elevated"
+      size="lg"
+      {...props}
+      className={classNames(styles.clickable, props.className)}
+      onClick={handleClick}>
```
night, line 35: this is a div with an `onClick`, so no keyboard focus or role. should it render as a button when clickable?

src/constants.js
```diff
@@ -158,6 +158,7 @@ export const PageTypes = {
   HIGHLIGHT_KEYWORDS: 4,
   BLACKLIST_KEYWORDS: 5,
   SELF_BOT_COMMANDS: 6,
+  PRO_HOME: 7,
```
night, line 161: heads up, #8201 also claims `7` for `SELF_BOT_TIMERS`. whichever lands second needs a bump.

src/modules/settings/pages/ProHome.jsx
```diff
@@ -0,0 +1,81 @@
...
+            {label: formatMessage({defaultMessage: 'BetterTTV Pro'})},
+            {label: formatMessage({defaultMessage: 'Perks'})},
+          ]}
+        />
+      }
+      footer={
+        <div className={styles.premiumActions}>
+          <Text size="md" className={styles.premiumThanks}>
+            <Icon icon={faHeart} size={16} className={styles.premiumThanksHeart} />
+            {formatMessage({defaultMessage: 'Thank you for supporting BetterTTV'})}
```
night, line 47: this shows "thank you for supporting" to non-pro users right next to "Upgrade to Pro". should it be pro-only, or different copy for free users?

src/modules/settings/components/SettingsModal.module.css
```diff
@@ -62,7 +62,7 @@
   flex-direction: column;
   gap: 20px;
   /* gap between the header and the first item */
-  padding: 20px 20px 24px;
+  padding: 20px;
```
night, line 65: this changes bottom padding on every settings page, not just the pro one. intentional?

src/modules/settings/components/pro/CloudBackupShowcaseCard.module.css
```diff
@@ -0,0 +1,122 @@
...
+    transform: translateY(-4px);
+  }
+}
+
+.cloudGraphic {
+  display: block;
+  width: 100%;
+}
+
+/* ios-style app icon at the top of the pane, as the thing being backed up; sits above the
```
night, line 29: this comment describes an outer span + inner body with a 3d tilt, but it's an `img` now. `.showcaseLabel` mentions a backdrop blur mask that isn't there, and `.cloudBeamSvg` is a webp. can we do a pass on stale comments/names?

night (changes requested):
see inline comments

src/modules/settings/components/Footer.jsx
night: footer link is fine. we don't want to nag users on every update, so no "what's new" modal. tucked away is the goal here.

src/modules/settings/pages/ProHome.jsx
night: username effects + emotes up top, then badges / priority approval, theming / self bot, cloud backup / autocomplete?

also the cards are all 232px, do we need multicol? a 2 col grid would make dom order = reading order.

src/modules/settings/components/pro/UsernameEffectShowcaseCard.jsx
dclstn: the 23fps trace was from before the effect sweeps were frozen — the animated gradient text was what refused to composite, not the transform. rewrote it as a css animation with `will-change` (rAF loops and per-frame `offsetWidth` reads are gone) and re-traced on the current build: 600/600 frames at 60fps while scrolling, 0 long tasks, p95 16.8ms, worst 17.7ms.

src/modules/settings/components/pro/CloudBackupShowcaseCard.module.css
dclstn: traced the full page scrolling on the current build: 60fps over 600 frames, 0 janky frames, 0 long tasks, worst frame 17.7ms — the offset-path glyphs are absorbed fine. added `prefers-reduced-motion: reduce` handling for the glyphs, the cloud bob, and the marquee.

PR #8161, opened 2026-07-24
chore(gitignore): ignore common OS/editor metadata

danielblake638:
The existing `.gitignore` doesn't cover macOS `.DS_Store`, Windows `Thumbs.db`, and JetBrains/VS Code project folders. Contributors on those platforms occasionally end up committing them by accident.

Appending standard entries so this doesn't happen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,21 @@ dist
 .vscode
 .idea
 .env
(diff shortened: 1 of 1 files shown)


night:
this doesn't seem to be needed

PR #8168, opened 2026-07-30
fix(7tv): update for v3 api changes

zxr173:
Updates 7TV channel emotes logic to accommodate an upcoming 7TV v3 API change where the `/v3/users/:platform/:platform_id` endpoint will no longer return every emote from the active emote set under `emote_set`.


dclstn:
@zxr173 lgtm, but you are aware this effectively doubles all requests?

PR #8181, opened 2026-08-14
fix: tolerate invalid stored primaryColor values

dclstn:
Legacy Pro accent colors were stored as arbitrary hex values (e.g. `#70cfce`), and cloud backup sync can still restore them into settings. Mantine only accepts keys of `theme.colors` as `theme.primaryColor`, so an invalid stored value crashed the UI in two places once auth resolved the user as Pro:

- `MantineProvider` threw `Invalid theme.primaryColor, it accepts only key of theme.colors` when rendering the settings UI
- the accent color module threw `Cannot read properties of undefined (reading '9')` reading `theme.colors[value][9]` on page load

### Fix

- Normalize the stored value against `theme.colors` before it reaches `MantineProvider` or the CSS variables resolver, falling back to `DEFAULT_PRIMARY_COLOR` (whose doc comment already promised a fallback for "unset or invalid" values).
- Treat palette-less colors as an unset accent in the accent color module instead of crashing.
- `mantineVariablesResolver` no longer mutates the shared module-level theme, which would have poisoned every other consumer with the unvalidated value.

### Testing

- Reproduced both crashes on a profile with `primaryColor: "#70cfce"` in `bttv_settings` (restored by cloud backup sync).
- With the fix: no console errors on load, the settings modal renders and navigates normally, and the accent falls back to the default red. Selecting palette colors still works.

### DIFF
--- a/src/modules/accent_color/index.js
+++ b/src/modules/accent_color/index.js
@@ -36,6 +36,16 @@ class AccentColor {
     const themeColor = parseThemeColor({color: value, theme});
     const baseColors = theme.colors[themeColor.color];
 
+    // Legacy Pro accent colors were stored as arbitrary hex values, which have no palette in
+    // theme.colors — treat them like an unset accent rather than crashing.
+    if (baseColors == null) {
+      if (cssVariablesStyle != null) {
+        cssVariablesStyle.remove();
+      }
+
+      return;
+    }
+
     const twitchPurpleColors = {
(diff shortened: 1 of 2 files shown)


dclstn:
Closing — the invalid value turned out to be a one-off malformed entry in my own settings backup, not a legacy storage format. Fixed the stored data instead; no code change needed.

Issue #6575, opened 2024-01-15
Disable auto play for homepage video player pausing main video player


Zilox135:
### Describe the bug

The feature to disable the auto play of homepage video player has some unintended side affects. If a stream from the following list is clicked on before the homepage is fully loaded, the active stream that was clicked on is paused automatically instead of the homepage streams. This can also happen with a refresh of the page when watching a stream.

### Steps to reproduce

1. Turn off the **Front-page Player (Enable auto play for homepage)** video player setting in the channel settings options.

2. Close the browser completely and reopen it.

3. Go to twitch.tv and click on a stream from the followed channels before the homepage has time to fully load.

4. Observe that the stream that was clicked on gets automatically paused after about 5-10 seconds.

### Expected behavior

Only streams on the homepage should be automatically paused and not one's that have been clicked on by the user.

### Screenshots

![gkMsXyt](https://github.com/night/betterttv/assets/89009170/238c15fb-a372-4293-8e18-b6ebae74bc7c)

### Device information

**Browser** - Brave Browser - v1.61.116 (Jan 10, 2024)
**OS** - Windows 11 Pro

PR #8195, opened 2026-08-24
fix: stop the autoplay settings leaving stale state on other players

### LINKED ISSUE #8192: Live stream sometimes starts paused
### Describe the bug

When visiting a live stream, the video player sometimes starts paused.

### Steps to reproduce

1. In BTTV settings, disable Autoplay for offline channels.
2. Visit and offline stream page. -- The video does get paused as expected.
3. Navigate to a live stream. -- The live stream is paused.
4. Reload the page. -- The live stream plays immediately.

### Expected behavior

The setting to disable autoplay for offline streams should not affect live streams.

### Screenshots

Settings
Paused live stream

### Device information

- Device: PC
- OS: Windows 11
- Browser: Google Chrome
- Version: 151.0.7922.173

### BetterTTV Version

7.7.25

### Additional information

I suspect the video player element is reused when navigating between pages. Since it got paused on one page, it remains paused on the next page.

### Version Check

dclstn:
Fixes #8192

`disable_offline_channel_autoplay` arms a one-shot `Playing` listener that only unregisters itself when it fires, so on an offline page, where the carousel gets paused before it ever plays, the listener often stays attached. Twitch reuses the player instance across the SPA navigation to a live channel, so it fires there instead and pauses the live stream, which is why the report saw it intermittently and why one reload cleared it. It now re-checks the carousel and returns early, the same guard `disable_homepage_autoplay` got in 4d8d01b for #6575. Auditing the other autoplay toggles turned up one more: `watchPlayerRecommendationVodsAutoplay` registers a fresh dom watcher on every `load.player` while the setting is off and keeps only the newest remover, so the callbacks pile up for the life of the tab and re-enabling the setting leaves the older ones clicking cancel until you reload. It now arms one and leaves it alone. Lint, prettier and build are clean, but video would not decode in my test browser, so `Playing` never fired and neither fix is verified end to end. Both toggles want a manual pass before merge.

### DIFF
--- a/src/modules/disable_homepage_autoplay/index.js
+++ b/src/modules/disable_homepage_autoplay/index.js
@@ -32,11 +32,12 @@ class DisableHomepageAutoplayModule {
 
     const stopAutoplay = () => {
       setTimeout(() => {
-        if (document.querySelector(FEATURED_VIDEO_SELECTOR) == null) {
-          return;
+        // this listener only removes itself once it fires, so it can outlive the featured video
+        // and end up pausing the player of whatever page was navigated to afterwards
+        if (document.querySelector(FEATURED_VIDEO_SELECTOR) != null) {
+          currentPlayer.pause();
         }
 
-        currentPlayer.pause();
         currentPlayer.setMuted(prevMuted);
       }, 0);
       if (currentPlayer.emitter) {
--- a/src/modules/disable_homepage_autoplay/index.js
+++ b/src/modules/disable_homepage_autoplay/index.js
@@ -32,12 +32,11 @@ class DisableHomepageAutoplayModule {
 
     const stopAutoplay = () => {
       setTimeout(() => {
-        // this listener only removes itself once it fires, so it can outlive the featured video
-        // and end up pausing the player of whatever page was navigated to afterwards
-        if (document.querySelector(FEATURED_VIDEO_SELECTOR) != null) {
-          currentPlayer.pause();
+        if (document.querySelector(FEATURED_VIDEO_SELECTOR) == null) {
+          return;
         }
 
+        currentPlayer.pause();
         currentPlayer.setMuted(prevMuted);
       }, 0);
       if (currentPlayer.emitter) {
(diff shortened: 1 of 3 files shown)


src/modules/disable_homepage_autoplay/index.js
```diff
@@ -32,11 +32,12 @@ class DisableHomepageAutoplayModule {
 
     const stopAutoplay = () => {
       setTimeout(() => {
-        if (document.querySelector(FEATURED_VIDEO_SELECTOR) == null) {
-          return;
+        // this listener only removes itself once it fires, so it can outlive the featured video
```
night: this change looks like it doesn't change anything? you invert the selector check, so now instead of it muting only featured videos it mutes all?

src/modules/disable_offline_channel_autoplay/index.js
```diff
@@ -23,7 +25,12 @@ class DisableOfflineChannelAutoplayModule {
 
     const stopAutoplay = () => {
       setTimeout(() => {
-        currentPlayer.pause();
+        // this listener only removes itself once it fires, so it can outlive the offline carousel
+        // and end up pausing the player of a live channel navigated to afterwards
+        if (document.querySelector(CAROUSEL_PLAYER_SELECTOR) != null) {
```
night: similarly here, it looks like this would need to check the selector and early return so we dont mute

src/modules/video_player/index.js
```diff
@@ -56,8 +56,14 @@ function createPictureInPictureButton(toggled) {
 let removeRecommendationWatcher;
 
 function watchPlayerRecommendationVodsAutoplay() {
+  // this runs on every player load, so drop the previous watcher before arming another one.
+  // otherwise they stack up and re-enabling the setting only removes the most recent one.
```
night: is this true? is the setting flag changing between runs? maybe the correct fix is to check if the watcher is registered and earlier return after the flag check?

PR #8201, opened 2026-08-28
feat: add self bot timers

dclstn:
**What**

- New Self Bot Timers page (Bots → Self Bot → Timers → Edit). Each timer has a message, an interval in minutes, and a Nightbot-style Chat Lines guard. The interval field is a plain number input (min 1, clamped on blur/Enter). Chat Lines mirrors Nightbot: the minimum number of non-broadcaster messages in the last 5 minutes, checked once per interval, with a failed check skipping that interval. Minimum (and default) is 2, capped at 100; the guard is always on.
- Countdowns start at activation and reset on deactivation; nothing persists across page loads. A newly added timer waits a full interval before its first send. Like Nightbot, timers only post while the stream is live. Timer messages cannot start with / or . and cap at 500 characters, so a timer can never run chat commands unattended.
- The Timers row shows Pro and New badges, and a fresh settings promotion dot (selfBotTimers) so users who already dismissed the Self Bot promotion get pointed at timers.
- Commands also gained per-row enable toggles and an example placeholder, and the three settings entry tables (commands, timers, keywords) now share one state hook.

### DIFF
--- a/src/modules/self_bot/index.js
+++ b/src/modules/self_bot/index.js
@@ -4,12 +4,17 @@ import socketClient from '@/socket-client';
 import useAuthStore from '@/stores/auth';
 import {messageTextFromAST} from '@/utils/chat-message-text';
 import {loadModuleForPlatforms} from '@/utils/modules';
+import {isUserPro} from '@/utils/pro';
 import twitch from '@/utils/twitch';
 import {getCurrentUser} from '@/utils/user';
 import watcher from '@/watcher';
 import {computeSelfBotCommands, matchesCommand, matchesUserLevel} from './commands';
+import {computeSelfBotTimers} from './timers';
 
 const COMMAND_COOLDOWN_MS = 2000;
+const TIMER_TICK_INTERVAL_MS = 15 * 1000;
+// mirrors nightbot: a timer's chat lines requirement is measured over the last 5 minutes
+const TIMER_CHAT_LINES_WINDOW_MS = 5 * 60 * 1000;
 // only one session per user may hold this lock, ensuring a single session replies
 const SELF_BOT_SESSION_LOCK = 'self_bot';
 
@@ -17,8 +22,20 @@ let computedCommands = [];
 const commandCooldowns = new Map();
 let loadTime = Date.now();
 
+let computedTimers = [];
+// timer id -> when the timer last completed a check: a send, a skipped
+// (guard-failed) check, or when it was first seen
+const timerLastCheckTimes = new Map();
+// send times of recent non-broadcaster messages, pruned to the window
+const recentChatLineTimes = [];
+let timersTickInterval = null;
+
 function recomputeCommands() {
   computedCommands = computeSelfBotCommands(settings.get(SettingIds.SELF_BOT_COMMANDS_LIST));
+}
+
+function recomputeTimers() {
+  computedTimers = computeSelfBotTimers(settings.get(SettingIds.SELF_BOT_TIMERS_LIST));
 }
 
 function isSelfBotActive() {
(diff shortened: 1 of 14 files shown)


dclstn:
https://github.com/user-attachments/assets/1f32f1c2-440d-4c92-a1ca-3ab39232d826

night (commented):
see inline comments

src/modules/self_bot/index.js
```diff
@@ -30,12 +56,132 @@ function isSelfBotActive() {
...
+
+    const dueAt = anchor.time + timer.intervalMinutes * 60 * 1000;
+    if (now < dueAt) {
+      continue;
+    }
+
+    // dead chat guard: the required chat lines are counted since the timer
+    // last sent, so a due timer holds until chat catches up rather than
+    // skipping a full interval
+    if (chatLineCount - anchor.lineCount < timer.lines) {
```
night: description says this mirrors nightbot (lines in the last 5 min, checked once per interval, failed check skips the interval), but this counts since the last send and holds until chat catches up. which one do we want? with hold, don't all the held timers fire 15s apart the moment chat wakes up?

src/modules/self_bot/timers.js
```diff
@@ -0,0 +1,45 @@
+export const TIMER_MIN_INTERVAL_MINUTES = 1;
+export const TIMER_MAX_INTERVAL_MINUTES = 24 * 60;
+export const TIMER_MIN_CHAT_LINES = 2;
```
night, line 3: description says 0 disables the guard and new entries default to 2, but the min is 2 here and `computeSelfBotTimers` drops anything lower. which is intended? should min and default be separate constants?

src/modules/self_bot/index.js
```diff
@@ -30,12 +56,132 @@ function isSelfBotActive() {
...
+      dueTime = dueAt;
+    }
+  }
+
+  if (dueTimer == null) {
+    return;
+  }
+
+  timerSendAnchors.set(dueTimer.id, {time: now, lineCount: chatLineCount});
+  twitch.sendChatMessage(dueTimer.message);
```
night, line 134: should timer messages starting with `/` or `.` be allowed? wouldn't this run `/clear`, `/raid` etc unattended? also does a 500+ char message just fail every interval? maybe a `maxLength` on the input?

src/modules/self_bot/index.js
```diff
@@ -4,23 +4,49 @@ import socketClient from '@/socket-client';
...
 import {loadModuleForPlatforms} from '@/utils/modules';
+import {isUserPro} from '@/utils/pro';
 import twitch from '@/utils/twitch';
 import {getCurrentUser} from '@/utils/user';
 import watcher from '@/watcher';
 import {computeSelfBotCommands, matchesCommand, matchesUserLevel} from './commands';
+import {computeSelfBotTimers} from './timers';
 
 const COMMAND_COOLDOWN_MS = 2000;
+const TIMER_TICK_INTERVAL_MS = 15 * 1000;
```
night, line 17: how does this behave in a background tab or popout? chrome throttles hidden-page timers to 1/min after a few minutes. with one send per tick, do multiple timers back up?

src/modules/settings/components/SettingSelfBotCommands.jsx
```diff
@@ -31,12 +32,23 @@ const FREE_COMMAND_LIMIT = 5;
...
+  const handleEnabledChange = useCallback(({target: {checked}}) => onUpdate({enabled: checked}), [onUpdate]);
   const commandInputRef = useCallback((ref) => commandInputRefCallback(id, ref), [commandInputRefCallback, id]);
 
   const userLevel = data.userLevel ?? SelfBotUserLevels.EVERYONE;
 
   return (
     <TableTr {...props}>
+      <TableTd className={classNames(tableStyles.dataCellMiddle, tableStyles.toggleColumn)}>
+        <label className={tableStyles.toggleCell}>
+          <Checkbox
```
night, line 45: the enabled toggle + placeholder changes for commands aren't in the description. intentional for this PR or should they be split out?

src/modules/settings/components/SettingSelfBotTimers.jsx
```diff
@@ -0,0 +1,353 @@
...
+}
+
+function SettingSelfBotTimers({value, setValue}) {
+  const entryList = useMemo(() => Object.entries(value ?? {}).reverse(), [value]);
+  const pendingMessageFocusRef = useRef(null);
+  const bttvUser = useAuthStore(useShallow((state) => state.user));
+
+  const isPro = isUserPro(bttvUser);
+
+  const newEntryHandler = useCallback(() => {
```
night, line 271: new/update/delete handlers + the focus ref look copied from `SettingSelfBotCommands` (and `SettingKeywords` has the same shape). worth a shared hook now that there are three?

src/modules/settings/components/SettingSelfBotTimers.jsx
```diff
@@ -0,0 +1,353 @@
...
+    );
+  }
+
+  return formatMessage(
+    {defaultMessage: 'Every {minutes, plural, one {minute} other {# minutes}}'},
+    {minutes: intervalMinutes}
+  );
+}
+
+function TimerNumberInput({value, min, max = Infinity, onCommit, formatValue, unitLabel}) {
```
night, line 77: nit: should chat lines have a max? a huge value parses fine and the timer just never fires.

src/modules/self_bot/index.js
```diff
@@ -56,18 +202,28 @@ class SelfBotModule {
     watcher.on('load.chat', () => {
       loadTime = Date.now();
       recomputeCommands();
-      updateSessionLock();
+      recomputeTimers();
+      updateSelfBotState();
+    });
+    watcher.on('chat.message', (_, messageObj) => {
+      countTimerChatLine(messageObj);
```
night: nit: `isExternalChatMessage` now runs twice per chat message when timers are scheduled (here and in `onMessage`). could we compute it once in this handler?

src/modules/self_bot/index.js
```diff
@@ -30,12 +56,132 @@ function isSelfBotActive() {
...
+  if (!isSelfBotActive()) {
     socketClient.releaseSessionLock(SELF_BOT_SESSION_LOCK);
+    return;
+  }
+
+  socketClient.ensureAuthentication();
+  socketClient.acquireSessionLock(SELF_BOT_SESSION_LOCK);
+}
+
+function sendDueTimerMessage() {
```
night, line 72: should timers only run while the stream is live? as written they post to offline chat as long as the tab is open and 2 people talk.

src/modules/settings/components/SettingSelfBotTimers.jsx
```diff
@@ -0,0 +1,353 @@
...
+
+  const handleBlur = useCallback(() => {
+    const parsedNumber = Number.parseInt(draft, 10);
+    setDraft(null);
+
+    if (Number.isNaN(parsedNumber)) {
+      return;
+    }
+
+    onCommit(Math.min(Math.max(parsedNumber, min), max));
```
night: nit: this commits on every blur even if the value didn't change, so focus + blur is a storage write and a `recomputeTimers`. skip when it equals `value`?

src/modules/settings/components/SettingSelfBotTimers.jsx
```diff
@@ -0,0 +1,353 @@
...
+  const onIntervalCommit = useCallback((intervalMinutes) => onUpdate({intervalMinutes}), [onUpdate]);
+  const onLinesCommit = useCallback((lines) => onUpdate({lines}), [onUpdate]);
+  const handleEnabledChange = useCallback(({target: {checked}}) => onUpdate({enabled: checked}), [onUpdate]);
+  const messageInputRef = useCallback((ref) => messageInputRefCallback(id, ref), [messageInputRefCallback, id]);
+
+  return (
+    <TableTr {...props}>
+      <TableTd className={classNames(tableStyles.dataCellMiddle, tableStyles.toggleColumn)}>
+        <label className={tableStyles.toggleCell}>
+          <Checkbox
```
night, line 157: nit: should the toggle have an `aria-label`? the info button has one. same for the commands table.

night (changes requested):
see inline comments

PR #8220, opened 2026-09-10
feat: add twitch gif picker to emote menu

### LINKED ISSUE #8210: Twtich New Feature - Send GIF in the chat
### Describe the bug

I noticed that with the new function that Twitch released to send GIFs in chat, the Betterttv emotes menu does not show the GIF option to choose and send. In a browser without the extension enabled, the new option normally appears in the Twitch chat menu.

### Steps to reproduce

1. Click in emote icon in the chat menu

### Expected behavior

Show two options - emotes and gif (new)

### Screenshots

Image

Image

### Device information

Windows 11
Brave

### BetterTTV Version

7.7.25

### Additional information

_No response_

### Version Check

dclstn:
Twitch lets Tier 2/3 subscribers post GIPHY GIFs in chat, but the selector only exists in the native emote picker. This adds a GIFs mode to the BetterTTV emote menu.

Closes #8210

**What**
- A segmented emote/GIF toggle next to the emote menu search input, shown only on channels where Twitch's gif picker is available (`gifPickerConfig` allowlisted with an api key, and a sub product with `hasGifs`).
- GIF mode reuses the search input (`Search GIPHY` placeholder): trending by default, search with a 300ms debounce, a two-column masonry capped at 20 results. Requests use the channel's GIPHY api key with the content rating capped at `pg`, matching what the native picker is willing to display.
- Clicking a gif sends it via the `sendGifMessage` GQL mutation. The server cooldown turns each tile's Send pill into a disabled countdown (ring progress against the 30s cooldown plus the remaining seconds), armed immediately on successful sends too. `TEMPORARILY_UNAVAILABLE` shows an unavailable state and viewers below Tier 2 an explanatory state instead of the grid, both styled like the emote list's empty-search state.
- Gif availability and results live in a zustand store (`gif-picker-store`), fetched from event handlers (menu open, mode switch, debounced search) rather than effects; the picker component itself has a single unmount-cleanup effect.
- Extracted a shared `EmptyState` component from EmoteList's inline empty-search markup (its `No results...` string is now i18n'd as a side effect).

**Verification**
Tested live on Twitch in popout and standard chat: channel availability gating (control hidden on channels without gifs), mode toggle both directions, trending, search, a real send, and the server 30s cooldown path. Embedded chat not explicitly tested; YouTube is untouched (feature is gated to Twitch).

**Notes**
- A network-failed send is silent, matching the native picker's behavior.


dclstn:
https://github.com/user-attachments/assets/ca217cef-c110-4f23-8ae4-ccc756b2a0e8

dclstn:
MVP: Repurpose the sidebar for search terms like funny, sad and etc. And also favourites and such

src/modules/emote_menu/utils/twitch-gifs.js
```diff
@@ -0,0 +1,174 @@
...
+export async function getGifPickerContext() {
+  const currentChannel = getCurrentChannel();
+  if (getPlatform() !== PlatformTypes.TWITCH || currentChannel?.provider !== 'twitch') {
+    return null;
+  }
+
+  try {
+    // relies on the twitch apollo client's cache, so reopening the menu on the
+    // same channel doesn't refetch
+    const {data} = await twitch.graphqlQuery(GIF_PICKER_CONTEXT_QUERY, {channelID: currentChannel.id});
```
night: is this really not something we can pull from the current chat data? we should avoid additional api requests wherever possible

src/modules/emote_menu/utils/twitch-gifs.js
```diff
@@ -0,0 +1,174 @@
...
+    searchTerm,
+    width,
+    height,
+  };
+}
+
+export async function fetchGifs({apiKey, rating, searchTerm}) {
+  const isSearch = searchTerm.length > 0;
+  const url = new URL(`${GIPHY_API_BASE_URL}/${isSearch ? 'search' : 'trending'}`);
+  url.searchParams.set('api_key', apiKey);
```
night, line 87: is this something we should be fetching directly? does twitch expose a method we can call instead?

src/modules/emote_menu/utils/twitch-gifs.js
dclstn: good call — twitch already fetches `gifPickerConfig`, `subscriptionProducts.hasGifs` and `subscriptionBenefit.tier` on chat load (verified: they're in the apollo cache on a cold load before our menu is ever opened, and there's no cache persistence so it's a live fetch each load). switched to a `cache-only` policy in e8e2f32d so we read those straight from twitch's cache and never issue our own request. a cache miss just fails closed to "gifs unavailable", which in practice can't happen since the menu button only appears after chat (and the prefetch) has loaded.

src/modules/emote_menu/utils/twitch-gifs.js
dclstn: this one's a mutation (sending the gif) rather than a fetch, so it can't come from cache. i dug through twitch's bundle — their picker sends via an inline handler in the ChatInput class: `apollo.performStandaloneGraphQLRequest(sendGifMessageDoc, {input})`. that's the same apollo client running the same mutation we run here (our `graphqlMutation` goes through twitch's client too, so same auth/integrity), just wrapped in a class method that isn't exposed anywhere reachable. there's no higher-level `sendGif` service on window or in the react tree to call, and no way to send a gif without issuing this mutation. open to another approach if you know of one.

dclstn: Twitch's api calls for giphy are scoped to the components It seems, there's nothing exposed that lets us reuse a call.. if that changes I'll update this

PR #8226, opened 2026-09-17
feat: add player reset button and update PiP button

MarcFryd:
Adds a Reset button beside player settings. Updates PiP styling and adds toggles for both buttons under Interface → Player. Hover labels follow Twitch’s theme.

## Test plan

- [x] Build, lint, and formatting pass. Same warnings confirmed on unchanged upstream code.
- [x] Reset recovers from a simulated playback error without refreshing the page.
- [x] VOD position is preserved after reset.
- [x] PiP enters and exits correctly.
- [x] Both toggles work independently and persist after reload.
- [x] Button sizing and hover styling checked, including when Twitch button classes are removed.
- [x] Hover labels match Twitch’s light and dark themes.
- [x] Installed the test build in Chrome for review.

Reset relies on Twitch’s internal player API.


MarcFryd:
Found a visual bug. Adding a hover check to make sure it matches Twitch’s style

